PR #4, opened 2014-10-31
Rewrite element displayed algorithm

andreastt:
Fixes bug 26304.

### DIFF
--- a/09_elements.html
+++ b/09_elements.html
@@ -289,4 +289,202 @@
       <p>All WebDriver implementations MUST support finding elements by XPath 1.0 [[!XPATH]] with the <a href="http://dev.w3.org/html5/spec/Overview.html#interactions-with-xpath-and-xslt">edits from section 3.3</a> of the [[!html51]] specification made. If no native support is present in the browser, a pure JS implementation MAY be used. When called, the returned values MUST be equivalent of calling "<a href="http://www.w3.org/TR/DOM-Level-3-XPath/xpath.html#XPathEvaluator-evaluate">evaluate</a>" function from [[DOM-LEVEL-3-XPATH]] with the result type set to "<a href="http://www.w3.org/TR/DOM-Level-3-XPath/xpath.html#XPathResult-ORDERED-NODE-SNAPSHOT-TYPE">ORDERED_NODE_SNAPSHOT_TYPE</a> (7).</p>
     </section>
   </section>
+
+  <section>
+   <h2 id=element-displayedness>Element Displayedness</h2>
+
+   <p>The visibility of a <a href=http://w3c.github.io/dom/#element>Document element</a>
+    is guided by what is perceptually visible to the human eye.
+    An element's displayedness does in this context thus not relate
+    in any way to the
+    <code><a href=http://dev.w3.org/csswg/css-box/#visibility>visibility</a></code>
+    or <code><a href=http://dev.w3.org/csswg/css-box/#display0>display</a></code>
+    style properties [[!CSS3BOX]].
+
+   <p>The approach used by WebDriver to ascertain an element's
+    visibility is based on crude approximations about the element's
+    nature and relationship in the tree.  An element is in general
+    to be considered visible if any part of it is drawn on the
+    canvas within the bounderies of the viewport.
+
+   <p>The <dfn>element displayed</dfn> algorithm is
+    a boolean state where true signifies that the element is displayed
+    and false not displayed.
+    To compute the state on <var>element</var>:
+
+   <ol>
+    <li>If the attribute <code>hidden</code> is set, return false.
+
+    <li>If the computed value of the <code>display</code> style
+     property is "none", return false.
+
+    <!-- TODO(ato): Not really sure what this means, needs review: -->
+    <li>If it has a [[!CSS3-2D-TRANSFORMS]] or [[!CSS3-3D-TRANSFORMS]]
+     style property that gives a negative X or Y coordinates to the canvas, return false.
+
+    <li>If <var>element</var> is the document's root element,
+     that is <code>document.documentElement</code>:
+
+     <ol>
+      <li>If the computed value of the <code>background-color</code>
+       property is "transparent", run these substeps:
+
+       <ol>
+        <li>If <var>element</var> is an HTML <code>HTML</code>
+         element [[!html51]], and the computed value of the
+         <code>background-color</code> style property of the first
+         <code>BODY</code> element descendant of the element in
+         <a href=http://www.whatwg.org/specs/web-apps/current-work/multipage/infrastructure.html#tree-order>tree
+         order</a>, relative to that element, is also "transparent",
+         return false.
+
+        <li>Otherwise return false.
+       </ol>
+     </ol>
+
+    <li>If <var>element</var> is an <code>option</code>
+     or <code>optgroup</code> element,
+     and <var>element</var>'s parent node is a <code>select</code> element:
+
+     <ol>
+      <li>Apply the <a>element displayed</a> algorithm on <var>element</var>'s parent node.
+      <li>If the return value is false, abort these steps and return that value.
+     </ol>
+
+    <li>If <var>element</var> is a <code>map</code> element,
+     run these substeps:
+
+     <ol>
+      <li>Let <var>any images visible</var> be a boolean initially
+       set to false.
+
+      <li>For each <code>img</code> element, <var>image element</var>
+       in the document with a <code>name</code> attribute matching
+       the value of <var>element</var>'s <code>usemap</code>
+       attribute, run these substeps:
+
+       <ol>
+        <li>Run the <a>element displayed</a> algorithm on
+         <var>image element</var> and set <var>any images visible</var>
+         to <var>any images visible</var> bitwise OR its return value.
+       </ol>
+
+      <li>If <var>any images visible</var> is true, abort these
+       steps and return its value.
+    </ol>
+
+    <li>If <var>element</var> is an <code>area</code> element:
+
+     <ol>
+      <li>For each ancestral element <var>parent</var>,
+       in <a href=http://www.whatwg.org/specs/web-apps/current-work/multipage/infrastructure.html#tree-order>tree order</a>,
+       run the following substeps:
+
+      <ol>
+       <li>If <var>parent</var> is a <code>map</code> element,
+        apply the <a>element displayed</a> algorithm on it.
+
+       <li>If the return value is false, abort these steps and
+        return that value.
+
+       <li>Otherwise apply step 7.1 on <var>parent</var>.
+      </ol>
+     </ol>
+
+    <li>If <var>element</var> is a [[!DOM4]] <a
+     href=http://www.w3.org/TR/2012/WD-dom-20120105/#interface-text>
+     text node</a>, return true.
+
+    <li>If it has equal to or more than one direct descendant elements:
+
+     <ol>
+      <li>Let <var>visible children</var> be a boolean initially set to false.
+
+     <li>For each descendant element <var>child</var>,
+      in <a href=http://www.whatwg.org/spec/web-apps/current-work/multipage/infrastructure.html#tree-order>tree order</a>,
+      run the following substeps:
+
+      <ol>
+       <li>Let <var>rectangle</var> be the <a
+        href=http://dev.w3.org/fxtf/geometry/Overview.html#dom-domrect-domrect>DOMRect</a>
+        returned by calling <code>getBoundingClientRect</code>
+        on <var>child</var>.
+
+       <li>If the value of the <code>height</code> property of
+       <var>rectangle</var> is
+        greater than zero CSS pixels, and the value of the
+        <code>width</code> property of <var>rectangle</var>
+        is greater than zero CSS pixels:
+
+        <ol>
+         <li>Set <var>visible children</var> to <var>visible children</var> bitwise OR true.
+        </ol>
+      </ol>
+     </ol>
+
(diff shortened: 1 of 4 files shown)


09_elements.html
```diff
@@ -97,7 +97,227 @@
...
+
+       <ol>
+        <li>If <var>element</var> is an HTML <code>HTML</code>
+         element [[!HTML51]], and the computed value of the
+         <code>background-color</code> style property of the first
+         <code>BODY</code> element descendant of the element in
+         <a
+         href=http://www.whatwg.org/specs/web-apps/current-work/multipage/infrastructure.html#tree-order>tree
+         order</a>, relative to that element, is also "transparent",
+         return false.
```
AutomatedTester: They way that I am reading this it is saying that HTML and BODY are no longer special cased. I can't find the minutes where this was agreed in the first place but in [Bug 24634](https://www.w3.org/Bugs/Public/show_bug.cgi?id=24634) it was agreed to special case it ( Scroll up a few lines in the [minutes](http://www.w3.org/2014/02/25-testing-minutes.html#action14) where there is a very brief discussion. 

I am ok with not special casing it but I would feel better if you could email the WG to make sure that we have consensus

09_elements.html
```diff
@@ -97,7 +97,227 @@
...
+     <code>optgroup</code> element:
+
+     <ol>
+      <li>For each immediate ancestral element <var>parent</var>
+      of <var>element</var>:
+
+      <ol>
+       <li>If <var>parent</var> is a <code>select</code> element,
+        return the result of applying the <a>element visibility</a>
+        algorithm on it if its result is false.
```
AutomatedTester: The end of the sentence loses me. Can you be more explicit instead of using `it` and `its`

09_elements.html
```diff
@@ -97,7 +97,227 @@
...
+      <p>All WebDriver implementations MUST support finding elements by XPath 1.0 [[!XPATH]] with the <a href="http://dev.w3.org/html5/spec/Overview.html#interactions-with-xpath-and-xslt">edits from section 3.3</a> of the [[!HTML51]] specification made. If no native support is present in the browser, a pure JS implementation MAY be used. When called, the returned values MUST be equivalent of calling "<a href="http://www.w3.org/TR/DOM-Level-3-XPath/xpath.html#XPathEvaluator-evaluate">evaluate</a>" function from [[DOM-LEVEL-3-XPATH]] with the result type set to "<a href="http://www.w3.org/TR/DOM-Level-3-XPath/xpath.html#XPathResult-ORDERED-NODE-SNAPSHOT-TYPE">ORDERED_NODE_SNAPSHOT_TYPE</a> (7).</p>
     </section>
   </section>
+
+  <section>
+   <h2 id=element-visibility>Element Visibility</h2>
```
AutomatedTester: At [TPAC 2013](http://www.w3.org/2013/11/11-testing-minutes.html#item08) it was decided that we will will not use displayed because there are situations where it is not visible to the eye because an element in a different branch of the tree could be obscuring it. It was decided that we will use `displayed`. Please update the prose below and the header above to match.

09_elements.html
andreastt: The conclusion we came to in the WG meeting is wrong because of the [last comment](https://www.w3.org/Bugs/Public/show_bug.cgi?id=24634#c6) in that bug.  Basically `body`'s `background-color` is only looked at if `html`'s `background-image` is unset and `background-color` is "transparent".  See section 3.11.2 in the [CSS Backgrounds specification](http://dev.w3.org/csswg/css-backgrounds/#special-backgrounds) which explains this quite clearly.

AutomatedTester: The working group decided to not follow other specs here. This has not been discussed again, please raise it again before we can go further with this

09_elements.html
```diff
@@ -97,7 +97,225 @@
...
+      <p>All WebDriver implementations MUST support finding elements by XPath 1.0 [[!XPATH]] with the <a href="http://dev.w3.org/html5/spec/Overview.html#interactions-with-xpath-and-xslt">edits from section 3.3</a> of the [[!HTML51]] specification made. If no native support is present in the browser, a pure JS implementation MAY be used. When called, the returned values MUST be equivalent of calling "<a href="http://www.w3.org/TR/DOM-Level-3-XPath/xpath.html#XPathEvaluator-evaluate">evaluate</a>" function from [[DOM-LEVEL-3-XPATH]] with the result type set to "<a href="http://www.w3.org/TR/DOM-Level-3-XPath/xpath.html#XPathResult-ORDERED-NODE-SNAPSHOT-TYPE">ORDERED_NODE_SNAPSHOT_TYPE</a> (7).</p>
     </section>
   </section>
+
+  <section>
+   <h2 id=element-displayedness>Element Disiplayness</h2>
```
abotalov: Disiplayness - probably typo

10_element_state.html
```diff
@@ -23,55 +23,74 @@
...
+          element</a> status and return <var>response</var>.
+
+         <li>If <var>element</var> represents <a
+          href=http://w3c.github.io/dom/#element>Document element</a>
+          that is no longer attached to the document's node tree,
+          set <var>response</var>'s <a href=#response-status>status</a>
+          to the <a>stale element reference</a> status and return
+          <var>response</var>.
+
+         <li>Apply the <a>element visibility</a> algorithm to
```
abotalov: Should the algorithm be also called "displayness" here and in other places?

09_elements.html
andreastt: https://lists.w3.org/Archives/Public/public-browser-tools-testing/2015JanMar/0088.html

andreastt:
Rebased PR.

09_elements.html
```diff
@@ -289,4 +289,202 @@ <h2 id="element-location-strategies">Element Location Strategies</h2>
...
+    <!-- TODO(ato): Not really sure what this means, needs review: -->
+    <li>If it has a [[!CSS3-2D-TRANSFORMS]] or [[!CSS3-3D-TRANSFORMS]]
+     style property that gives a negative X or Y coordinates to the canvas, return false.
+
+    <li>If <var>element</var> is the document's root element,
+     that is <code>document.documentElement</code>:
+
+     <ol>
+      <li>If the computed value of the <code>background-color</code>
+       property is "transparent", run these substeps:
```
sevaseva: The sub-list (logical branch) that follows seems to always resolve to false; logically redundant as it is. Some logic accidentally left out?

09_elements.html
```diff
@@ -289,4 +289,202 @@ <h2 id="element-location-strategies">Element Location Strategies</h2>
...
+     href=http://www.w3.org/TR/2012/WD-dom-20120105/#interface-text>
+     text node</a>, return true.
+
+    <li>If it has equal to or more than one direct descendant elements:
+
+     <ol>
+      <li>Let <var>visible children</var> be a boolean initially set to false.
+
+     <li>For each descendant element <var>child</var>,
+      in <a href=http://www.whatwg.org/spec/web-apps/current-work/multipage/infrastructure.html#tree-order>tree order</a>,
```
sevaseva: "in tree order" is redundant here (given that all children are to be processed).

andreastt: Patches welcome (-:

09_elements.html
```diff
@@ -289,4 +289,202 @@ <h2 id="element-location-strategies">Element Location Strategies</h2>
...
+    <li>If <var>element</var> is a [[!DOM4]] <a
+     href=http://www.w3.org/TR/2012/WD-dom-20120105/#interface-text>
+     text node</a>, return true.
+
+    <li>If it has equal to or more than one direct descendant elements:
+
+     <ol>
+      <li>Let <var>visible children</var> be a boolean initially set to false.
+
+     <li>For each descendant element <var>child</var>,
```
sevaseva: if "descendant elements" really means "direct descendant elements" here, please change to "direct descendant elements". If the text really means to say "descendant elements" (all of them), it would be helpful to emphasize that in the text (because "direct descendant elements" mentioned two sentences above may create confusion).

09_elements.html
```diff
@@ -289,4 +289,202 @@ <h2 id="element-location-strategies">Element Location Strategies</h2>
...
+    </ol>
+
+    <li>Return true.
+   </ol>
+
+   <p>If the steps above return true, then the element MUST also
+    be <a href=#interactable-elements>interactable</a> should it
+    meet the other criteria for being interactable.
+    If any part of the <code>BODY</code> can be brought into the current viewport,
+    the return value MUST be true.
```
sevaseva: What's the idea behind this sentence? It is not a part of the algorithm, is it? Is it a partial 'test' for the algorithm defined above? It is not clear to me how an implementer of this spec would use this sentence.

sevaseva: Oh... did this sentence mean to say how to determine displayedness of a BODY element?
(If so, I think that is unclear from the text.)

09_elements.html
```diff
@@ -289,4 +289,202 @@ <h2 id="element-location-strategies">Element Location Strategies</h2>
...
+     </ol>
+    </ol>
+
+    <li>Return true.
+   </ol>
+
+   <p>If the steps above return true, then the element MUST also
+    be <a href=#interactable-elements>interactable</a> should it
+    meet the other criteria for being interactable.
+    If any part of the <code>BODY</code> can be brought into the current viewport,
```
sevaseva: What is 'any part of the BODY' ?
It may be just me, but I think that is ambiguous, vague and does not signify a thing specific enough for the purpose of this document.

Next, 'can be brought into the current viewport' is, I think, broad and unspecific. For example, one may be able to bring a 'part of the BODY' into the current viewport by modifying current state of the DOM (hiding, showing some elements, changing their size and other properties). I think this sentence didn't mean to 'allow' such means for bringing 'any part of the BODY' into the current viewport; it should be changed to actually rule out such things.

09_elements.html
sevaseva: I will do patches (especially for simple things like this one; I promise).

If you welcome me send patches for some of other comments I just left, where resolution is less obvious, may I ask you please reply to those comments and confirm/clarify things. For most comments I am missing the context (that's in your head) necessary to resolve the issues.

sevaseva:
General comment/question.
Previous version of the displayedness algorithm was written after the implementation in the selenium atoms. So, by design, we knew if was at least implementable in a usable way. That felt comforting to me. Now I don't feel as comfortable because I don't know of an implementation of this algorithm that's working. Especially that this version seems to be traversing the DOM a lot more and potentially much slower. I don't feel certainty that it is implementable in a practically useful way.

sevaseva:
So the question part of the 'General comment/question' would be:
Do you know of an existing implementation of this algorithm or a specific plan to create one? (hopefully licensed openly enough so it is relatively easy for browser vendors to use that directly if they want and allow themselves to [although I realize that is in no way a requirement])

09_elements.html
sevaseva: See #22

09_elements.html
sevaseva: See #21

09_elements.html
sevaseva: See #23

09_elements.html
```diff
@@ -289,4 +289,202 @@ <h2 id="element-location-strategies">Element Location Strategies</h2>
...
+      </ul>
+
+      <li>Run step 10 on the parent elements of <var>parent</var>, if any.
+     </ol>
+    </ol>
+
+    <li>Return true.
+   </ol>
+
+   <p>If the steps above return true, then the element MUST also
```
sevaseva: Proposed to remove this statement as incorrect, #24

09_elements.html
sevaseva: See #25

09_elements.html
andreastt: I guess “direct descendant elements” is more accurate.

09_elements.html
andreastt: I honestly don't know, it was there from before I made my changes.  I think it's probably wrong.

09_elements.html
andreastt: Again, I didn't add this.

andreastt:
> Previous version of the displayedness algorithm was written after the implementation in the selenium atoms.

I reviewed the previous edition against the atoms, and it was lacking in so many ways that I don't think we can say that it was either spec-proof or compliance-proof.

> I don't feel certainty that it is implementable in a practically useful way.

I agree.  I think it's a bad idea to implement element visibility altogether, and this algorithm proves it's hard to get right.

> Do you know of an existing implementation of this algorithm or a specific plan to create one?

There are no implementations of this algorithm.  I think you'll find it has a lot of problems if you attempt an implementation.

sevaseva:
Interesting... You say "I think it's a bad idea to implement element visibility altogether". Do you actually mean that **implementing it** is a bad idea or **spec'ing it in algorithm stype** (which btw matches my opinion that didn't meet much support a couple of face-to-faces ago) or something else? Also, what do you think is a better idea here?

09_elements.html
sevaseva: Ah, right, sorry, I missed the fact that it was there already. Anyway, it looks like removing both these sentences should work and be agreable.

andreastt:
Sorry, I should've been clearer.  I mean that element visibility, even for a tiny subset of parameters with quite clear statements of limitations, is difficult for us to define at all in the specification.  I will write more about this soon, as there new web technologies that puts our thinking around visibility even more at edge with reality, maybe even pushing it to the point of us doing the _wrong_ thing here.

But all of this is off-topic, and we should have our discussion elsewhere.

PR #7, opened 2014-12-30
Fix several typos in the getElementAttribute prose

abotalov:
Also I don't understand the reason why `getAttributeNode` is invoked instead of `getAttribute`. `specified` [always returns true](http://www.w3.org/TR/dom/#dom-attr-specified)


andreastt:
Can you file an issue about the needless use of `specified`? (Also you should never refer to specs in /TR, but https://dom.spec.whatwg.org/#attr agrees.)

abotalov:
I filed https://www.w3.org/Bugs/Public/show_bug.cgi?id=27733

abotalov:
@shs96c I updated PR (if -> iff) with force to create fewer commits.

shs96c:
Merged into master. Thank you!

PR #9, opened 2015-02-11
sessions: attempt to reuse expired session should result in error

andreastt:
Fixes #27998

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -425,10 +425,10 @@ code a:visited, code a:link {
  or implicitly when <a href=#close>close</a> is called at the last remaining
  <a href=http://www.w3.org/TR/html5/browsers.html#top-level-browsing-context>top level browsing context</a>.
 
-<p>Each <a>remote end</a> has an associated <dfn>list of active sessions</dfn>,
+<p>A <a>remote end</a> has an associated <dfn>list of active sessions</dfn>,
  which is a list of all <a rel="session">sessions</a> that are currently started.
 
-<p>The <a>remote end</a> has an associated <a>current session</a>
+<p>A <a>remote end</a> has an associated <a>current session</a>
  that is the <a rel="active session">session that is active</a>
  for the duration of a <a>command</a>.
 
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -436,7 +436,7 @@ code a:visited, code a:link {
  used to uniquely identify thi session.
  Unless stated otherwise it is null.
 
-<p>A <a>session</a> has an associated <dfn>current browsing context</dfn>
+<p>A <a>session</a> has an associated <dfn>active browsing context</dfn>
  (a reference to <a href=http://www.w3.org/TR/html5/browsers.html#top-level-browsing-context>top level browsing context</a>) that indicates what <a>browsing context</a>
  commands are executed against.
 
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -459,7 +459,7 @@ code a:visited, code a:link {
 
 <p>A <a>session</a> has an associated <dfn>known sessions</dfn>
  (a sequence of <a>UUID</a>'s) that list
- <a rel="active session">sessions that are active in a <a>remote end</a>.
+ <a rel="active session">sessions that are active</a> in a <a>remote end</a>.
 
 <p>A <a>current session</a> (a <a>UUID</a>)
  is the <a rel="active session">session that is active</a>
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -196,13 +196,39 @@ code a:visited, code a:link {
       <p>When the HTTP request has been parsed a call to the underlying WebDriver Command using the following algorithm:</p>
       <ol>
         <li>Let <var>command</var> be the command that has the corresponding URI as described throughout this document</li>
-        <li>Let <var>result</var> be an object that contains the following:</li>
+        <li>Let <var>response</var> be an initially empty object with the following fields:</li>
           <ul>
-            <li>Let <var>status</var> contain the error status if a command has failed.</li>
-            <li>Let <var id='return-value'>value</var> contain any return values from the command.
-                The contents are described in the methods throughout this document. If <var>status</var>
-                is not empty <var>value</var> will contain the error message describing why there was a failure.</li>
+            <li><var>status</var>, initially set to <a>success</a>.
+            <li><var>value</var>, initially set to null.
           </ul>
+        <li>If the <a>session</a>'s <a>alive flag</a> is not set
+         and the <a>session</a>'s <a>session ID</a> is not null, run these substeps:
+          <ul>
+            <li>Set <var>response</var>'s <var>status</var> field to <a>session expired</a>.
+            <li>Return <var>response</var>.
+          </ul>
+        <li>Else if the <a>session</a>'s <a>session ID</a> is null
+         and <var>command</var> is not <a href=#newsession>new session</a>, run these substeps:
+          <ul>
+            <li>Set <var>response</var>'s <var>status</var> field to <a>invalid sessionid</a>.
+            <li>Return <var>response</var>.
+          </ul>
+        <li>Else if <var>session ID</var> is not equal to the <a>session</a>'s <var>session ID</var>,
+         run these substeps:
+          <ul>
+            <li>Set <var>response</var>'s <var>status</var> field to <a>invalid sessionid</a>.
+            <li>Return <var>response</var>.
+          </ul>
+        <li>If evaluating <var>command</var> yielded an error:
+          <ul>
+            <li>Set <var>response</var>'s <var>status</var> to
+             the appropriate status code corresponding to the error,
+             or <a>unknown error</a>.
+            <li>Return <var>response</var>.
+           </ul>
+        <li>Let <var>response</var>'s <var>value</var> be the return value
+         from the evaluated <var>command</var>.
+        <li>Return <var>response</var>.
       </ol>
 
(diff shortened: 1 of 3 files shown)


andreastt:
Updated PR to incorporate the ideas from @jgraham on intermediaries.

webdriver-spec.html
```diff
@@ -375,20 +404,79 @@ <h4 id='returning-http-response'>Returning a HTTP response to the local end</h4>
...
+<p>A <a>current session</a> (a <a>UUID</a>)
+ is the <a rel="active session">session that is active</a>
+ for the duration of a <a>command</a>.
+<!-- TODO: Doesn't define what "active" means -->
+
+<p class=note>The concept of sessions
+ makes it possible to implement <a>intermediary notes</a>
+ that handle requests from multiple clients simultaneously.
+
+<p>A <a>session</a> is started when
```
jgraham: This should be marked as non-normative.

webdriver-spec.html
```diff
@@ -375,20 +404,79 @@ <h4 id='returning-http-response'>Returning a HTTP response to the local end</h4>
...
+<p>A <a>session</a> has an associated <dfn>known sessions</dfn>
+ (a sequence of <a>UUID</a>'s) that list
+ <a rel="active session">sessions that are active in a <a>remote end</a>.
+
+<p>A <a>current session</a> (a <a>UUID</a>)
+ is the <a rel="active session">session that is active</a>
+ for the duration of a <a>command</a>.
+<!-- TODO: Doesn't define what "active" means -->
+
+<p class=note>The concept of sessions
```
jgraham: Actually this isn't quite true (I know I wrote it). The concept of a session allows global state in the remote end that lasts for the duration of a session. The ability to have multiple sessions in a specific remote end allows multiplexing intermediaries. On a related note, defining the behaviour of |new session| probably requires that we allow an implementation-specific |maximum number of sessions|, which we can check against the length of |known sessions| to see if we can start a session or not.

webdriver-spec.html
```diff
@@ -375,20 +404,79 @@ <h4 id='returning-http-response'>Returning a HTTP response to the local end</h4>
...
+ that specifies a time in milliseconds to wait for scripts to run.
+
+<p>A <a>session</a> has an associated <dfn>page loading timeout</dfn> (an integer)
+ that specifies a time in milliseconds to wait for the page loading to complete.
+
+<p>A <a>session</a> has an associated <dfn>known sessions</dfn>
+ (a sequence of <a>UUID</a>'s) that list
+ <a rel="active session">sessions that are active in a <a>remote end</a>.
+
+<p>A <a>current session</a> (a <a>UUID</a>)
```
jgraham: Surely the |current session| is a |session|, not a UUID.

webdriver-spec.html
andreastt: Marking this paragraph as non-normative, but I think once we move all the definitions out we can mark the whole section as non-normative.

webdriver-spec.html
andreastt: I'll use that prose exactly and we can improve it later. Filed [bug 28002](https://www.w3.org/Bugs/Public/show_bug.cgi?id=28002) about introducing a `maximum number of sessions` concept.

webdriver-spec.html
```diff
@@ -375,20 +395,75 @@ <h4 id='returning-http-response'>Returning a HTTP response to the local end</h4>
...
+<!--
+ TODO(jgraham): Need to explain that there must be
+ a relationship between the current session
+ and the browser state.
+-->
+
+<p>The <a>session</a> is set up at the invocation of
+ a <a href=#newsession>new session</a>,
+ and torn down at some later point;
+ either explicitly by invoking <a href=#>quit</a>,
```
jgraham: Not sure that "quit" is a thing (I assume you mean the command that results from DELETE /session/{sessionId}, but we can fix that later.

webdriver-spec.html
```diff
@@ -375,20 +395,75 @@ <h4 id='returning-http-response'>Returning a HTTP response to the local end</h4>
...
+ or implicitly when <a href=#close>close</a> is called at the last remaining
+ <a href=http://www.w3.org/TR/html5/browsers.html#top-level-browsing-context>top level browsing context</a>.
+
+<p>Each <a>remote end</a> has an associated <dfn>list of active sessions</dfn>,
```
jgraham: s/Each/A/ and s/The/A/ in this and the following paragraph, I think.

webdriver-spec.html
```diff
@@ -375,20 +395,75 @@ <h4 id='returning-http-response'>Returning a HTTP response to the local end</h4>
...
+
+<p>The <a>remote end</a> has an associated <a>current session</a>
+ that is the <a rel="active session">session that is active</a>
+ for the duration of a <a>command</a>.
+
+<p>A <a>session</a> has an associated <dfn>session ID</dfn> (a <a>UUID</a>)
+ used to uniquely identify thi session.
+ Unless stated otherwise it is null.
+
+<p>A <a>session</a> has an associated <dfn>session capabilities</dfn>
```
jgraham: I don't actually know if this is needed once the session is started. Does any behaviour actually switch on the values of the capabilities?

webdriver-spec.html
```diff
@@ -375,20 +395,75 @@ <h4 id='returning-http-response'>Returning a HTTP response to the local end</h4>
...
+ object (a <a>capabilities</a> object)
+ describing the functionality provided by this <a>session</a>.
+ Unless stated otherwise it is null.
+
+<p>A <a>session</a> has an associated <dfn>current browsing context</dfn>
```
jgraham: Can we give this a name without "current" in it? Because I think you will often want to say "the |current session|'s |current browsing context|", which sounds weird. "active" is quite close to what HTML uses.

webdriver-spec.html
```diff
@@ -375,20 +395,75 @@ <h4 id='returning-http-response'>Returning a HTTP response to the local end</h4>
...
+ (a reference to <a href=http://www.w3.org/TR/html5/browsers.html#top-level-browsing-context>top level browsing context</a>) that indicates what <a>browsing context</a>
+ commands are executed against.
+
+<p>A <a>session</a> has an associated <dfn>script timeout</dfn> (an integer)
+ that specifies a time in milliseconds to wait for scripts to run.
+
+<p>A <a>session</a> has an associated <dfn>page loading timeout</dfn> (an integer)
+ that specifies a time in milliseconds to wait for the page loading to complete.
+
```
jgraham: There is also an implict wait timeout and a page loading strategy, I think.

webdriver-spec.html
```diff
@@ -375,20 +395,75 @@ <h4 id='returning-http-response'>Returning a HTTP response to the local end</h4>
...
+<p class=note>The concept of sessions
```
jgraham: I think this is covered at the top.

webdriver-spec.html
andreastt: I checked a couple of common capabilities and Marionette, and the only situation I can think of where one might want to do that is the driver is asked to enable some proprietary feature and you don't want to store that in a separate variable.

So I think you're right that we can actually remove this.

webdriver-spec.html
```diff
@@ -375,20 +395,75 @@ <h4 id='returning-http-response'>Returning a HTTP response to the local end</h4>
...
+ allows global state in the <a>remote end</a>
+ that lasts for the duration of a session.
+ The ability to have multiple sessions in a specific remote end
+ allows multiplexing <a rel="intermediary node">intermediaries</a>.
+
+<p><em>Non-normative:</em>
```
jgraham: I'm not sure but maybe put all the non-normative bits together (in a sub-section)?

webdriver-spec.html
```diff
@@ -375,20 +395,75 @@ <h4 id='returning-http-response'>Returning a HTTP response to the local end</h4>
...
+ or implicitly when <a href=#close>close</a> is called at the last remaining
+ <a href=http://www.w3.org/TR/html5/browsers.html#top-level-browsing-context>top level browsing context</a>.
+
+<p>Each <a>remote end</a> has an associated <dfn>list of active sessions</dfn>,
+ which is a list of all <a rel="session">sessions</a> that are currently started.
+
+<p>The <a>remote end</a> has an associated <a>current session</a>
+ that is the <a rel="active session">session that is active</a>
+ for the duration of a <a>command</a>.
+
```
jgraham: Need to add something like:

A remote end also has a |maximum number of active sessions|, which may be "unlimited". For a browser this is typically one.

webdriver-spec.html
andreastt: Changing to “active browsing context”.   So it'll bee “the current session's active browsing context”, which indeed sounds a lot better.

webdriver-spec.html
andreastt: I feel this will be solved when we extract all the definitions into a separate chapter, because we can then mark the whole section as non-normative.

webdriver-spec.html
andreastt: Added, and filed [bug 28004](https://www.w3.org/Bugs/Public/show_bug.cgi?id=28004) and [bug 28005](https://www.w3.org/Bugs/Public/show_bug.cgi?id=28005) about discussing the defaults in the WG.

webdriver-spec.html
andreastt: Yes, exactly. But it isn't defined yet so we'll have to return to this.

webdriver-spec.html
andreastt: Fixed in f59baad.

AutomatedTester:
796c959c68d68a40ac3ad1ef5f6afdd9bb05dc95

PR #15, opened 2015-02-16
Rephrase the algorithm for handling HTTP requests.

jgraham:
This makes it clear what error is generated in various edge cases
and builds out some infrastructure needed to make various future parts
of the spec more well-defined.

### DIFF
--- a/02_commands.html
+++ b/02_commands.html
@@ -1,206 +1,524 @@
 <section>
-  <h2>Communication between the local end and the remote end</h2>
+  <h2>The WebDriver Protocol</h2>
   <section>
-    <h3>The JSON Wire Protocol</h3>
-    <p>When the Local End and the Remote End communicate with each other they MUST do so via HTTP. When describing the URL to use for each <code>Command</code> this specification will use URI paths encoded using the scheme described in [[!URI-TEMPLATE]]. The following substitutions are commonly made:</p>
-
-      <dl>
-        <dt>{sessionId}</dt>
-        <dd>The <code>sessionId</code> of the selected session.</dd>
-
-        <dt>{ELEMENT}</dt>
-        <dd>The unique identifier of the <code>WebElement</code> instance being referred to</dd>
-      </dl>
-
-    <p>The URLs given in this specification MUST have the same path prefix. Implementations are free to redirect those as necessary.</p>
-
-    <p class="note">The reason for doing this is to reduce the complexity of local-end implementations. They need only append the path to a base URL in order to create a working implementation.</p>
-
-    <section>
-      <h3>Handling HTTP requests and responses between the Remote End and Local End</h3>
-      <p>When communicating between the local end and the remote end the following algorithm needs to be followed.</p>
-      <ol>
-        <li><a href="#receiving-http-requests">Receive HTTP request from the local end</a></li>
-        <li><a href="#calling-webdriver-commands">Call the WebDriver command</a></li>
-        <li><a href="#returning-http-response">Returning HTTP response to the local end</a></li>
-      </ol>
-
-      <h4 id='receiving-http-requests'>Receiving HTTP requests from the local end</h4>
-      <p>When handling requests from the local end to the remote end we must do the following first and then follow the steps :
-      <ol>
-        <li>Let <var>sessionid</var> be the value in the {sessionId} section in the [[!URI-TEMPLATE]]</li>
-        <li>if {ELEMENT} section in the [[!URI-TEMPLATE]] is populated let <var id='uri-template-element'>element</var> be equal to the {ELEMENT} section</li>
-        <li>If the HTTP request is a POST let <var id='post-payload'>payload body</var> contain the
-          <a href='http://www.ecma-international.org/ecma-262/5.1/#sec-15.12.2'>parsed JSON</a>
-         in the payload body of the request.
-          If <var>payload body</var> does not <a href='http://www.ecma-international.org/ecma-262/5.1/#sec-15.12.2'>parse into JSON</a> return a HTTP response with status code 500 and the payload body containing details of the parsing error if available.
-        </li>
-      </ol>
-
-      <h4 id='calling-webdriver-commands'>Call the WebDriver command</h4>
-      <p>When the HTTP request has been parsed a call to the underlying WebDriver Command using the following algorithm:</p>
-      <ol>
-        <li>Let <var>command</var> be the command that has the corresponding URI as described throughout this document</li>
-        <li>Let <var>result</var> be an object that contains the following:</li>
-          <ul>
-            <li>Let <var>status</var> contain the error status if a command has failed.</li>
-            <li>Let <var id='return-value'>value</var> contain any return values from the command.
-                The contents are described in the methods throughout this document. If <var>status</var>
-                is not empty <var>value</var> will contain the error message describing why there was a failure.</li>
-          </ul>
-      </ol>
-
-      <h4 id='returning-http-response'>Returning a HTTP response to the local end</h4>
-      <p>when the command has completed encode the result using the following algorithm:</p>
-
-      <ol>
-        <li>Set the HTTP response status code to 200 unless described differently in the commands below</li>
-        <li>Let <var>sessionId</var> be the unique identifier for the current session</li>
-        <li>Add the contents of <var>result</var> to an object could be <a href='http://www.ecma-international.org/ecma-262/5.1/#sec-15.12.3'>stringified in to JSON</a>
-        <li>Let the payload body of the response be the <a href='http://www.ecma-international.org/ecma-262/5.1/#sec-15.12.3'>stringified JSON</a>
-          object.
-          <ol>
-            <li>If <a href="#newsession">newSession()</a> is called then the response payload body would look like the following
-              <pre>
-                {
-                  "sessionId" : "string value",
-                  "value" : capabilities-object,
-                }.
-              </pre>
-            </li>
-            <li>
-              If the WebDriver Command has to return data it should be the whole of the response payload body and look like the following
-              <pre>
-                {
-                  "value": "data from WebDriver Commands"
-                }
-              </pre>
-            </li>
-            <li>
-              If the WebDriver Command results in an error then the response payload body would look like the following
-              <pre>
-                {
-                  "status": "string value",
-                  "value" : "string message"
-                }
-              </pre>
-            </li>
-          </ol>
-        </li>
-        <li>Set HTTP headers for the follow keys and their values.
-          <ul>
-            <li><code>cache-control: no-cache</code></li>
-            <li><code>content-type: application/json</code></li>
-          </ul>
-        </li>
-        <li>Return HTTP response to the Local End</li>
-      </ol>
-    </section>
+    <h3>Introduction</h3>
+    <p>TODO</p>
   </section>
   <section>
-    <h3>Status Codes</h3>
-    <p>The WebDriver API indicates the success or failure of a command invocation via <var>status</var>. The following values are used and have the following meanings.</p>
+    <h3>Common Infrastructure</h3>
+    <h4>Algorithms</h4>
+    <p>Various parts of this specification are written in terms of step-by-step algorithms. The details of these algorithms do not have any normative significance; implementations are free to adopt any implementation strategy that produces equivalent output to the specification. In particular algorithms in this document are optimised for readability rather than performance.</p>
+    <p>Where algorithms that return values are fallible, they are written in terms of returning either <dfn>success</dfn> or <dfn>error</dfn>. A <a>success</a> value has an associated <var>data</var> field which encapsulates the value returned, whereas an <a>error</a> response has an associated <a>error code</a>.</p>
+  </section>
+
+  <section>
+    <h3>Processing Model</h3>
+    <p>The <a>remote end</a> acts as a HTTP server reading requests from the client and writing responses, typically over TCP socket. For the purposes of this specification we model the data transmission between a particular <a>local end</a> and <a>remote end</a> with a <dfn>connection</dfn> object to which the <a>remote end</a> may <dfn>write bytes</dfn> and <dfn>read bytes</dfn>. However the exact details of how this connection works and how it is established are out of scope.</p>
+
+    <p>After such a <a>connection</a> has been established, a remote end MUST run the following steps:</p>
+
+    <ol>
+      <!-- Should mention what happens if it can't be decoded as HTTP -->
+      <!-- Note: the way that fetch is used here isn't quite right because it doesn't
+           specify how to construct a request from network data, or serialize a response. -->
+
+      <li><p><a>Read bytes</a> from the connection until a complete HTTP request can be constructed from the data. Let <var>request</var> be a <a href="https://fetch.spec.whatwg.org/#concept-request" class="externalRef">request</a> object constructed from the received data, according to the requirements of [[!RFC7230]].</p></li>
+
+      <li><p>Let <var>request match</var> be the result of <a>matching a request</a> with <var>request</var>'s <a href="https://fetch.spec.whatwg.org/#concept-request-method" class="externalRef">method</a> and <a href="https://fetch.spec.whatwg.org/#concept-request-url" class="externalRef">url</a> as arguments.</p></li>
+
+      <li><p>If <var>request match</var> is of type <a>error</a>, <a>return an error</a> with <var>request match</var>'s <a>error code</a> and jump to step 1.</p>
+
+        <p>Otherwise, let <var>command</var>, <var>session id</var> and <var>element id</var> be <var>request match</var>'s data.</p></li>
+
+      <li><p>If <var>command</var> is not <a>new session</a>:</p>
+        <ol>
+          <li><p>If <var>session id</var> is not equal to the <a>id</a> of any session in the <a>list of active sessions</a>, <a>return an error</a> with <a>error code</a> <a>invalid session id</a>, then jump to step 1.</p>
+
+          <p>Otherwise, let the <a>current session</a> be the session with <a>id</a> <var>session id</var></p>
+          </li>
+        </ol>
+      </li>
+
+      <li><p>If <var>request</var>'s <a href="https://fetch.spec.whatwg.org/#concept-request-method" class="externalRef">method</a> is POST:</p>
+        <ol>
+
+          <li><p>Let <var>success</var> and <var>parameters</var> be the result of <a>parsing as JSON</a> with <var>request</var>'s <a href="https://fetch.spec.whatwg.org/#concept-request-body" class="externalRef">body</a> as the argument.</p></li>
+
+          <li><p>If <var>success</var> is false, or <var>success</var> is true and <var>parameters</var> is not an Object object, <a>return an error</a> with status <a>error code</a> <a>invalid argument</a> and jump back to step 1 in this overall algorithm.</p></li>
+        </ol>
+
+        <p>Otherwise, let <var>parameters</var> be null</p>
+      </li>
+
+      <li><p>Let <var>response data</var> be a <a>command response</a> object obtained by running running the steps defined by <var>command</var> with arguments <var>element id</var> and <var>parameters</var>.</p></li>
+
+      <li><p>If <var>response data</var> is an <a>error</a>, <a>return an error</a> with <a>error code</a> equal to <var>response data</var>'s <a>error code<a>.</p>
+
+        <p>Otherwise <a>send a response</a> with status 200 and <var>response data</var>'s data.</p></li>
+
+      <li><p>Jump to step 1</p></li>
+    </ol>
+
+    <p>When required to <dfn>send a response</dfn>, with arguments <var>status</var> and <var>data</var>, a remote end must run the following steps:</p>
+
+    <ol>
+      <li><p>Let <var>response</var> be a new <a href="https://fetch.spec.whatwg.org/#concept-response" class="externalRef">response</a>.</p></li>
+
+      <li><p>Set <var>response</var>'s <a href="https://fetch.spec.whatwg.org/#concept-response-status" class="externalRef">status</a> to <var>status</var>, and <a href="https://fetch.spec.whatwg.org/#concept-response-status-message" class="externalRef">status message</a> to the string corresponding to the description of <var>status</var> in the <a href="http://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml" class="externalRef">status code registry</a>.</p></li>
+      <li><p>If <var>data</var> is not null, let <var>response</var>'s <a href="https://fetch.spec.whatwg.org/#concept-response-body" class="externalRef">body</a> be the result of <a>serializing as JSON</a> with <var>data</var> as the argument.</p></li>
+
+      <li><p>Serialize <var>response</var> according to the rules in [[!RFC7230]] and write the resulting bytes to the <a>connection</a></p></li>
+
+    </ol>
+
+    <p>When required to <dfn>return an error</dfn>, with <var>error code</var>, run the following steps:</p>
+
+    <ol>
+      <li><p>Let <var>http status</var> and <var>name</a> be the <a>error response data</a> for <var>error code</var></p></li>
+      <li><p>Let <var>data</var> be a <a>new JSON Object</a>.</p>
+
+      <li><p>Let <var>message</var> be an implementation-defined string containing a human-readable description of the reason for the error.</p></li>
+
+        <!-- TODO: really need a better way of constructing JSON literals -->
+      <li><p><a>Set the properties</a> of <var>object</var> with values ("error", <var>name</var>), ("message", <var>message</var>).</p></li>
+
+      <li><a>Send a response</a> with status <var>status</var> and data <var>data</var></li>
+
+    </ol>
+
+  </section>
+
+  <section>
+    <h3>Routing Requests</h3>
+
+    <!-- TODO: define what a command is -->
+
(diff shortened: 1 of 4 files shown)


02_commands.html
```diff
@@ -1,206 +1,524 @@
...
-    </section>
+    <h3>Introduction</h3>
+    <p>TODO</p>
   </section>
   <section>
-    <h3>Status Codes</h3>
-    <p>The WebDriver API indicates the success or failure of a command invocation via <var>status</var>. The following values are used and have the following meanings.</p>
+    <h3>Common Infrastructure</h3>
+    <h4>Algorithms</h4>
+    <p>Various parts of this specification are written in terms of step-by-step algorithms. The details of these algorithms do not have any normative significance; implementations are free to adopt any implementation strategy that produces equivalent output to the specification. In particular algorithms in this document are optimised for readability rather than performance.</p>
```
andreastt: No change necessary here, but:  It seems this applies to algorithms also in other chapters apart from the this about protocol.  We should consider uplifting this to a separate chapter on _Infrastructure_ in the future.

02_commands.html
```diff
@@ -1,206 +1,524 @@
...
+    <p>Where algorithms that return values are fallible, they are written in terms of returning either <dfn>success</dfn> or <dfn>error</dfn>. A <a>success</a> value has an associated <var>data</var> field which encapsulates the value returned, whereas an <a>error</a> response has an associated <a>error code</a>.</p>
```
andreastt: Both success and error responses have a `status` field.  Conceptually there isn't a difference between “error code” and a “status”.

andreastt: The s/status/error code/ here makes sense.  I re-read (the minutes)[http://www.w3.org/2014/10/30-testing-minutes.html] and I think this encapsulates the meaning of the action.

02_commands.html
```diff
@@ -1,206 +1,524 @@
...
+  </section>
+
+  <section>
+    <h3>Processing Model</h3>
+    <p>The <a>remote end</a> acts as a HTTP server reading requests from the client and writing responses, typically over TCP socket. For the purposes of this specification we model the data transmission between a particular <a>local end</a> and <a>remote end</a> with a <dfn>connection</dfn> object to which the <a>remote end</a> may <dfn>write bytes</dfn> and <dfn>read bytes</dfn>. However the exact details of how this connection works and how it is established are out of scope.</p>
```
andreastt: s/object//

02_commands.html
```diff
@@ -1,206 +1,524 @@
...
+
+      <li><p>If <var>request match</var> is of type <a>error</a>, <a>return an error</a> with <var>request match</var>'s <a>error code</a> and jump to step 1.</p>
+
+        <p>Otherwise, let <var>command</var>, <var>session id</var> and <var>element id</var> be <var>request match</var>'s data.</p></li>
+
+      <li><p>If <var>command</var> is not <a>new session</a>:</p>
+        <ol>
+          <li><p>If <var>session id</var> is not equal to the <a>id</a> of any session in the <a>list of active sessions</a>, <a>return an error</a> with <a>error code</a> <a>invalid session id</a>, then jump to step 1.</p>
+
+          <p>Otherwise, let the <a>current session</a> be the session with <a>id</a> <var>session id</var></p>
```
andreastt: `<a rel="session id">id</a>` probably

02_commands.html
```diff
@@ -1,206 +1,524 @@
...
+          </li>
+        </ol>
+      </li>
+
+      <li><p>If <var>request</var>'s <a href="https://fetch.spec.whatwg.org/#concept-request-method" class="externalRef">method</a> is POST:</p>
+        <ol>
+
+          <li><p>Let <var>success</var> and <var>parameters</var> be the result of <a>parsing as JSON</a> with <var>request</var>'s <a href="https://fetch.spec.whatwg.org/#concept-request-body" class="externalRef">body</a> as the argument.</p></li>
+
+          <li><p>If <var>success</var> is false, or <var>success</var> is true and <var>parameters</var> is not an Object object, <a>return an error</a> with status <a>error code</a> <a>invalid argument</a> and jump back to step 1 in this overall algorithm.</p></li>
```
andreastt: It's unclear to me exactly what `success` indicates here.  Does it tell us if the body could successfully be parsed as JSON?  Perhaps this will be defined in `parsing as JSON`, but I'm curious.

s/Object object/valid JSON object/

02_commands.html
```diff
@@ -1,206 +1,524 @@
...
+        </ol>
+
+        <p>Otherwise, let <var>parameters</var> be null</p>
+      </li>
+
+      <li><p>Let <var>response data</var> be a <a>command response</a> object obtained by running running the steps defined by <var>command</var> with arguments <var>element id</var> and <var>parameters</var>.</p></li>
```
andreastt: twice mention of “running”

02_commands.html
```diff
@@ -1,206 +1,524 @@
...
+
+      <li><p>If <var>request</var>'s <a href="https://fetch.spec.whatwg.org/#concept-request-method" class="externalRef">method</a> is POST:</p>
+        <ol>
+
+          <li><p>Let <var>success</var> and <var>parameters</var> be the result of <a>parsing as JSON</a> with <var>request</var>'s <a href="https://fetch.spec.whatwg.org/#concept-request-body" class="externalRef">body</a> as the argument.</p></li>
+
+          <li><p>If <var>success</var> is false, or <var>success</var> is true and <var>parameters</var> is not an Object object, <a>return an error</a> with status <a>error code</a> <a>invalid argument</a> and jump back to step 1 in this overall algorithm.</p></li>
+        </ol>
+
+        <p>Otherwise, let <var>parameters</var> be null</p>
```
andreastt: Missing full stop.

02_commands.html
```diff
@@ -1,206 +1,524 @@
...
+      </li>
+
+      <li><p>Let <var>response data</var> be a <a>command response</a> object obtained by running running the steps defined by <var>command</var> with arguments <var>element id</var> and <var>parameters</var>.</p></li>
+
+      <li><p>If <var>response data</var> is an <a>error</a>, <a>return an error</a> with <a>error code</a> equal to <var>response data</var>'s <a>error code<a>.</p>
+
+        <p>Otherwise <a>send a response</a> with status 200 and <var>response data</var>'s data.</p></li>
+
+      <li><p>Jump to step 1</p></li>
```
andreastt: Missing full stop.

02_commands.html
jgraham: Oh, right, I should rewrite this in terms of an algorithm that can return success or failure.

But yeah, the idea was that `success` is a bool indicating whether the attempt to parse the body as JSON succeeded.

02_commands.html
andreastt: Drop “object” from “connection object”; I think we can just talk about this thing that is a read-writer stream.

02_commands.html
```diff
@@ -1,206 +1,524 @@
...
+
+      <li><p>Let <var>request match</var> be the result of <a>matching a request</a> with <var>request</var>'s <a href="https://fetch.spec.whatwg.org/#concept-request-method" class="externalRef">method</a> and <a href="https://fetch.spec.whatwg.org/#concept-request-url" class="externalRef">url</a> as arguments.</p></li>
+
+      <li><p>If <var>request match</var> is of type <a>error</a>, <a>return an error</a> with <var>request match</var>'s <a>error code</a> and jump to step 1.</p>
+
+        <p>Otherwise, let <var>command</var>, <var>session id</var> and <var>element id</var> be <var>request match</var>'s data.</p></li>
+
+      <li><p>If <var>command</var> is not <a>new session</a>:</p>
+        <ol>
+          <li><p>If <var>session id</var> is not equal to the <a>id</a> of any session in the <a>list of active sessions</a>, <a>return an error</a> with <a>error code</a> <a>invalid session id</a>, then jump to step 1.</p>
```
andreastt: “jump to the outer step 1” or something, because this is step 1.

02_commands.html
andreastt: Maybe we should merge `element ID` into `parameters` iff it's defined after step 3 in the overall algorithm.  That will give us one parameters map to pass as argument to `command`.

02_commands.html
```diff
@@ -1,206 +1,524 @@
...
+
+          <li><p>If <var>success</var> is false, or <var>success</var> is true and <var>parameters</var> is not an Object object, <a>return an error</a> with status <a>error code</a> <a>invalid argument</a> and jump back to step 1 in this overall algorithm.</p></li>
+        </ol>
+
+        <p>Otherwise, let <var>parameters</var> be null</p>
+      </li>
+
+      <li><p>Let <var>response data</var> be a <a>command response</a> object obtained by running running the steps defined by <var>command</var> with arguments <var>element id</var> and <var>parameters</var>.</p></li>
+
+      <li><p>If <var>response data</var> is an <a>error</a>, <a>return an error</a> with <a>error code</a> equal to <var>response data</var>'s <a>error code<a>.</p>
```
andreastt: I'm a little unclear on what it means to _be_ an `error` here, but it reads nice in English.  E.g. better than saying if `response data` has `error code` defined, then take that an map it to the appropriate `error code`.

andreastt: `return an error` → `send an error` to match `send a response`?

02_commands.html
```diff
@@ -1,206 +1,524 @@
...
+
+        <p>Otherwise <a>send a response</a> with status 200 and <var>response data</var>'s data.</p></li>
```
andreastt: I'm assuming we're doing the mapping of HTTP status code for sending errors in `return an error`, i.e. look up the error code, get the status code associated with it, and assign that to the response's HTTP status.

If we do that, I don't think we should mention status 200 explicitly here.  If it's a non-error response, it's implicit that it's status 200 and we could define that in `send a response`.

02_commands.html
```diff
@@ -1,206 +1,524 @@
...
+
+      <li><p>Jump to step 1</p></li>
+    </ol>
+
+    <p>When required to <dfn>send a response</dfn>, with arguments <var>status</var> and <var>data</var>, a remote end must run the following steps:</p>
```
andreastt: Same here, isn't it implicit that a response is always 200?  Hence the `status` variable is unnecessary.

02_commands.html
```diff
@@ -1,206 +1,524 @@
...
+
+    <ol>
+      <li><p>Let <var>response</var> be a new <a href="https://fetch.spec.whatwg.org/#concept-response" class="externalRef">response</a>.</p></li>
```
andreastt: I'm curious here whether we have to say anything about mutability or not.  I see other specs using prose like “initially left empty” when talking about objects that are created but which can be modified later.

02_commands.html
```diff
@@ -1,206 +1,524 @@
...
+
+      <li><p>Set <var>response</var>'s <a href="https://fetch.spec.whatwg.org/#concept-response-status" class="externalRef">status</a> to <var>status</var>, and <a href="https://fetch.spec.whatwg.org/#concept-response-status-message" class="externalRef">status message</a> to the string corresponding to the description of <var>status</var> in the <a href="http://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml" class="externalRef">status code registry</a>.</p></li>
+      <li><p>If <var>data</var> is not null, let <var>response</var>'s <a href="https://fetch.spec.whatwg.org/#concept-response-body" class="externalRef">body</a> be the result of <a>serializing as JSON</a> with <var>data</var> as the argument.</p></li>
+
+      <li><p>Serialize <var>response</var> according to the rules in [[!RFC7230]] and write the resulting bytes to the <a>connection</a></p></li>
```
andreastt: Missing full stop.

andreastt: Split this into two steps: First we serialise the HTTP response, then we write it to `connection`.

02_commands.html
```diff
@@ -1,206 +1,524 @@
...
+
+    </ol>
+
+    <p>When required to <dfn>return an error</dfn>, with <var>error code</var>, run the following steps:</p>
```
andreastt: s/return an error/send an error/

02_commands.html
```diff
@@ -1,206 +1,524 @@
...
+
+    <ol>
+      <li><p>Let <var>http status</var> and <var>name</a> be the <a>error response data</a> for <var>error code</var></p></li>
```
andreastt: Should reuse the response concept used in `send response` from the Fetch spec?

02_commands.html
andreastt: FWIW I think it would make more sense if `return an error` came before `send a response`.

02_commands.html
andreastt: `name` is undefined here.  The table mentions (error code, HTTP status, JSON status).

`<var>` should be closed with `</var>`.

Missing full stop.

02_commands.html
```diff
@@ -1,206 +1,524 @@
...
+      <li><p>Let <var>data</var> be a <a>new JSON Object</a>.</p>
+
+      <li><p>Let <var>message</var> be an implementation-defined string containing a human-readable description of the reason for the error.</p></li>
```
andreastt: Should be more concrete here about how we map that.  I'm assuming you mean the “JSON status” from the table?

02_commands.html
```diff
@@ -1,206 +1,524 @@
...
+
+        <!-- TODO: really need a better way of constructing JSON literals -->
+      <li><p><a>Set the properties</a> of <var>object</var> with values ("error", <var>name</var>), ("message", <var>message</var>).</p></li>
```
andreastt: `object` is undefined.  `data`?

02_commands.html
```diff
@@ -1,206 +1,524 @@
...
+
+      <li><a>Send a response</a> with status <var>status</var> and data <var>data</var></li>
```
andreastt: “Send a response with status and data as arguments” I think would be better.

Missing full stop.

02_commands.html
```diff
@@ -1,206 +1,524 @@
...
+    </ol>
+
+  </section>
+
+  <section>
+    <h3>Routing Requests</h3>
+
+    <!-- TODO: define what a command is -->
+
+    <p>Request routing is the process of going from a HTTP request to the series of steps needed to implement the <a>command</a> represented by that request.</p>
```
andreastt: `<dfn>Request routing</dfn>`?

02_commands.html
```diff
@@ -1,206 +1,524 @@
...
+
+    <p>A <a>remote end</a> may have a <dfn>url prefix</dfn>, which is a base url that is appended to the start of all WebDriver URLs. By default this is the empty string.</p>
```
andreastt: If it's an empty string by default, it can't be optional?

02_commands.html
```diff
@@ -1,206 +1,524 @@
...
+      <li><p>Let <var>command</var> by the command represented by the "command" column of <var>entry</var></p></li>
+
+      <li><p>Return <a>success</a> with data <var>session id</var>, <var>element id</var> and <var>command</var></li>
+
+    </ol>
+
+  </section>
+
+  <section>
+    <h3>List of endpoints</h3>
```
andreastt: s/List/Table/

02_commands.html
```diff
@@ -1,206 +1,524 @@
...
+    <p>A <a>remote end</a> may have a <dfn>url prefix</dfn>, which is a base url that is appended to the start of all WebDriver URLs. By default this is the empty string.</p>
+
+    <aside class="note">
+      <p>For example a remote end wishing to run alongside other services on <code>example.com</code> might set its <a>url prefix</a> to <code>/wd</code> so that a <a>new session</a> command would be invoked by sending a POST request to <code>/wd/session</code> rather than <code>/session</code>.</p>
+    </aside>
+
+    <p>In order to <dfn>match a request</dfn> given a <var>method</var> and <var>url</var>, the following steps must be taken:</p>
+
+    <ol>
+      <li><p>Let <var>endpoints</var> be a list containing each row in the <a>table of endpoints</a>.</p></li>
```
andreastt: “Lists” have a lot of implementation connotations.  Maybe “sequence” or “array” would be better.

02_commands.html
```diff
@@ -1,206 +1,524 @@
...
+
+   <li><p>Remove each entry in <var>endpoints</var> for which the concatenation of the <a>url prefix</a> and the production in the "url" column does not match <a>url</a> according to the rules in [[!URI-Template]].</p></li>
+
+      <li><p>If there are no entries in <var>endpoints</var>, return an <a>error</a> with code <a>unknown command</a></p></li>
```
andreastt: Missing full stop.

02_commands.html
```diff
@@ -1,206 +1,524 @@
...
+    <aside class="note">
+      <p>For example a remote end wishing to run alongside other services on <code>example.com</code> might set its <a>url prefix</a> to <code>/wd</code> so that a <a>new session</a> command would be invoked by sending a POST request to <code>/wd/session</code> rather than <code>/session</code>.</p>
+    </aside>
+
+    <p>In order to <dfn>match a request</dfn> given a <var>method</var> and <var>url</var>, the following steps must be taken:</p>
+
+    <ol>
+      <li><p>Let <var>endpoints</var> be a list containing each row in the <a>table of endpoints</a>.</p></li>
+
+   <li><p>Remove each entry in <var>endpoints</var> for which the concatenation of the <a>url prefix</a> and the production in the "url" column does not match <a>url</a> according to the rules in [[!URI-Template]].</p></li>
```
andreastt: Just curious, is this to filter out invalid URIs?  I.e. if `url prefix` contains a sequence of characters that's invalid according to URI Template?

02_commands.html
```diff
@@ -1,206 +1,524 @@
...
+      <li><p>Remove each entry in <var>endpoints</var> for which the "method" column is not an exact case-sensitive match for <var>method</var>.</p></li>
+
+      <li><p>If there are no entries in <var>endpoints</var>, return an <a>error</a> with code <a>unknown method</a>.</p></li>
+
+      <li><p>There is now exactly one entry in <var>endpoints</var>; let <var>entry</var> be this entry.</p></li>
+
+
+      <!-- TODO: Check that URI-templates is being used in the right way here -->
+
+      <li><p>If the match in step 2 populated a template variable called "sessionId", let <var>session id</var> be the value it was populated with. Otherwise let <var>session id</var> be null.</p></li>
```
andreastt: Can't see that step 2 populates anything.

02_commands.html
```diff
@@ -1,206 +1,524 @@
...
+
+      <li><p>If the match in step 2 populated a template variable called "elementId", let <var>element id</var> be the value it was populated with. Otherwise let <var>element id</var> be null.</p></li>
+
+      <li><p>Let <var>command</var> by the command represented by the "command" column of <var>entry</var></p></li>
```
andreastt: s/by/be/

02_commands.html
```diff
@@ -1,206 +1,524 @@
...
+
+      <li><p>Return <a>success</a> with data <var>session id</var>, <var>element id</var> and <var>command</var></li>
```
andreastt: It's a bit unclear to me the return types of these functions.  You alluded at them in a comment earlier, but I think it should be made clearer.

Missing full stop.

02_commands.html
```diff
@@ -1,206 +1,524 @@
...
+
+    </ol>
+
+  </section>
+
+  <section>
+    <h3>List of endpoints</h3>
+    <p>The following <dfn>table of endpoints</dfn> lists the method, url and command for each WebDriver command.</p>
```
andreastt: s/url/URL/

02_commands.html
```diff
@@ -1,206 +1,524 @@
...
+        <td>GET</td>
+        <td>/session/{sessionId}/screenshot</td>
+        <td><a>Take Screenshot</a></td>
+      </tr>
+    </table>
+  </section>
+
+  <section>
+    <h3>Handling Errors</h3>
+    <p>Errors are represented in the WebDriver protocol with a HTTP response with a HTTP status in the 4xx or 5xx range and a JSON body containing details of the error. This JSON body has two fields; <code>error</code>, containing a string indicating the error type and <code>error</code> containing an implementation-defined string containing a human readable description of the kind of error that occured.</p>
```
andreastt: Link to Fetch spec.

02_commands.html
```diff
@@ -1,206 +1,524 @@
...
+
+    <aside class="example">
+      <p>A <code>DELETE</code> request to <code>/session/1234</code>, where <code>1234</code> is not the id of a currently active session would return a HTTP <code>404</code> response with a body of the form:</p>
```
andreastt: `<a rel="current active session">currently active session</a>`

andreastt: s/would/might/ because the `error` string is implementation specific?

02_commands.html
```diff
@@ -1,206 +1,524 @@
...
+        <pre class="highlight">
+{
+  "status": "invalid session id",
+  "error": "No active session with id 1234"
+}</pre>
+    </aside>
+
+    <p>The following table lists each <dfn>error code</dfn>, its associated HTTP status, JSON status values, and gives a non-normative description of the error. The <dfn>error response data</dfn> for a particular <a>error code</a> is the values of the HTTP Status and JSON Status columns for the row corresponding to that <a>error code</a>.</p>
```
andreastt: s/JSON status values/JSON status value/

02_commands.html
```diff
@@ -1,206 +1,524 @@
...
+        <td>400</td>
+        <td><code>invalid element state</code></td>
         <td>An element command could not be completed because the element is in an invalid state (e.g. attempting to click an element that is no longer attached to the Document).</td>
       </tr>
       <tr>
-        <td><dfn id="status-invalid-selector">invalid selector</dfn></td>
+        <td><dfn>invalid selector</dfn></td>
+        <td>400</td>
+        <td><code>invalid selector</code></td>
         <td>Argument was an invalid selector (e.g. XPath/CSS).</td>
```
andreastt: Think we should remove the example here.  I'm confused by it.

02_commands.html
jgraham: So the idea is supposed to be defined earlier, but not very clearly. It is really a prose representation of something like Rust's Result type, with a definition like
```
enum Result<T> {
    Error(ErrorCode),
    Success(T)
}
```
and every algorithm that can fail returning a Result that can either be an Error or a Success.

02_commands.html
jgraham: Some data of type string, the exact content of which is not specified but is up to the implementation. How should I make that clearer?

02_commands.html
jgraham: It's supposed to filter out URLs that don't represent WebDriver commands.

02_commands.html
andreastt: No I think this is fine actually.  I misunderstood the context at first pass.

However, is this string required?  What happens if it's not defined?

02_commands.html
andreastt: I like that, because it reads clearly in English.  Let's do it.

02_commands.html
jgraham: Right, so I'm not clear on how to say what the return type is, exactly. It's a tuple of (session id, element id, command) wrapped in "Success". "command" is a reference to the command being run, so it's like a function pointer or something rather than a type containing a pure value.

02_commands.html
jgraham: I think that's a pretty bad idea; if you literally put it in the JSON structure then it adds the possibility of a clash with some real key, if you don't then it's just like this solution but expressed slightly differently.

02_commands.html
jgraham: I think the whole way we're using fetch is pretty dubious here, but it seems like a good plan to use it in a rough way and then work together with Anne to figure out if we can do better.

02_commands.html
jgraham: It's required (per the text above). Why wouldn't it be?

02_commands.html
jgraham: The whole algorithm isn't a sane implementation strategy, so if you try to copy it directly you're going to have a bad time. I'm not really sure why "list" is a worse word than any of the other possibilities.

02_commands.html
jgraham: Well the point is that in step 2 you take a URI template like /session/{sessionId} and match it against the incoming URL which will either match (and give {sessionId} a value, or not match in which case the row is filtered out). Here I want to know what actually matched in those cases.

02_commands.html
jgraham: I think this is OK? It's non-normative anyway and "of the form" was supposed to be the clue that the structure is important but the details aren't.

02_commands.html
andreastt: I guess we don't have to worry too much about immutability in our data structures since you can't have two WebDrivers at the same time.  How UAs choose to do their data structures is out of scope.  So please ignore my initial comment. 😊

02_commands.html
andreastt: I didn't say “worse”, but it's my feeling that lists might lead us to design algorithms that use common linked list operations when some implementations either can't easily do those or the semantics of their implementation varies from the intent of the spec.

But I'm happy to leave this as-is for now and we can revisit if that is/will be an actual problem.

02_commands.html
andreastt: I was looking at the wrong `li`.

andreastt:
This looks good enough to land now, I feel.  @AutomatedTester, do you want to take a look?

AutomatedTester:
landed in 2ceb655f9b8887a778ecb7bf6f507bf299d2d140

PR #17, opened 2015-03-27
Rewrite getCookies to be more specific and allow getting a cookie by nam...

AutomatedTester:
...e

The change allows us to getCookies and getCookies(name) and return them to the remote end. The notable change is the change from
GET to POST to allow a body that can be parsed to get the cookie name

### DIFF
--- a/01_introduction.html
+++ b/01_introduction.html
@@ -114,7 +114,7 @@
     from [[!ECMA-262]] with the exception that non-valid integer
     or float return values are returned as 0 or 0.0.
 
-    <h4>Top Browsing content no longer open</h4>
-    <p>A top browsing context is <dfn>no longer open</dfn> if it has been <a href="https://html.spec.whatwg.org/#a-browsing-context-is-discarded">discarded</a>
+    <h4>Top Level Browsing Context no longer open</h4>
+    <p>A top level browsing context is <dfn>no longer open</dfn> if it has been <a href="https://html.spec.whatwg.org/#a-browsing-context-is-discarded">discarded</a>
   </section>
 </section>
--- a/01_introduction.html
+++ b/01_introduction.html
@@ -64,7 +64,7 @@
    <h3>Terminology</h3>
 
    <p>In equations, all numbers are integers,
-    subtraction is represented by “−”, and bitwise OR by “|”. 
+    subtraction is represented by “−”, and bitwise OR by “|”.
     The characters “(” and “)” are used to provide logical grouping in these contexts.
   </section>
 
@@ -113,5 +113,8 @@
     and <code><a href=http://www.ecma-international.org/ecma-262/5.1/#sec-15.1.2.3>parseFloat</a></code>
     from [[!ECMA-262]] with the exception that non-valid integer
     or float return values are returned as 0 or 0.0.
+
+    <h4>Top Browsing content no longer open</h4>
+    <p>A top browsing context is <dfn>no longer open</dfn> if it has been <a href="https://html.spec.whatwg.org/#a-browsing-context-is-discarded">discarded</a>
   </section>
 </section>
--- a/12_cookies.html
+++ b/12_cookies.html
@@ -11,7 +11,7 @@
     <p>When returning Cookie objects, the server SHOULD include all optional fields it is capable of providing the information for.</p>
     <section>
       <!-- getCookie() -->
-      <h3>getCookie()</h3>
+      <h3>getCookie</h3>
         <p>
           <table class="simple jsoncommand">
             <tr>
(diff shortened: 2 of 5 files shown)


01_introduction.html
```diff
@@ -113,5 +113,8 @@
     and <code><a href=http://www.ecma-international.org/ecma-262/5.1/#sec-15.1.2.3>parseFloat</a></code>
     from [[!ECMA-262]] with the exception that non-valid integer
     or float return values are returned as 0 or 0.0.
+
+    <h4>Top Level Browsing Context no longer open</h4>
+    <p>A top level browsing context is <dfn>no longer open</dfn> if it has been <a href="https://html.spec.whatwg.org/#a-browsing-context-is-discarded">discarded</a>
```
andreastt: Can't we just use the definition from HTML, that a browser context has been “discarded”?

12_cookies.html
```diff
@@ -38,20 +20,56 @@
...
-        </ul>
-        <dl class='parameters'>
-          <dt>optional DOMString? name</dt>
-          <dd>The name (e.g. <code>name=</code>) of the cookie that needs to be returned.</dd>
-        </dl>
+          <p>The <a>remote end steps</a> for the <a>getCookie</a> command are:</p>
+          <ol>
+            <li><p>If the <a>current browsing context</a> is <a>no longer open</a>, return an error with code <a>no such window</a>.</p></li>
+            <li><p>Let <var>name</var> be the result of <a>getting a property</a> named "name" from the <var>parameters</var> argument.</p></li>
+            <li><p>Let <var>result</var> be an empty JSON List</p></li>
```
andreastt: s/List/list/ and missing full period.

12_cookies.html
```diff
@@ -38,20 +20,56 @@
...
+            <li><p>Let <var>cookies</var> be a list of all cookies [[!RFC6265]] in the cookie store associated with <a href="https://dom.spec.whatwg.org/#concept-document-url"> the current document or frame address</a></p></li>
```
andreastt: Missing full stop.

12_cookies.html
```diff
@@ -38,20 +20,56 @@
...
+            <li><p>Let <var>length</var> be the length of <var>cookies</var>.</p></li>
+            <li><p>Let <var>k</var> be 0</p></li>
+            <li><p>While <var>k</var> < <var>length</var>: </p>
```
andreastt: s/</&lt;/

12_cookies.html
```diff
@@ -38,20 +20,56 @@
...
+              <ol>
+                <li><p>Let <var>cookie</var> be the value in <var>cookies</var> at index <var>k</var>.</p></li>
+                <li><p>If <var>name</var> is undefined:</p>
+                  <ol>
+                    <li><p>Append a <a>serialised cookie</a> to <var>result</var>.</p></li>
```
andreastt: s/serialised/serialized/

12_cookies.html
```diff
@@ -38,20 +20,56 @@
...
+                  </ol>
+                </li>
+                <li><p>if <var>name</var> is defined and is equal to <var>cookie-name</var>:</p>
```
andreastt: s/if/If/

12_cookies.html
```diff
@@ -38,20 +20,56 @@
...
+                  <ol>
+                    <li><p>Append a <a>serialised cookie</a> to <var>result</var>.</p></li>
+                  </ol>
+                </li>
+                <li><p>Increase the value of <var>k</var> by 1</p></li>
```
andreastt: This doesn't jump back to continue the loop, but just increases `k`.  I strongly suggest we just say “iterate through the list” and “for each `<var>cookie</var>`, do:”.

12_cookies.html
```diff
@@ -38,20 +20,56 @@
...
+              </ol>
+            </li>
+            <li><p>Return <a>success</a> with data <var>result</var></p></li>
```
andreastt: Missing full stop.

12_cookies.html
```diff
@@ -38,20 +20,56 @@
...
+          </ol>
+          <p>A <dfn>serialised cookie</dfn> is created with the following algorithm:</p>
+          <p>
+            <ol>
+              <li><p>Let <var>serialised cookie</var> be an empty map.</p></li>
```
andreastt: a map left initially empty

12_cookies.html
AutomatedTester: This follows the same format for `forEach` so will be keeping this as is. Adding `Go To Step x` is frowned upon, apparently, unless it is for highly complex loops.

12_cookies.html
AutomatedTester: in [rfc6265](http://tools.ietf.org/html/rfc6265#section-4.1.1) in the `cookie-pair` is `cookie-name = cookie-value` so would like to preserve some of the semantics by using `cookie-name`

AutomatedTester:
landed in 08382499cb775645b61b313be4fd441b72259b56

PR #19, opened 2015-04-01
Merge all sources to a single file

andreastt:


shs96c:
Why? This makes it a lot nastier to work with the sections.

andreastt:
My opinion is that it's not easy to work with the way the source files are currently organised because you never seem to know in which file a particular definition is referenced.  My experience is that it's easier to search and replace inside a single file.

It’s also hard to introduce new chapters because they are named by number and concatenated in an order given by your shell's glob.  This is an issue if you want to add a new chapter before an existing, or reorder the existing chapters (as renaming files isn't great for tracking history).

The build system for capabilities is overly complex and didn’t reflect the reality in the prose, and the mappings include wasn’t used at all.

My argument is that a single canonical file is easier to understand because of the lack of any “magic” involved in the build step.  No build step also means we can safely ignore all the times PRs forget to regenerate the `webdriver-spec.html` file.

shs96c:
Generating the webdriver-spec.html isn't too onerous, and I find the fact that we can generate chunks of the spec is enormously useful (eg. we should be generating section 2.6 (the list of endpoints) rather than attempting to keep more than one piece of information up to date) If we're not being efficient with our usage of it, then we're missing out.

My workflow at the moment is to open the entire directory in a text editor. Search and replace then works consistently across all the files.

I do agree that renumbering the files is a PITA, but git handles moves efficiently, and we seldom add new sections, so it's never been a problem (particularly with a little command line scripting).

jgraham:
We seldom add or remove sections because it's a PITA.

The current structure of the spec is rather broken because of this.

andreastt:
1. As an example `03_capabilities.html` is completely empty because reordering sections across files is unnecessarily hard.  The main reason I'm pushing back on this is because I need to add a new chapter about technical model/terminology closer to the start so that the spec reads more naturally.  This isn't possible right now unless I just add it to `02_commands.html` (which isn't primarily about commands…), like with all other changes that have been made recently.

2. File numbering is just another thing which makes this inconvenient.  We could eliminate this aspect by defining the explicit ordering in `Makefile`, but it would still mean you'd have to modify the build steps to reorder some content.  It also doesn't encourage restructuring through iterative improvements.

3. Please let's relate to text editing practices without the need to follow a special workflow or using a specific editor.  It _is_ cumbersome the way the build step riddles the directory with temporary files which get picked up by grep unless you limit it by a specific file syntax.  Search and replace (either by sed or some editor built-in functionality) doesn't work if you just open every file in the directory because you risk changing `webdriver-spec.html` or any of the other temporary files.

4. I also can't tell which generated chunks are “enormously useful”.  As far as I can tell `capabilities.html`, `command.html`, `caps.fragment`, `appendix-mapping-header.html`, and `appendix-mapping-footer.html` aren't used. And `appendix-commandformat-footer.html`, `appendix-commandformat-header.html` are scheduled to be removed.  That means I can't find a single use case that isn't confusing to people editing the spec as of right now.

5. Examining sections 2 and 3 I can't tell what of that needs to be generated elsewhere in the text.  You say we should generate section 2.6.  Where does it need to be repeated?

AutomatedTester:
This is something that I think we should do however I want to clear down the PR queue and do it then, It should be trivial to redo this at that time.

Closing this for now.

PR #24, opened 2015-04-09
Neither one of "visibility", "interactibility" implies the other.

sevaseva:
https://www.w3.org/Bugs/Public/show_bug.cgi?id=28407 is related (but I think it makes a falsy statement where it uses logical 'weaker').

### DIFF
--- a/09_elements.html
+++ b/09_elements.html
@@ -481,9 +481,6 @@
     <li>Return true.
    </ol>
 
-   <p>If the steps above return true, then the element MUST also
-    be <a href=#interactable-elements>interactable</a> should it
-    meet the other criteria for being interactable.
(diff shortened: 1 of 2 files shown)


andreastt:
Can probably be replaced by #25 or the other way around.

sevaseva:
I'll cherry-pick that commit to this PR later today

sevaseva:
Actually.. why not consider/submit deletion of these two sentences separately. There are different reasons for deleting them. The two changes may meet separate discussions, too...

andreastt:
I'm sure that's OK.

AutomatedTester:
landed in 3eb3dd3

PR #25, opened 2015-04-09
Remove a sentence that has unclear purpose and other issues.

sevaseva:
Other issues:
- 'any part of the BODY' is not specific enough?
- 'can be brought into the current viewport' allows (I believe, unintentionally) doing anything in order to bring the thing into the viewport, e.g. modifying current page's DOM.


andreastt:
I think probably this whole paragraph should go.

AutomatedTester:
landed in 1ceee40

PR #29, opened 2015-04-16
Specifying name of capability for page load strategy on session initiali...

jimevans:
...zation

### DIFF
--- a/05_navigation.html
+++ b/05_navigation.html
@@ -8,7 +8,7 @@
   <section>
     <h3 id="page-load-strategies">The Page Load Strategy</h3>
 
-    <p><a>Local ends</a> may choose at what point during the page load cycle the remote end returns a response to navigation commands by requesting a particular <dfn>page load strategy</dfn> when creating a <a>new session</a>. The available page load strategies are:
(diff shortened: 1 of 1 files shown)


andreastt:
That the session has a **page loading strategy** state associated with it is already described under the model of [Session](https://w3c.github.io/webdriver/webdriver-spec.html#sessions).

The description of [newSession](https://w3c.github.io/webdriver/webdriver-spec.html#newsession) is completely wrong however, and that's where we should say to look for the `pageLoadStrategy` capability and assign it to the **page load strategy** state of the session we're creating.

jimevans:
I honestly don't care where the thing gets defined. I care that the capability name is strictly specified. Feel free to do with this whatever you see fit, close it, rework it, whatever.

andreastt:
Yes, we need this. But looking at the state [newSession](https://w3c.github.io/webdriver/webdriver-spec.html#newsession) is in, I wonder if it's best to rework that first because it's not currently considering _any_ capabilities.

I know @AutomatedTester is working on it, do you have an ETA on that?

AutomatedTester:
I am currently working on capabilites and this would part of that. Thanks for your time in creating this PR but I am going to close it as it's already part of my capabilities patch.

PR #30, opened 2015-04-17
Reorganise content near start

andreastt:
This patch reorganises chapters and sections near the start of the document.  Specifically, it:

* Reduces number of sections
* Moves terminology definitions to the **Terminology** chapter
* Merges the mention of references to other specs under **Conformance**
* Removes **normalizing style pixel values by integer** as it was not in use
* Moves the **normalizing style pixel values by float** to **Element State** (where it's used)
* Deletes the first section defining **Local** and **Remote** in favour of the section that comes a little bit later that is much better
* Removes the empty **Common Infrastructure** section under **The WebDriver Protocol** and replaces it with **Algorithms**
* Moves the definition of **no longer open** to the **Sessions** chapter, where top-level browsing context is explained and where the term is used


00_header.html
```diff
@@ -100,5 +100,5 @@
...
+<p>This specification is intended for implementors of the WebDriver API. It is not intended as light bed time reading. This specification is still under active development and may not be stable. Any implementors who are not actively participating in the preparation of this specification may find unexpected changes occurring. It is suggested that any implementors join the WG for this specification. Despite not being stable, it should be noted that this specification is strongly based on an existing Open Source project &mdash; <a href="http://selenium.googlecode.com/">Selenium WebDriver</a> &mdash; and the existing implementations of the API defined within that project.</p>
```
AutomatedTester: We should update the OSS project link to the new github link

AutomatedTester:
landed in 7aa91c7

PR #31, opened 2015-04-28
Add stacktrace field to error responses

andreastt:
According to [discussion on mailing list](https://lists.w3.org/Archives/Public/public-browser-tools-testing/2015AprJun/0008.html).


AutomatedTester:
landed in 5960810

PR #32, opened 2015-04-28
Documenting the process of handling capabilities

AutomatedTester:
First draft, I expect I have missed something but want to get this out there to start getting comments

[Review on Reviewable](https://reviewable.io/reviews/w3c/webdriver/32)

PR #34, opened 2015-04-29
reword new Session to be more precise

AutomatedTester:
[Review on Reviewable](https://reviewable.io/reviews/w3c/webdriver/34)

### DIFF
--- a/04_sessions.html
+++ b/04_sessions.html
@@ -96,7 +96,7 @@ an <a href="https://html.spec.whatwg.org/#ancestor-browsing-context">ancestor
  requires passing a <a>session ID</a>.
 
   <section>
-    <h3>newSession()</h3>
+    <h3>New Session</h3>
     <p><table class="simple jsoncommand">
       <tr>
         <th>HTTP Method</th>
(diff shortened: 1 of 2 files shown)


04_sessions.html
```diff
@@ -109,89 +109,36 @@
...
-      possible.
-      How the new session is created depends on the implementation of this specification.
-        <ul>
-          <li>If any of the "requiredCapabilities" cannot be fulfilled by the new session, the remote end MUST quit the session and return the
-            <code><a href="#status-session-not-created">session not created</a></code> error code. The <var>sessionId</var> MUST be left as the default value unless one was set in the initial request, in which case that value MUST be used. The error message SHOULD list all unmet required capabilities though only a single required capability MUST be given.</li>
-          <li>If a capability is listed in both the requiredCapabilities and desiredCapabilities, the value in the requiredCapabilities MUST be used.</li>
-          </ul>
+      <li>
+        <p>
+         Let <var>resultant capabilities</var> be the result of calling <a>process capabilities</a>
```
AutomatedTester: `process capabilities` is documented in #32

04_sessions.html
```diff
@@ -96,7 +96,7 @@
  requires passing a <a>session ID</a>.
 
   <section>
-    <h3>newSession()</h3>
+    <h3>newSession</h3>
```
andreastt: New Session because it's a header and we don't use a command list any longer.  This will match the reference in the HTTP mapping table above.

04_sessions.html
andreastt: Missing full stop at end of sentence.

04_sessions.html
```diff
@@ -109,89 +109,36 @@
...
-      How the new session is created depends on the implementation of this specification.
-        <ul>
-          <li>If any of the "requiredCapabilities" cannot be fulfilled by the new session, the remote end MUST quit the session and return the
-            <code><a href="#status-session-not-created">session not created</a></code> error code. The <var>sessionId</var> MUST be left as the default value unless one was set in the initial request, in which case that value MUST be used. The error message SHOULD list all unmet required capabilities though only a single required capability MUST be given.</li>
-          <li>If a capability is listed in both the requiredCapabilities and desiredCapabilities, the value in the requiredCapabilities MUST be used.</li>
-          </ul>
+      <li>
+        <p>
+         Let <var>resultant capabilities</var> be the result of calling <a>process capabilities</a>
+         and use <var>capabilities</var> as an argument. If a <a>session not created</a> is raised
```
andreastt: Better wording would be: `with <var>capabilities</var> as the argument.`

Also I think it would be better to say `if the result is an <a>error</a>, return an <a>error</a> with the code <a>session not created</a>.`

(See how [Get](https://w3c.github.io/webdriver/webdriver-spec.html#get) does it.)

04_sessions.html
```diff
@@ -109,89 +109,36 @@
...
-        <dt><code><a href="#status-timeout">timeout</a></code></dt>
-        <dd>The new session could not be created within the time allowed for command execution on the remote end. This
-        time MAY be infinite (in which case this status won't be seen). The "value" field of the Response SHOULD contain a string explaining that a timeout has
-        occurred, but it MAY be left as the null value or filled with the empty string.</dd>
-        <dt><code><a href="#status-unknown-error">unknown error</a></code></dt>
-        <dd>An unhandled error of some sort has occurred. The "value" field of the Response SHOULD contain a more detailed
-        description of the error as a string.</dd>
-      </dl>
+      <li><p>Let <var>session id</var> be the result of creating a new <a>session ID</a> and
+           appending to <a>list of active sessions</a>.</p></li>
```
andreastt: You need to define how to create a new session ID under Session, or we can do this later, but it should read something along the lines of:

```html
<li><p>Let <var>session id</var> be the result of <a>generating a UUID</a>.</p></li>
<li><p>Append <var>session id</var> to the <a>list of active sessions</a>.</p></li>
```

04_sessions.html
```diff
@@ -109,89 +109,36 @@
...
+      <li>
+         <p>
+           Let <var>data</var> be an empty JSON Object.
+         </p>
+      </li>
+      <li>
+         <p>
+           Let "sessionId" be an entry in <var>data</var> with the value of <var>session Id</var>.
```
andreastt: `"<code>sessionId</code>"`

04_sessions.html
```diff
@@ -109,89 +109,36 @@
...
+         </p>
+      </li>
+      <li>
+        <p>
+        Let "capabilities" be an entry in <var>data</var> with the value of <var>resultant capabilities</var>.
```
andreastt: Same here with respect to assigning a property on `data` rather than creating a variable in the local scope.

04_sessions.html
jgraham: You need to mention that this can fail e.g. if there's a session already running.

04_sessions.html
```diff
@@ -109,89 +109,36 @@
         <td></td>
       </tr>
     </table>
-    <p>The process for successfully creating a session MUST be:</p>
+    <p>The <a>remote end steps</a> for the <dfn>New Session</dfn> command are:</p>
```
jgraham: There should probably be a note here about intermediate nodes, or something, so that someone doesn't complain that this isn't how they work when they're distributing jobs. Basically the intermediate node can do anything it likes with the capabilities and typically just forwards the session id from the furthest remote.

AutomatedTester:
@jgraham @andreastt I have updated everything after your awesome comments

jgraham:
<sup>**[04_sessions.html, line 128 \[r1\]](https://reviewable.io:443/reviews/w3c/webdriver/34#-JoEgQDfLY2Z-U0oonAV)** ([raw file](https://github.com/w3c/webdriver/blob/48f8a03528278f3e293db69a99ee933859aa47ab/04_sessions.html#L128)):</sup>
You can't have normative languages in notes.

You can probably just make step 1 implementation-defined behaviour like:

```
"If the node is an intermediate node, take implementation-defined steps that either result in returning an error with status code <a>session not created</a> or in returning a <a>success</a> with data that is isomorphic to that returned by <a>remote ends</a> according to (some other steps in this algorithm).

<p class=note>This allows intermediary nodes to use the capabilities data in any way they want e.g. to select a specific browser to test based on a combination of the required and desired capabilities. Typically the new session response from the remote end selected in this process will then be  relayed directly to the client</p>"
```

---

<sup>**[04_sessions.html, line 133 \[r1\]](https://reviewable.io:443/reviews/w3c/webdriver/34#-JoEj0VsKNyp6tJh6KtX)** ([raw file](https://github.com/w3c/webdriver/blob/48f8a03528278f3e293db69a99ee933859aa47ab/04_sessions.html#L133)):</sup>
That doesn't sound right? If it's supposed to be a constraint that we can only have one session per remote end ever why is it a list of active sessions rather than just an active session that's initially null?

---

<sup>**[04_sessions.html, line 144 \[r1\]](https://reviewable.io:443/reviews/w3c/webdriver/34#-JoEjKOHTYNApeZsOquf)** ([raw file](https://github.com/w3c/webdriver/blob/48f8a03528278f3e293db69a99ee933859aa47ab/04_sessions.html#L144)):</sup>
I think something more is probably needed to ensure that the session is started with the right internal state e.g. timeouts.

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/w3c/webdriver/34)

jgraham:
Reviewed files:
- 04_sessions.html @ r1
- webdriver-spec.html @ r1

---





Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/w3c/webdriver/34)

AutomatedTester:
<sup>**[04_sessions.html, line 133 \[r1\]](https://reviewable.io:443/reviews/w3c/webdriver/34#-JoEj0VsKNyp6tJh6KtX)** ([raw file](https://github.com/w3c/webdriver/blob/48f8a03528278f3e293db69a99ee933859aa47ab/04_sessions.html#L133)):</sup>
My thinking was related to `intermediary nodes`. `Intermediary nodes` can keep a list of `session ids` but a remote end probably shouldn't allow more than one session at a time. I think if I do something based on your previous comment that should be good enough

---

<sup>**[04_sessions.html, line 144 \[r1\]](https://reviewable.io:443/reviews/w3c/webdriver/34#-JoEjKOHTYNApeZsOquf)** ([raw file](https://github.com/w3c/webdriver/blob/48f8a03528278f3e293db69a99ee933859aa47ab/04_sessions.html#L144)):</sup>
This is actually something I wanted to email the working group about. Some have suggested that capabilities be a set of immutable objects. I am of the opinion that it should be a way to set up the session and what is returned is what was set (either by the browser or by the user)

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/w3c/webdriver/34)

jgraham:
<sup>**[04_sessions.html, line 144 \[r1\]](https://reviewable.io:443/reviews/w3c/webdriver/34#-JoEjKOHTYNApeZsOquf)** ([raw file](https://github.com/w3c/webdriver/blob/48f8a03528278f3e293db69a99ee933859aa47ab/04_sessions.html#L144)):</sup>
Either way you need to actually initalise the session in the remote end, I think.

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/w3c/webdriver/34)

jgraham:
<sup>**[04_sessions.html, line 133 \[r1\]](https://reviewable.io:443/reviews/w3c/webdriver/34#-JoEj0VsKNyp6tJh6KtX)** ([raw file](https://github.com/w3c/webdriver/blob/48f8a03528278f3e293db69a99ee933859aa47ab/04_sessions.html#L133)):</sup>
https://w3c.github.io/webdriver/webdriver-spec.html#sessions already defined a maximum number of sessions. So basically you just check if the length of the list of sessions is equal to the maximum number of sessions and, if it is, error, otherwise create the session.

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/w3c/webdriver/34)

andreastt:
<sup>**[04_sessions.html, line 116 \[r1\]](https://reviewable.io:443/reviews/w3c/webdriver/34#-JojLOWIHgnL5jPPDmrF)** ([raw file](https://github.com/w3c/webdriver/blob/48f8a03528278f3e293db69a99ee933859aa47ab/04_sessions.html#L116)):</sup>
`"<code>capabilities</code>"` for consistency with the rest of the document.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/w3c/webdriver/34)

andreastt:
<sup>**[04_sessions.html, line 121 \[r1\]](https://reviewable.io:443/reviews/w3c/webdriver/34#-JojM3ZaI5nqWMtWUjuE)** ([raw file](https://github.com/w3c/webdriver/blob/48f8a03528278f3e293db69a99ee933859aa47ab/04_sessions.html#L121)):</sup>
I think we should rename `process capabilities` to `processing capabilities` so the prose flows better:

> Let `resultant capabilities` be the result of processing capabilities `capabilities`.

---

<sup>**[04_sessions.html, line 128 \[r1\]](https://reviewable.io:443/reviews/w3c/webdriver/34#-JoEgQDfLY2Z-U0oonAV)** ([raw file](https://github.com/w3c/webdriver/blob/48f8a03528278f3e293db69a99ee933859aa47ab/04_sessions.html#L128)):</sup>
This isn't resolved AFAICT? (I don't know how to un-resovle in Reviewable.)

---

<sup>**[04_sessions.html, line 128 \[r1\]](https://reviewable.io:443/reviews/w3c/webdriver/34#-JojMisYXYSKVw3TS35o)** ([raw file](https://github.com/w3c/webdriver/blob/48f8a03528278f3e293db69a99ee933859aa47ab/04_sessions.html#L128)):</sup>
s/Remote End/remote end/

---

<sup>**[04_sessions.html, line 144 \[r1\]](https://reviewable.io:443/reviews/w3c/webdriver/34#-JoEjKOHTYNApeZsOquf)** ([raw file](https://github.com/w3c/webdriver/blob/48f8a03528278f3e293db69a99ee933859aa47ab/04_sessions.html#L144)):</sup>
Most of the properties associated with session have a default, so I think most of them should be fine left untouched, but we definitely need to say something along the lines of creating a new instance of session and setting its `session id` property.

---

<sup>**[04_sessions.html, line 147 \[r1\]](https://reviewable.io:443/reviews/w3c/webdriver/34#-JojNf0iFwIXpdoyjP4F)** ([raw file](https://github.com/w3c/webdriver/blob/48f8a03528278f3e293db69a99ee933859aa47ab/04_sessions.html#L147)):</sup>
s/\./:/

---

<sup>**[04_sessions.html, line 152 \[r1\]](https://reviewable.io:443/reviews/w3c/webdriver/34#-JojNiy6IYdYYO8NHD7E)** ([raw file](https://github.com/w3c/webdriver/blob/48f8a03528278f3e293db69a99ee933859aa47ab/04_sessions.html#L152)):</sup>
s/Id/id/

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/w3c/webdriver/34)

AutomatedTester:
landed in 6b0b8151743cab8cabbcae251c53c2d5216d55eb

PR #35, opened 2015-05-02
Harmonise titles for commands on elements

andreastt:
Most commands related to elements use the word “element” in the title
to indicate that the operation they perform happen on the element.

This patch updates the following command titles:

Is Displayed -> Is Element Displayed
Is Selected -> Is Element Selected
Get CSS Value -> Get Element CSS Value
Is Enabled -> Is Element Enabled

### DIFF
--- a/02_commands.html
+++ b/02_commands.html
@@ -144,10 +144,11 @@
   </section>
 
   <section>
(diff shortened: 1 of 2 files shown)


AutomatedTester:
landed in 7c424e3

PR #36, opened 2015-05-02
Fix Get example and punctuation

andreastt:
The example related to command Get was incorrect, and fixed two cases
of punctuation.

### DIFF
--- a/05_navigation.html
+++ b/05_navigation.html
@@ -26,50 +26,80 @@
     In addition, implementors MAY add additional page loading strategies.</p>
   </section>
 
-  <section>
-    <h3>get</h3>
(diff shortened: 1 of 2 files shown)


AutomatedTester:
The changes to the white space at the beginning of each line makes it quite difficult to view the actual changes. What has actually changed?

andreastt:
Sorry, I should've pointed that out:

- Removes padding in highlighted example of JSON body
- Corrects the `url` field's value by adding a slash near the end to match the prose in the JSON body
- Adds a full stop at the last point in the algorithm.

Other than that, no changes were made.

AutomatedTester:
landed in 9db6434

PR #38, opened 2015-05-02
Fix current top-level browsing context

andreastt:
When HTML refers to “top-level browsing context” (with hyphen) we
should too.


AutomatedTester:
landed in d2a72b5

PR #40, opened 2015-05-06
Specify Get Window Handle

andreastt:
Specifies getting a window's handle.  I'm a bit unsure if I got the relationships between **current browsing contexts** and **current top-level browsing contexts** right.  Also unsure about the section on **creating a window handle** for a given browsing context, because we need to ensure that it's unique across _all_ browsing contexts; not just top-level browsing contexts since they may be the same.


AutomatedTester:
landed in 0563edc

PR #42, opened 2015-05-06
Rename command Close to Close Window

andreastt:
Gives a nicer consistency because it's not obvious what “Close” by itself means.  This matches the other window control command names.


AutomatedTester:
landed in 9202965

PR #44, opened 2015-05-06
Define infrastructure for web elements

andreastt:
This defines the model for dealing with web element references in WebDriver, improving on what's there already.

There are areas here that I'm uncertain about, especially with regards to the **known elements** table being associated with the **top-level browsing context**, which isn't really a concept defined in WebDriver.

The implementation in Marionette of this table is (roughly) tied to the lifetime of the session.  But advantage of tying it to the browser context is that it it's more specific.

Also maybe it should be **browser context** instead of **top-level browsing context** since you can only access elements in your current context?

### DIFF
--- a/09_elements.html
+++ b/09_elements.html
@@ -1,87 +1,100 @@
 <section>
 <h2>Elements</h2>
 
-<p>A <dfn>web element</dfn> is an abstraction of a document element
- used for 
+<p>A <dfn>web element</dfn> is an abstraction over
+ a [[!DOM4]] <a href=https://dom.spec.whatwg.org/#concept-element>document element</a>
+ used to reference elements
+ over the <a href=#the-webdriver-protocol>WebDriver protocol</a>.
 
-<p>A <a>web element</a> has an associated <dfn>DOM element</dfn>,
- a reference to a [[!DOM4]] <a href=http://w3c.github.io/dom/#element>Element</a>.
+<p>When <a>web elements</a> are transported
+ across the <a href=#the-webdriver-protocol>protocol</a>
+ to the <a>local end</a>,
+ they are indentified by the <a>web element reference</a> associated with them.
+ The reference is used as part of the arguments
+ to commands relevant to operations
+ on <a href=https://dom.spec.whatwg.org/#concept-element>elements</a>.
+
+<p>A <a>web element</a> has an associated <dfn>document element</dfn>,
+ a reference to a <a href=https://dom.spec.whatwg.org/#concept-element>element</a>.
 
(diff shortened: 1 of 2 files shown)


09_elements.html
```diff
@@ -1,26 +1,88 @@
...
-  </p>
-
-  <div class="note">
-    <p>This requirement around <a>WebElement</a> identifiers allows for efficient equality
-    checks when the WebDriver API is being used out of process.</p>
-  </div>
+<h2>Elements</h2>
 
+<p>A <dfn>web element</dfn> is an abstraction of a document element
+ used for
```
AutomatedTester: used for?

09_elements.html
```diff
@@ -1,26 +1,88 @@
...
+<p>A <a>web element</a> has an associated <dfn>DOM element</dfn>,
+ a reference to a [[!DOM4]] <a href=http://w3c.github.io/dom/#element>Element</a>.
+
+<p>A <a>web element</a> has an associated <dfn>web element reference</dfn>
+ (a UUID) used to uniquely identify the <a>web element</a>'s <a>DOM element</a>.
+
+<p>The <a>web element reference</a> for every <a>web element</a> representing
+ the same [[!DOM4]] <a href=http://w3c.github.io/dom/#element>Element</a> MUST be the same.
+
+<p class=note>Since multiple <a>web element</a> instances
```
AutomatedTester: While this is correct I am not sure if we need to put this in the specification since it's a guide on how things can be done on the local end

09_elements.html
```diff
@@ -1,26 +1,88 @@
...
+ have identical <a>web element reference</a>s,
+ efficient equality checks may be performed at the <a>local end</a>.
+
+<p>Each <a href=https://html.spec.whatwg.org/#top-level-browsing-context>top-level browsing context</a>
+ has an associated <dfn>known elements</dfn> table,
+ where <a>web element references</a> are mapped
+ to [[!DOM4]] <a href=http://w3c.github.io/dom/#element>Element</a>'s.
+
+<p>The <a>known elements</a> table is discarded when
+ the <a href=https://html.spec.whatwg.org/#top-level-browsing-context>top-level browsing context</a>
```
AutomatedTester: I think that you are right and we need to have this at a `browser context` instead of the `top-level browsing context` since that way, if things like `iframe`s are removed we can release things better

andreastt:
Pushed changes which I think makes this a lot better.  At least it’s a much better starting point than what is in the specification currently.

andreastt:
Also added a section about **serialising a web element**.

AutomatedTester:
landed in 92f4978

PR #45, opened 2015-05-07
Fix respec warning about multiple HTML51 definitions

andreastt:


AutomatedTester:
I had already done this locally and have landed in d1d88d8

PR #46, opened 2015-05-16
Rewrite the text about window handles and closing sessions.

jgraham:
This adds the behaviour that the Close Window command returns a
list of the remaining window handles.


AutomatedTester:
Spotted a few minor nits that I will fixup in later pushes. landed in c5df01b

PR #47, opened 2015-05-17
Section on non-HTML content edited for readability

zenlambda:


andreastt:
My feeling is that this entire section should be removed.

zenlambda:
@andreastt Is finding browser DOM objects generated by browsers not something webdriver implementors would want to support? I actually found it encouraging that it was mentioned in the draft spec.

I find it strange how, with various webdriver implementations (Firefox, Chrome, Selenium Server), I can do getPageSource() and get a load of html for a plain text document, but on the other hand a simple call to .findElement(By.cssSelector('body')) will hang until the timeout expires. 

On the other hand some projects like [nightwatch](https://github.com/beatfactor/nightwatch) will let you find DOM elements inserted by the browser.

andreastt:
@zenlambda It is, but this condition should be embedded in the remote end steps algorithm for **Find Element** and **Find Elements**, not as a separate point.

zenlambda:
@andreastt Ah. I think I see how extending the specification on **Find Element** and **Find Elements** to include elements that have been inserted automatically by the browser would be more precise. Is that a realistic prospect?

andreastt:
Sorry for the inconvenience that the specification is in such a high level of flux )-:  Please drop by `#webdriver` on irc.w3.org if you have any questions.

And thanks for engaging with us!

PR #50, opened 2015-05-19
Avoid redefinition of "script timeout"

andreastt:
"script timeout" was defined twice, as both an error and an associated
state to a session.


AutomatedTester:
landed in a59f6d4

PR #53, opened 2015-05-20
Define Get Element Rect and some web element infra

andreastt:


webdriver-spec.html
```diff
@@ -2259,23 +2296,55 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+  is not null:
+
+  <ol>
+   <li><p>Set <var>x</var> to (<var>x</var> +
+    <var>element</var>’s <a href=http://dev.w3.org/csswg/cssom-view/#dom-htmlelement-offsetleft>offsetLeft</a>).
+
+   <li><p>Set <var>y</var> to (<var>y</var> +
+    <var>element</var>’s <a href=http://dev.w3.org/csswg/cssom-view/#dom-htmlelement-offsettop>offsetTop</a>).
+
+   <li><p>Set <var>element</var> to <var>element</var>’s
```
AutomatedTester: what is `element` used for?

andreastt: It’s being used for iteration.  To simplify this with some pseudo code:

```
while (element.offsetParent !== null) {
  x += element.offsetLeft
  y += element.offsetTop
  element = element.offsetParent
```

Eventually `element`’s `offsetParent` will be `null` when it reaches the final element.

AutomatedTester:
landed in 5311795

PR #58, opened 2015-05-22
Rename "document element" to "web element node"

andreastt:
A document element is both misleading and untrue: It’s not the
document’s documentElement and it’s actually allowed to be a DOM
node, not just the element superset.


andreastt:
I think we want a subset of Node, e.g. DOMElements including Document and possibly Window?

andreastt:
@jgraham, I like your reverse approach of describing web element serialisations.  I’ll try to fix up this patch.

andreastt:
Closing in favour of #76.

PR #59, opened 2015-05-22
Remove status code handling on the local end

andreastt:
We decided not to talk so much about what the local end’s responsibilities are, as this is out of scope of the WebDriver implementations.  How the local ends choose to implement errors are entirely up to them.


AutomatedTester:
Landed in b04ec3e

PR #62, opened 2015-05-22
Add dfn for Switch To Window and correct links

andreastt:
Needs to be improved further later because we need to iterate list of browsing contexts.


AutomatedTester:
landed in 3fe1b1b

PR #74, opened 2015-06-02
Define Refresh command

andreastt:


andreastt:
FWIW we should consider dropping *Refresh* altogether as it offers nothing *Get* doesn’t.

PR #75, opened 2015-06-04
Specify the Refresh command in more detail

jgraham:


andreastt:
Did you see #74?  As far as I can tell **Refresh** doesn’t do anything differently than **Get**, if this is true I think we should consider removing it altogether.

jgraham:
I did; I meant to have PRd this already but forgot, so I thought I should throw it out there for comparison.

I agree that Refresh is rather similar to Get with the current url as an argument.

andreastt:
@jgraham, yours fixes up a lot of other small things that are useful so I’ll go ahead and close #74 in favour of this one.

I talked to @AutomatedTester that I think we should have a serious discussion about the usefulness of **Refresh** with the rest of the WG.

Unless we can come up with compelling arguments of what makes **Refresh** novel, it’s my opinion that we should remove it.  One possible action we could take to make it novel is to add a flag for force-reloading document resources (Shift-Cmd-R) but I’m wary of adding new things to the specification at this point.  Then, if anything, we should slim it down and take things away.

AutomatedTester:
landed in e98c78e. I have landed as is since it moves us forward and we can sort if we really want refresh at another date

PR #76, opened 2015-06-04
Let elements have an associated web element reference

andreastt:
Rather than defining an abstraction over DOM elements called web
element, letting each element have an ssociated web element reference
makes for a cleaner overall approach.  The implementation specifics
can then be left as an exercise for the reader.


AutomatedTester:
landed in aff1631

PR #83, opened 2015-06-05
Define Get Active Element command

andreastt:
Note that this needs further work once #76 lands as the approach used here will be incompatible.


AutomatedTester:
#76 has landed, let me know when you had done the further work and I will review and land accordingly

andreastt:
@AutomatedTester: Rebased and fixed up!

AutomatedTester:
merged in a26b9bf

PR #86, opened 2015-06-05
Remove string quote marks in validated file name

andreastt:
The string message from validator.nu that follows uses “ and ” so this should be fine.  Unfortunately BSD sed does not support the 2g flag that GNU sed does to specify the number of instances to replace.


AutomatedTester:
landed in 0c188f4

PR #87, opened 2015-06-08
Correct uses of error fields

andreastt:
An error is a JSON Object with the "error", "message", and "stacktrace"
fields.  Reflect this in the redefined commands and the document’s
data model.


andreastt:
Addresses bug 28780.

AutomatedTester:
landed in 1f55f3ec285daf2665f28305c1e24595a4eb040e

PR #91, opened 2015-06-09
Be more specific about what it means to end a session

andreastt:
* Defines what it means for a session to be **no longer active**.

* Introduces the concept of **remote end post steps** which is necessary because to **close the current session** _may_ entail that the associated browser process terminates.  Hence we need to first allow the response to be sent back, and then for the processing loop to (potentially) destroy the process.

* Appends a `closed: true` header so that a local end can tell if an unsuspecting command, i.e. **Close Window**, inadvertently ended the session.

* Changes **active sessions** from being a list to a being a set.  (Not sure if this matters so much, but semantically its easier to think about I think.)

* Removes the **current session** from the set of **active sessions** when it is closed.

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -211,11 +211,26 @@ code a:visited, code a:link {
     name <var>name</var>.
   </section>
 
-  <section>
-    <h3>Commands</h3>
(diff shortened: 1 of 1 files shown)


andreastt:
@jgraham: I’ve addressed comments in this PR in PRs #105 and #106 which replaces this.  Moreover, I’ve abandoned the idea of post steps by clarifying the note related to implementation specific post steps.  I’m sure further refinement could be done.

PR #93, opened 2015-06-09
Make a distinction between Remote End and Endpoint Node

andreastt:
A **remote end** may be either an **intermediary node** or an **endpoint node**.  The reason for this distinction is that we in one place need to talk about the ultimate, final remote end in a chain, and having a definition for that is useful.

It also helps to not have a double-meaning for the term **remote end**, so that a **remote end** can be either/or/both.


AutomatedTester:
landed in 13bc9b7

PR #97, opened 2015-06-09
First cut at updating the execute script algorithm.

jgraham:
This more precisely defines the behaviour for converting values
on entry and exit, and tries to more precisely define the context
in which the script is executed. However it is probably wrong and
will need more revision

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -3194,8 +3194,8 @@ The DOM Element <dfn>represented by</dfn> an <var>id</var> is the Element that h
 </section> <!-- /Element State -->
 
 <section>
-  <h2>Executing Javascript</h2>
-  <div class="note">
+  <h2>Executing Script</h2>
+  <div class="issue">
     <p>Open questions: What happens if a user's JS triggers a modal dialog?
       Blocking seems like a reasonable idea, but there is an assumption that
       WebDriver is not threadsafe.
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -3237,183 +3237,384 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
       on a page? How does a user of the API obtain the list of errors? Is that
       list cleared upon read?
-    </div>
-    <p class="note">Conformance tests for this section can be found in the <a href="https://github.com/w3c/web-platform-tests/tree/master/webdriver/">webdriver module</a> under the "ecmascript" folder.</p>
+  </div>
 
-    <section>
-      <h2>Javascript Command Parameters</h2>
+  <p>When required to <dfn>JSON deserialize</dfn> with
+  argument <var>input value</var>, and optional
```
jleyba: You could shorten `input value` to just `value` without loss of meaning.

webdriver-spec.html
```diff
@@ -3237,183 +3237,384 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+  </ol>
+
+  <p>The steps for <dfn>extracting the script arguments from a request</dfn>,
+  are:</p>
+   <ol>
+      <li><p>If the <a>current browsing context</a> is <a>no longer
+          open</a>, return an <a>error</a> with code <a>no such
+          window</a>.</p></li>
+
+      <li><p>Let <var>body</var> be the result of getting a property
```
jleyba: s/body/script/ to be consistent with the parameters property key?

webdriver-spec.html
```diff
@@ -3237,183 +3237,384 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
       list cleared upon read?
-    </div>
-    <p class="note">Conformance tests for this section can be found in the <a href="https://github.com/w3c/web-platform-tests/tree/master/webdriver/">webdriver module</a> under the "ecmascript" folder.</p>
+  </div>
 
-    <section>
-      <h2>Javascript Command Parameters</h2>
+  <p>When required to <dfn>JSON deserialize</dfn> with
+  argument <var>input value</var>, and optional
+  argument <var>seen</var>, which is always ignored, a remote end must
```
jleyba: `seen` is not always ignored - it is used as an input to `clone an object`.

webdriver-spec.html
```diff
@@ -3237,183 +3237,384 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
-        </section>
-        <section>
-          <h2>Reporting Errors</h2>
-          <p></p>
-        </section>
+            <li><p>Return the result of running the <a>clone an
+                  object</a> algorithm with arguments <var>input
+                  value</var> and <var>seen</var>, and the <a>internal
+                  JSON clone algorithm</a> as the clone algorithm.</li>
+            <!-- TODO: Map and Set objects? -->
```
jleyba: Map to a vanilla object (implies keys converted to strings)
Set to an array; element order is consistent with set iteration order

?

webdriver-spec.html
```diff
@@ -3237,183 +3237,384 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+          <a href="https://html.spec.whatwg.org/#environment-settings-object">environment
+            settings object</a> for <var>window</var>.</p></li>
+
+      <li><p>Let <a>script environment</a> be
+          the <a href="https://html.spec.whatwg.org/#script-execution-environment">script
+          execution environment</a> for JavaScript obtained from
+          <var>environment settings</var>.</p></li>
+
+      <li><p>If <var>body</var> is not parsable
+          as <a href="http://es5.github.io/#x13"><i>FunctionBody</i></a>
```
jleyba: s/as /as a /

webdriver-spec.html
```diff
@@ -3237,183 +3237,384 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+          or if parsing detects an <i>early error</i>,
+          return <a>error</a> with code <a>javascript
+          exception</a>.</p></li>
+
+      <li><p>If <var>body</var> begins with
```
jleyba: Should this be left to the script engine when it compiles the function body?  What if the 'use strict' directive comes after a comment?

webdriver-spec.html
```diff
@@ -3237,183 +3237,384 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+      </dl>
+  </ol>
+  <p>When required to <dfn>clone an object</dfn> with
+    arguments <var>input value</var> and <var>seen</var> and a value
+    clone algorithm <var>clone algorithm</var>, run the
+    following steps:</p>
+
+  <ol>
+    <li><p>Let <var>result</var> be a newly
+        constructed object with the same <code>[[\Class]]</code>
```
jleyba: So cloning a NodeList should yield another NodeList and not a Array?

webdriver-spec.html
jgraham: I copied this from the way that HTML does it for event handler attributes, which work in the same way.

webdriver-spec.html
jgraham: My point was that in this case we never add a value to it; it's just there to make reusing the `clone an object` algorithm simpler.

webdriver-spec.html
jgraham: Oh, I think NodeList probably needs special handing because it's a host object not a normal `Object` object. Are there other DOM types that we are supposed to support? HTMLCollection presumably.

webdriver-spec.html
```diff
@@ -3228,7 +3228,7 @@ <h2 id=element-displayedness>Element Displayedness</h2>
   </section>
 </section>
 <section>
-  <h2>Executing Javascript</h2>
+  <h2>Executing Script</h2>
   <div class="note">
```
andreastt: s/note/issue/

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
-    <p class="note">Conformance tests for this section can be found in the <a href="https://github.com/w3c/web-platform-tests/tree/master/webdriver/">webdriver module</a> under the "ecmascript" folder.</p>
+  </div>
 
-    <section>
-      <h2>Javascript Command Parameters</h2>
+  <p>When required to <dfn>JSON deserialize</dfn> with
+  argument <var>value</var>, and optional argument <var>seen</var>, a
+  remote end must run the following steps:</p>
+  <ol>
+    <li><p>If <var>seen<var> was not provided, let <var>seen</var> be
```
andreastt: Unclosed `<var>` tag

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+    an empty set.</p></li>
+    <li><p>Jump to the first appropriate step below:
+      <dl>
+        <dt>If <var>value</var> is <code>null</code>, or has
+        type <code>Boolean</code>, <code>Number</code>
+        or <code>String</code>:</dt>
+        <dd><p>Return <a>success</a> with data <var>value</var></p></dd>
```
andreastt: Missing full stop

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+
+        <dt>If <var>value</var> is an <code>Object</code>
+        that <a>represents a web element</a>:</dt>
+        <dd><p>Return the result of running the <a>deserialize a web
```
andreastt: s/a/the/ to match the **deserialise the web element** algorithm, or use the `title` attribute

andreastt: Return **success** with the result of running…

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+        element</a> algorithm with object <var>value</var>.</p></dd>
+
+        <dt>If <var>value</var> if <var>value</var> is
+        an <code>Array</code> object or an <code>Object</code> object:
+        </dt>
+        <dd><p>Return the result of running the <a>clone an object</a>
```
andreastt: Return **success**

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+    </li>
+  </ol>
 
-      <p>When calling <a href="#sync-js-execution">Synchronous JavaScript Execution</a> or calling <a href="async-js-execution">Asynchronous JavaScript Execution</a> we need to extract
-        the following from <var><a href="#post-payload">payload body</a></var> into <dfn>JavaScript Command Parameters</dfn>:</p>
+  <p>When required to make a <dfn>JSON clone</dfn> with
+  argument <var>value</var>, a <a>remote
+  end</a> must run the following steps:</p>
+  <ol>
+    <li>Let <var>seen</var> be an empty set</li>
```
andreastt: Missing full stop

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+    <li>Return the result of calling the <a>internal JSON clone
+  algorithm</a> with arguments <var>value</var> and <var>seen</var>
```
andreastt: Same here

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+            <!-- TODO: Map and Set objects? -->
+          </ol>
+        </dd>
+        <dt>Anything else:</dt>
+        <dd><p>Return <a>error</a> with code <a>javascript error</a>.</dd>
+      </dl>
+  </ol>
+  <p>When required to <dfn>clone an object</dfn> with
+    arguments <var>value</var> and <var>seen</var> and a value
+    clone algorithm <var>clone algorithm</var>, run the
```
andreastt: I’d cut the second mention of value for clarity or write this in a clearer way.

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
-                      <li>a <code>Sequence &lt;Node&gt; nodes</code>, then return the result of recursively applying this algorithm to <code>result</code>.
-                      <!-- Should there be a limit? -->
-                      <li>an object, then return the <code>object</code> created by recursively applying this algorithm to each property in <code>result</code>.
-                    </ol>
-                  </ol>
-                  <li>Set the response <a href="#response-value">value</a> to <code>value</code>.</li>
-                </ol>
-                <li>Return the response.</li>
+              <ol>
+                <li><p>Let <var>result</var> be a new <code>Array</code> object</p></li>
```
andreastt: Missing full stop

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+
+                <li><p>Let <var>length</var> be the result of <a>getting a
+                      property</a> named <code>length</code>
+                    from <var>value</var>.</p></li>
+
+                <li><p><a>Set a property</a> on <var>result</var> named <code>length</code>
```
andreastt: Missing full stop

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+            <dd>
+              <ol>
+                <li><p>Let <var>result</var> be a
+                    new <code>Object</code> object.</p></li>
+                </ol>
+            </dd>
+          </dl>
+        </li>
+
+        <li><p>For each enumerable own property in <var>value</var>,
```
andreastt: Link to [own property][http://www.ecma-international.org/ecma-262/5.1/#sec-4.3.30)

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+        run the following substeps:</p>
+          <ol>
+            <li><p>Let <var>name</var> be the name of the property.</li>
+
+            <li><p>Let <var>source property value</var> be the result
+                of <a>getting a property</a> named <var>name</var>
+                from <var>value</var>. If doing so causes script to be
+                run, and that script throws an exception,
+                return <a>error</a> with code <a>javascript
+                error</a>.</li>
```
andreastt: This is the correct behaviour, but I’m wondering if we can mandate that the implementation defined `message` string is the `message` property of the error and same with the `stacktrace` property?

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+
+        <li><p>For each enumerable own property in <var>value</var>,
+        run the following substeps:</p>
+          <ol>
+            <li><p>Let <var>name</var> be the name of the property.</li>
+
+            <li><p>Let <var>source property value</var> be the result
+                of <a>getting a property</a> named <var>name</var>
+                from <var>value</var>. If doing so causes script to be
+                run, and that script throws an exception,
```
andreastt: s/exception/error/ and also link to the concept in the ECMAScript standard?

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+          </ol>
+        </li>
+  </ol>
+
+  <p>The steps for <dfn>extracting the script arguments from a request</dfn>,
+  are:</p>
+   <ol>
+      <li><p>If the <a>current browsing context</a> is <a>no longer
+          open</a>, return an <a>error</a> with code <a>no such
+          window</a>.</p></li>
```
andreastt: I don’t think this should be part of this definition, let’s put it on each command instead.

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+
+      <li><p>Let <var>script</var> be the result of getting a property
+          named <code>script</code> from the parameters argument.</p></li>
```
andreastt: I’d just say `from <var>parameters</var>` here

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+
+      <li><p>If <var>script</var> is not a <code>String</code> return
+          an <a>error</a> with code <a>invalid argument</a>.</p></li>
+
+      <li><p>Let <var>args</var> be the result of getting a property
+          named <code>args</code> from the parameters argument.</p></li>
+
+      <li><p>If <var>args</var> is not an <code>Object</code> or
+          its <code>[[\Class]]</code> internal property is
```
andreastt: Link?

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+          not <code>Array</code> or <code>Object</code>,
+          return <a>error</a> with code <a>invalid
+          argument</a>.</p></li>
+
+       <li><p>Let <var>arguments</var> be a <a>List</a> consisting of
+          a <a>json deserialization</a> of each item
```
andreastt: We can leave the definition of a list as an exercise for the reader.

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+          in <var>args</var> with the order preserved.</p></li>
+
+       <li><p>Return <a>success</a> with data <var>script</var>
+       and <var>arguments</var></p></li>
```
andreastt: Missing full stop

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+
+   <p>When required to <dfn>execute a function body</dfn> with
+   arguments <var>body</var> and <var>arguments</var>, a <a>remote
+   end</a> must run the following steps:
+   <ol>
+      <li><p>Let <var>window</var> be the
+          <a href="https://html.spec.whatwg.org/#concept-document-window">associated
+            window</a> of the <a>current browsing
+            context</a>'s <a href="https://html.spec.whatwg.org/#active-document">active
+            document.</a></li>
```
andreastt: Put full stop outside link

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+
+      <li><p>Let <var>environment settings</var> be the
+          <a href="https://html.spec.whatwg.org/#environment-settings-object">environment
+            settings object</a> for <var>window</var>.</p></li>
+
+      <li><p>Let <a>script environment</a> be
```
andreastt: Should be `<var>` instead of `<a>`

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+          the <a href="https://html.spec.whatwg.org/#script-execution-environment">script
+          execution environment</a> for JavaScript obtained from
+          <var>environment settings</var>.</p></li>
+
+      <li><p>If <var>body</var> is not parsable as
+          a <a href="http://es5.github.io/#x13"><i>FunctionBody</i></a>
+          or if parsing detects an <i>early error</i>,
```
andreastt: Reference please

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+          return <a>error</a> with code <a>javascript
+          exception</a>.</p></li>
```
andreastt: s/exception/error/

webdriver-spec.html
andreastt: I’m assuming this is the default that applies to the outer scope.

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+
+      <li><p>If <var>body</var> begins with
+          a <a href="http://es5.github.io/#x14.1">Directive
+          Prologue</a> that contains
+          a <a href="http://es5.github.io/#x14.1">Use Strict
+          Directive</a> then let <var>strict</var> be true, otherwise
+          let <var>strict</var> be false.</p></li>
+
```
andreastt: When evaluating a script in Marionette we need to say what the default JavaScript version is (see for example [Components.utils.evalInSandbox](https://developer.mozilla.org/en/docs/Components.utils.evalInSandbox)).  It feels we should say something about that as well.

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+      <li><p>Using
+          the <a href="https://html.spec.whatwg.org/#script-execution-environment">script
+          execution environment</a> <var>script environment</var>,
+          let <var>function</var> be the result of
+          calling <a href="http://es5.github.io/#x13.2">create a
+          function object</a> with parameter list of an
+          empty <a href="http://es5.github.io/#x8.8">List</a>,
+          body <var>body</var>, scope of
+          the <a href="http://es5.github.io/#x10.2.3">global
+          environment</a> and strict flag <var>strict</var></p></li>
```
andreastt: Missing full stop

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+
+      <li><p>Let <var>script</var> be a
+          new <a href="https://html.spec.whatwg.org/#concept-script">script</a>.</p></li>
+
+      <li><p>Let <var>script</var>'s <a href="">code entry-point</a>
+            be <var>function</var></p></li>
```
andreastt: Missing reference and full stop

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+
+      <li><p>Let script's <a href="">settings object</a> object
+          be <var>script settings</var>.</p></li>
```
andreastt: Reference again

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+      <!-- the stuff from here on down is quite dubious -->
+      <!-- this incumbent scripts stuff is in webidl but the references don't
+           lead anywhere -->
+      <!-- li><p><a href="">Push <var>script</var> onto <a>the stack of
+              incumbent scripts</a>.</li-->
+
+      <li><p>Invoke the <code>[[\Call]]</code> internal method
+          of <var>function</var>, providing <var>window</var> as the
+          this value and <var>parameters</var> as the argument
+          values. If doing so does not produce an exception,
```
andreastt: s/exception/error/

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+          let <var>result</var> result be <a>success</a> with data set
```
andreastt: Remove one case of “result”

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+          <th>Path Template</th>
+          <th>Notes</th>
+        </tr>
+        <tr>
+          <td>POST</td>
+          <td id='post-execute_script'>/session/{sessionId}/execute</td>
+          <td></td>
+        </tr>
+      </table>
+    <p>The <dfn>Execute Script</dfn> command executes a javascript
```
andreastt: s/javascript/JavaScript/

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+      function in the context of the <a>current browsing context</a>
+      and return the return value of the function.
```
andreastt: s/return/returns/

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+
+      <li><p>Let <var>body</var> and <var>arguments</var>
+      be <var>script arguments</var>' data.</p></li>
+
+      <li><p>Let <var>result</var> be the result of calling <a>execute
+      a function body</a>, with arguments <var>body</var>
+      and <var>arguments</var>.</p></li>
+
+      <li><p>If <var>result</var> is an <a>error</a>, return result,
+      otherwise let <var>value</var> be <var>result</var>'s data</p></li>
```
andreastt: Missing full stop

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+
+      <li><p>Let <var>data</var> be a new <code>Object</code></p></li>
```
andreastt: Missing full stop

andreastt: I notice you’ve started using `<code>Object</code>` instead of JSON Object.  I think this makes sense, but just to clarify, are we then using internal JS data structures which our serialisation definitions will convert into JSON Object’s on their way out?

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+
+      <li><p>Set the property <code>value</code> of <var>data</var>
```
andreastt: `<a>Set the property</a>`

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+      to <var>value</var></p></li>
```
andreastt: Missing full stop

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+    <p>
+      <table class="simple jsoncommand">
+        <tr>
+          <th>HTTP Method</th>
+          <th>Path Template</th>
+          <th>Notes</th>
+        </tr>
+        <tr>
+          <td>POST</td>
+          <td id='post-execute_async'>/session/{sessionId}/execute_async</td>
```
andreastt: The `id` attribute is not necessary.

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+  </section>
+
+  <section>
+    <h3 id="async-js-execution">Execute Async Script</h3>
+    <p>
+      <table class="simple jsoncommand">
+        <tr>
+          <th>HTTP Method</th>
+          <th>Path Template</th>
+          <th>Notes</th>
```
andreastt: Remove the _Notes_ column

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+
+      <li><p>Set the property <code>value</code> of <var>data</var>
+      to <var>value</var></p></li>
+
+      <li><p>Return <a>success</a> with data <var>data</var>.</p></li>
+    </ol>
+  </section>
+
+  <section>
+    <h3 id="async-js-execution">Execute Async Script</h3>
```
andreastt: `id` attribute can be removed

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+    <p>
+      <table class="simple jsoncommand">
+        <tr>
+          <th>HTTP Method</th>
+          <th>Path Template</th>
+          <th>Notes</th>
+        </tr>
+        <tr>
+          <td>POST</td>
+          <td id='post-execute_script'>/session/{sessionId}/execute</td>
```
andreastt: Remove the `id` attribute

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+    </ol>
+
+  <section>
+    <h3>Execute Script</h3>
+    <p>
+      <table class="simple jsoncommand">
+        <tr>
+          <th>HTTP Method</th>
+          <th>Path Template</th>
+          <th>Notes</th>
```
andreastt: Remove the _Notes_ column

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+        </tr>
+        <tr>
+          <td>POST</td>
+          <td id='post-execute_async'>/session/{sessionId}/execute_async</td>
+          <td></td>
+        </tr>
+      </table>
+      <p>The <dfn>Execute Async Script</dfn> command causes javascript
```
andreastt: s/javascript/JavaScript/

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+      to execute as an anonymous function. Unlike the <a>Execute
+      Script</a> command, the result of the function is
+      ignored. Instead an additional argument <code>callback</code> is
```
andreastt: Since this is an unnamed argument that is always the last argument to `arguments`, it seems misleading to talk about a name for it here?

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+      provided to the function. This is a function that, when called,
+      returns its first argument as the webdriver response.</p>
```
andreastt: s/webdriver// but possibly the whole paragraph could do with better phrasing

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+      <li><p>Let <var>body</var> and <var>arguments</var>
+      be <var>script arguments</var>' data.</p></li>
+      <!-- this next step might not quite set up all the right
+      machinary -->
+
+      <li><p>Let <var>webdriver callback result</var> be
+      a flag which can have three
+      values; <code>unset</code>, <code>expired</code>,
+      or <code>set</code>, with the <code>set</code> value having
+      associated data. Initially it is in the <code>unset</code> state.</p></li>
```
andreastt: Hm.  Should we consider extracting this to a command specific definition to simplify this algorithm?

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+
+      <li><p>Let <var>callback</var> be a javascript function
```
andreastt: s/javascript/JavaScript/, also can we provide a reference?

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+          result.</p></li>
+
+      <li><p>Wait for <var>webdriver callback result</var> to enter
+      the <code>set</code> state, or for <a>session script timeout</a>
+      milliseconds to expire, whichever happens sooner.</p></li>
+
+      <li><p>If the previous step completed due the timeout being
+          reached, set <var>webdriver callback result</var>
+          to <code>expired</code> and return <a>error</a> with
+          status <a>script timeout</a>.</p></li>
```
andreastt: “the timeout” → reference `session script timeout` instead to be specific

andreastt: s/status/code/

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+
+      <li><p>Otherwise, let <var>result</var> be <var>webdriver
+      callback result</var>'s data.</p></li>
+
+      <li><p>If <var>result</var> is an error,
+          return <var>result</var>.</p></li>
```
andreastt: `<a>error</a>`

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+
+      <li><p>Let <var>value</var> be <var>result</var>'s data</p></li>
```
andreastt: Missing full stop

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+
+      <li><p>Let <var>data</var> be a new <code>Object</code></p></li>
```
andreastt: Missing full stop

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+
+      <li><p>Set the property <code>value</code> of <var>data</var>
+           to <var>value</var></p></li>
```
andreastt: Missing full stop

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+
+      <li><p>Return <a>success</a> with data <var>data</var>.</p></li>
+    </ol>
+
+    <p>The <dfn>execute async script callback</dfn> algorithm is
+    initialized with a single argument <var>webdriver callback
+    state</var>. It defines a javascript function with a single
```
andreastt: s/javascript/JavaScript/ and please add a reference

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+          the <code>unset</code> state, return <code>undefined</code>.</p></li>
+
+      <li><p>Otherwise, if <var>result</var> is not present,
+          let <var>result</var> be <code>null</code>.</p></li>
+
+      <li><p>Let <var>json result</var> be a <a>JSON clone</a>
+          of <var>result</var>.</p></li>
+
+      <li><p>Set the <var>webdriver callback state</var>
+          to <code>set</code> with data <var>josn result</var>.</p></li>
```
andreastt: s/josn/json/

webdriver-spec.html
```diff
@@ -3237,183 +3237,428 @@ <h2 id=element-displayedness>Element Displayedness</h2>
...
+
+      <li><p>Return undefined</p></li>
```
andreastt: `<code>undefined</code>`, and missing full stop

andreastt:
This is a very good improvement to the existing prose.  Although I’m very unfamiliar with this part of HTML I think it’s more accurate than what we had and I’m in favour of merging it pretty much in its current state – the fixups we can do later and incrementally.

Whilst looking through this PR I haven’t compared it to the existing implementations and what they do.  I imagine many of them will not have the same level of fine grained control over JavaScript execution that this PR requests, but I don’t anticipate this to be a problem since an implementation is said _to be compliant_ if its output matches what the specification prescribes.

webdriver-spec.html
jgraham: It means a specific thing in the ES spec, which is what I need here. The link is missing though.

webdriver-spec.html
jgraham: Well it just means that if you do 

    execute_script("""
        "use strict"
        /* more stuff */
    """)

It will run in strict mode, as you might expect. I'm not sure how terribly useful it is since the arguments object won't be available; at least `execute_async_script` really requires this, but we could fix that I think.

webdriver-spec.html
jgraham: Well OK, I guess, but I feel like "Executing Script" is fine since it avoids the whole JavaScript vs Javascript vs ECMAScript thing, and matches the command names.

webdriver-spec.html
jleyba: In lieu of using `arguments`, perhaps we should mandate returning a promise-like object in `execute_async_script`?

If we add support for promises, we probably could just drop `execute_async_script`.

webdriver-spec.html
andreastt: I didn’t complain about the naming of the command.  There are some cases of lowercased "javascript" elsewhere.  I think "script" is better, for the reasons you mention.

jgraham: Oh, sorry, I misunderstood GH because the highlighted lines and the lines you commented on were not the same.

webdriver-spec.html
jgraham: The idea of multiple incompatible javascript versions is Mozilla-proprietary so we don't need to cover it.

webdriver-spec.html
jgraham: To where?

webdriver-spec.html
jgraham: ES talks about scripts throwing (error objects as) exceptions. I don't know where I would link to.

webdriver-spec.html
jgraham: "Implementation defined" seems transitive; I don't think that we can call it implementation defined and mandate a specific value. I think we could add a note suggesting that for javascript errors the message and stacktrace properties should match those on the corresponding js object, where available.

webdriver-spec.html
jgraham: I don't know what's best really. We will need to clean all of this up at some point, which will be tedious :(

AutomatedTester:
landed in 8a6fbe2

PR #98, opened 2015-06-10
Make some preliminary changes to define the screenshot section better.

jgraham:


andreastt:
Ship it 👍🏻

thejohnjansen:
fix the misspelling of "and" on line 4528 :-)

thejohnjansen:
Based on our conversation at the face-to-face, I thought we agreed the default would be to capture the viewport, but that browsers could also get the entire document by using getSize() of the body element.

http://www.w3.org/2014/10/30-testing-minutes.html#item10

andreastt:
Yes, we still need to redefine the screenshot commands.

andreastt:
@jgraham Will you rebase this?

AutomatedTester:
landed in 5c3376c

PR #100, opened 2015-06-12
Explicitly require navigation actions to end with top-level context.

sevaseva:
<p class=note>Navigation actions - Get, Back, Forward, Refresh - always,
  unless they abort the navigation, complete with the <a>current top-level browsing context</a>
  being the <a>current browsing context</a>. Even if we navigate Back or Forward to a state
  where a different <a href=https://html.spec.whatwg.org/#browsing-context>browsing context</a>
  was the <a>current browsing context</a>.

Alternative behavior *could* be for the remote end to "remember"
what was the <a>current browsing context</a> at each of the state
of the "history". I don't support that behavior because I believe
it is more complex in implementation
([speculating] other than WebDriver, browser components don't seem
to already define the notion of the <a>current browsing context</a>
so WebDriver would have to "remember" those for itself [/speculating]).

This suggestion was prompted by the regression (in this yet unspecified behavior)
in chrome/chromedriver:
https://code.google.com/p/chromedriver/issues/detail?id=1106

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -1244,6 +1244,12 @@ associated browser process to be killed</p>.
   time that commands will block before returning with a <a>timeout</a>
   error.
 
+  <p class=note>Navigation actions - Get, Back, Forward, Refresh - always,
+  unless they abort the navigation, complete with the <a>current top-level browsing context</a>
+  being the <a>current browsing context</a>. Even if we navigate Back or Forward to a state
+  where a different <a href=https://html.spec.whatwg.org/#browsing-context>browsing context</a>
+  was the <a>current browsing context</a>.
+
 <section>
 <h3>Get</h3>
 
@@ -1313,6 +1319,12 @@ associated browser process to be killed</p>.
     <!-- perhaps? Not sure if this is what the spec is trying to say -->
     return an <a>error</a> with code <a>timeout</a>.
   </ol>
+
+ <li><p>Set the <a>current browsing context</a> to the
+  <a>current top-level browsing context</a> as if <a>Switch to Frame</a>
+  command was executed with no specified
+  <a href="https://html.spec.whatwg.org/#child-browsing-context">
+  child browsing context</a>.
 
  <li><p>Return <a>success</a> with data null.
 </ol>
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -1314,6 +1320,12 @@
     return an <a>error</a> with code <a>timeout</a>.
   </ol>
 
+ <li><p>Set the <a>current browsing context</a> to the
+  <a>current top-level browsing context</a> as if <a>Switch to Frame</a>
+  command was executed with no specified
+  <a href="https://html.spec.whatwg.org/#child-browsing-context">
+  child browsing context</a>.
```
andreastt: I don’t think we should make a reference to **Switch To Frame** at all here.  If we need some generalised prose related to resetting the **current browsing context** we should add that as a definition, but I don’t think we do.

webdriver-spec.html
```diff
@@ -1244,6 +1244,12 @@
   time that commands will block before returning with a <a>timeout</a>
   error.
 
+  <p class=note>Navigation actions - Get, Back, Forward, Refresh - always,
+  unless they abort the navigation, complete with the <a>current top-level browsing context</a>
+  being the <a>current browsing context</a>. Even if we navigate Back or Forward to a state
+  where a different <a href=https://html.spec.whatwg.org/#browsing-context>browsing context</a>
+  was the <a>current browsing context</a>.
```
andreastt: What do you mean by “abort the navigation”?  I think this note is unnecessarily complicated if this is explained by the algorithms.

sevaseva: "unless they abort the navigation" is here for the following reason:

Back and Forward commands can complete returning success in certain cases, specifically: "If the task queued by that algorithm aborted before traversing the history, return success with data null."

That step in Back and Forward algorithms, I assume, exists for a reason. But it doesn't allow to assume that a navigation command that returned success performed navigation.

There are certainly ways to make this note more brief. I am happy to amend stuff if there are concrete proposals; I also trust editors just submitting something along the lines with this, without using this PR directly.

sevaseva: Correction to my last comment:

s/Back and Forward commands can complete returning success in certain cases, specifically/Back and Forward commands can complete returning success in certain cases **where the navigation wasn't actually performed**, specifically/

Putting what I wanted to say in the last comment in other words, I would have used a simple language like [1] if a successful response from those actions meant that the navigation has been successfully performed, but that is not guaranteed to always be the case. For that reason I use more verbose and complex language to stay logically correct.

Hope this makes some sense.

[1] <a>current browsing context</a> is guaranteed to be the <a>current top-level browsing context</a> after a successful response sent by a navigation command (Get, Back, Forward or Refresh) **even** if it wasn't before.

webdriver-spec.html
sevaseva: Agreed, 

" as if <a>Switch to Frame</a>
 command was executed with no specified
 <a href="https://html.spec.whatwg.org/#child-browsing-context">
 child browsing context</a>"

can be removed completely. I can update this PR when the other comment in this PR is resolved.

andreastt:
I believe this is mostly replaced by #275.

andreastt:
@AutomatedTester I believe this can now be closed.

PR #105, opened 2015-06-15
Indicate in header list that session was closed

andreastt:


andreastt:
@AutomatedTester r?

AutomatedTester:
I have been meaning to ask @jgraham to have a look at this.

andreastt:
Closing this as I reminded myself that we changed **Close Window** to return the set of window handles, which the local end can use for determining whether the session was closed or not.

This doesn’t cater for unintended/unexpected session closures as a result of fatal errors in the browser, but then this might not be within WebDriver’s scope.

PR #107, opened 2015-06-29
/session/{sessionId}/alert_text: fix repeated and misplaced sentences

sevaseva:


AutomatedTester:
This whole section of the spec needs rewriting in a more prescriptive manner  so am going to close this for now

PR #108, opened 2015-07-16
add editable command

ddavison:
From [SeleniumHQ/selenium#798](https://github.com/seleniumhq/selenium/issues/798)...

Editable would be a useful action to perform to see if the element in question, is editable.

I'll later make a push to selenium master if this gets merged

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -525,6 +525,10 @@ code a:visited, code a:link {
         <td><a>Is Element Enabled</a></td>
       </tr>
       <tr>
+        <td>GET</td>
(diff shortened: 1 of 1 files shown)


AutomatedTester:
At the moment this does not feel like it should be an end point to WebDriver. This can be achieved purely on the client side by doing

``` python
def is_editable(self):
      enabled = self.is_enabled()
      editabled = self.get_attribute('contenteditable')
      return enabled and (editable or False)
```

If you really feel strongly that this should be added to the specification can you email the [W3C Working group](http://www.w3.org/testing/browser/) with why it should be added as well as relevant use cases. Make sure to like back to this PR.

PR #110, opened 2015-07-20
import CSS WG styling

andreastt:
Imports the CSS WG’s specification styling as-is.

CSS style applied to WebDriver specification.


AutomatedTester:
I actually don't like the CSS styling. The main issue is links are not obvious and then there are a number of things that just bug me.

What is it from this that you want?

andreastt:
There are a few things that annoy me with the default W3C styling:

* Free flowing text with no maximum width makes the document hard to read in wide windows on big monitors
* Ability to click headers and be given the anchor
* Styling of the menu is closer to good typographic standards
* The link styling is more consistent (ours masks visited references as regular text)
* Excess in various forms, i.e. overflows and elements expanding beyond the boundaries of their parent elements, is better handled

All these issues can be addressed in style directive directly in our specification.  Would you rather I did that?

andreastt:
It was also my impression that W3C would adopt portions of the CSS WG’s styling into theirs, at least if I’ve read the feedback thread correctly.  But I agree most (probably ~80%) of the rules they have make no difference to us.

andreastt:
Talked to @AutomatedTester and decided we should apply inline style fixes to the concrete problems we do have, instead of importing several hundred lines of unfamiliar stylesheet directives we don't know.

PR #111, opened 2015-07-21
Add a switch case statement

andreastt:


AutomatedTester:
There are a number of other of other switch statements in the specification that not have been updated.

I quite like the original had the type in there, while I don't think it would look nice, I would like to see if there is something that we can do to show the type there in case we do switch statements that have something other than strings

andreastt:
> There are a number of other of other switch statements in the specification that not have been updated.

Yes, I can update them either as part of this PR or later.

> I quite like the original had the type in there, while I don't think it would look nice, I would like to see if there is something that we can do to show the type there in case we do switch statements that have something other than strings

It's implicit that `"<code>foo</code>"` is a string.  I stole the idea from the [Fetch spec](https://fetch.spec.whatwg.org/) which uses this construct also as a match group, in the “run the steps of the first matching statement” style.

AutomatedTester:
Looking at [Main Fetch](https://fetch.spec.whatwg.org/#main-fetch) from the [Fetch](https://fetch.spec.whatwg.org/) specification I think if we can use similar word (use `match` over `switch` then let's do this.)

andreastt:
Yes, that certainly makes sense when you're matching on a combination of multiple different properties and types.

I'll rebase this and update the remainder of the missing possible switch statements.

AutomatedTester:
<3



—

AutomatedTester:
Reviewed 1 of 1 files at r2.
Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/w3c/webdriver/111)

AutomatedTester:
landed in 3f9e590

PR #114, opened 2015-07-21
Restyle element displayedness

andreastt:
There are **no functional changes** in this PR except restyling to match the rest of the specification.


AutomatedTester:
landed as d9f1ae4 and 7c69e90

PR #120, opened 2015-07-21
Change Set Timeout to accept dictionary of timeouts

andreastt:
This changes some characteristics about the Set Timeout command.
It will not complain about unknown timeout members on the dictionary
it receives, but will complain about the data sanity of the value
attached to known timeouts, listed in the new table of session
timeouts.

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -1273,7 +1273,7 @@ code a:visited, code a:link {
 
    <li><p>If <var>timeout</var> is of the
    <code><a href=http://www.ecma-international.org/ecma-262/5.1/#sec-4.3.19>Number</a></code> type
-   and a positive integer,
+   and a non-negative integer,
    set the <var>type</var> timeout to
    <var>timeout</var>’s value in milliseconds.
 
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -1208,57 +1208,83 @@
...
+    an <a href=http://www.ecma-international.org/ecma-262/5.1/#sec-4.3.30>own property</a>
+    <var>identifier</var>, continue to the next entry.
+
+   <li><p>Let <var>timeout</var> be the result of
+    <a>getting a property</a> using <var>identifier</var>
+    from <var>parameters</var>.
+
+   <li><p>If <var>timeout</var> is of the
+   <code><a href=http://www.ecma-international.org/ecma-262/5.1/#sec-4.3.19>Number</a></code> type
+   and a positive integer,
```
AutomatedTester: Resetting a timeout should happen if we are are given 0 so limiting it to positive integers means we can't reset it.

andreastt: Good point and well spotted. I'll address that in a follow up.

andreastt: Changed it to "non-negative integer", which was suggested by Wikipedia.

AutomatedTester:
landed in 09bd2c5

PR #123, opened 2015-07-23
Makes Modals and following chapters top-level chapters

andreastt:
Otherwise no function changes.


AutomatedTester:
landed in 9f30f44

PR #125, opened 2015-07-24
Fix some errors in the algorithm of parsing capabilities

sevaseva:


sevaseva:
This fixes several tiny issues making this algorithm ... not achieving what it's meant to.

E.g. if "<var>unprocessed capabilities</var> be a JSON List that contains <var>required capabilities</var> and <var>desired capabilities</var>" then <var>unprocessed capabilities</var> will always have exactly two elements in the list. et cetera.

sevaseva:
So I was reviewing f05cbd4f94c05d17d1adef698dff319972bcbcee and decided to sent a couple of PRs including this.

I do think this part of the spec is going too low level in spelling out algorithms.

Two more issues this PR does NOT fix (and I don't plan to send PRs to fix them):

1. <var>server capabilities</var> is only ever assigned to once, and the assigned value is not a specific value. (Let <var>server capabilities</var> be a JSON Object with the following entries:..). The algorithm then never modify that assigned value. Then the algorithm goes ahead and returns <var>server capabilities</var>. What does it return?

2. "append a string" substring (case insensitive) appears in this document (both times in this capabilities algorithm) twice. In both cases it is unclear, at least to me, where the string is being appended *to*.

AutomatedTester:
Seva,

Can you raise bugs for those and I will look into them. Thanks for the PR

AutomatedTester:
landed in 3249a56

PR #126, opened 2015-07-24
More validation fixes

andreastt:

PR #127, opened 2015-07-24
Issue exit code reflecting output from validator

andreastt:
Replaces _Makefile_ with a validation script that uses the correct exit code.


andreastt:
We are expecting this PR to fail the checks until #126 is merged.

PR #137, opened 2015-07-26
Define Get Element CSS Value

andreastt:
It is safe to remove the provision about CSS colour as the computed
value will be calcaulated to an absolute value.

See the CSS Cascade specification on computed values for more information:

	https://drafts.csswg.org/css-cascade-4/#computed-value

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -3014,7 +3014,7 @@ code a:visited, code a:link {
  </tr>
  <tr>
   <td>GET</td>
-  <td>/session/{session id}/element/{element id}/css/{property name}</td>
+  <td>/session/{sessionId}/element/{elementId}/css/{propertyName}</td>
  </tr>
 </table>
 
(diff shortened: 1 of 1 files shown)


AutomatedTester:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[webdriver-spec.html, line 3017 \[r1\]](https://reviewable.io:443/reviews/w3c/webdriver/137#-JvQbqSWlCnw3ofuK-je)** ([raw file](https://github.com/w3c/webdriver/blob/a33b88c39bd301307d3fba5029c059e371e9116e/webdriver-spec.html#L3017)):</sup>
The spec currently has no space here. Please change back and we can discuss for a future PR

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/w3c/webdriver/137)

andreastt:
@AutomatedTester Reverted the changes to the URL template as requested.

The problem we're facing is that the template for the URL does not match how we treat them as variables.  We should revisit this.

AutomatedTester:
I am not disagreeing with that, I just don't think it is appropriate for this PR.



—

andreastt:
@AutomatedTester Absolutely.  We will address that later.

AutomatedTester:
Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/w3c/webdriver/137)

AutomatedTester:
landed in f70e543

PR #138, opened 2015-07-26
Is Element Displayed: correct retrieval of element

andreastt:
The command is being passed an element reference, not a web element
object.  Thus it needs to use the `getting a known element` algorithm
rather than the deserialisation algorithm it was using prior to this
patch.

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -2967,7 +2967,7 @@ URL, and command for each WebDriver <a>command</a>.
  </tr>
 </table>
 
-<p>The <dfn>Is Element Displayed</dfn> command
+<p>The <dfn>Is Element Displayed</dfn> <a>command</a>
  is used to determine the <a title="element displayed">element displayedness</a>
  of a <a>web element</a>.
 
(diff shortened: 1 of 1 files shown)


AutomatedTester:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

<sup>**[webdriver-spec.html, line 2838 \[r1\]](https://reviewable.io:443/reviews/w3c/webdriver/138#-JvToNwUrcV0b7zfIaly)** ([raw file](https://github.com/w3c/webdriver/blob/8959f7fc0aecba3eac5df9439e2a724785dc0a50/webdriver-spec.html#L2838)):</sup>
revert the URL please as previously discussed :)

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/w3c/webdriver/138)

AutomatedTester:
also could you rebase? :heart:

AutomatedTester:
landed in cc1f952

AutomatedTester:
Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/w3c/webdriver/138)

PR #154, opened 2015-07-30
Protocol Extensions: first stab at defining extensions to the protocol

andreastt:
I might be doing something very wrong here.


AutomatedTester:
landed in 84a60f5

PR #167, opened 2015-08-03
Hide <hr>

andreastt:


AutomatedTester:
where do we have a `<hr>` ?

andreastt:
@AutomatedTester ReSpec adds an `<hr>` when it generates the header information from the inline JS config for some bizarre reason.

Anyways, it makes the spec look like it’s from 2004.

AutomatedTester:
landed in 03018a4

PR #169, opened 2015-08-03
Security and Privacy: individual sections and bring them to the left

andreastt:
Otherwise no functional changes, except some new references.


AutomatedTester:
landed in d584b6f

PR #185, opened 2015-08-04
run validation script on all PR branches

andreastt:
after_deploy does not trigger on PR branches

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,10 @@
 language: c
 script: ./validate
 sudo: false
-branches:
-  only:
-    - gh-pages
 env:
   global:
     - URL="https://w3c.github.io/webdriver/echidna-manifest.txt"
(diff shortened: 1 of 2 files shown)


andreastt:
Should be a bit cautious with landing this as it can break autopublishing with Echidna.

sideshowbarker:
> Should be a bit cautious with landing this as it can break autopublishing with Echidna.

Don’t be too cautious on that. If the Echidna autopublishing setup breaks, we just fix it. And now is the best time to be breaking it (rather than making changes later and having it break then).

We’re still one of the earliest users of auto-publishing tools, and so far in getting it set up for WebDriver, I’ve exposed at least 2 or 3 bugs that need to be fixed in those tools.

AutomatedTester:
landed in 0182943. Only took the .travis.yml stuff as the other was handled in #183

PR #187, opened 2015-08-05
Get Element Attribute: make it actually return the attribute value

andreastt:
We should not coerce values of boolean attributes as this is the job for
getting an element’s property.  The use case for this is wanting to get
"foo" from `<input disabled="foo">`.

The patch also addresses getting the attribute of web element's element.

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -3553,20 +3553,21 @@ following must be <var>parameters</var> after the local end has made a request t
 <h3>Get Element Attribute</h3>
 
 <table class="simple jsoncommand">
-  <tr>
-    <th>HTTP Method</th>
-    <th>Path Template</th>
-  </tr>
-  <tr>
-    <td>GET</td>
-    <td>/session/{<var>session id</var>}/element/{<var>element id</var>}/attribute/{<var>name</var>}</td>
-  </tr>
+ <tr>
+  <th>HTTP Method</th>
+  <th>Path Template</th>
+ </tr>
+ <tr>
+  <td>GET</td>
+  <td>/session/{<var>session id</var>}/element/{<var>element id</var>}/attribute/{<var>name</var>}</td>
+ </tr>
 </table>
 
 <p>The <dfn>Get Element Attribute</dfn> <a>command</a>
- will return the <a>attribute</a> of a <a>web element</a>.
+ will return the <a>attribute</a> of the given <a>element</a>.
 
 <p>The <a>remote end steps</a> are:
+
 <ol>
  <li><p>If the <a>current top-level browsing context</a> is <a>no longer open</a>,
   return <a>error</a> with <a>error code</a> <a>no such window</a>.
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -3220,35 +3222,26 @@ <h2 id="element-location-strategies">Element Location Strategies</h2>
...
-
-   <li><p>Return <a>success</a> with data <var>body</var>.
+ <li><p>If <var>name</var> is undefined,
+  return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
+
+ <li><p>Let <var>attribute</var> be the result of
+  <a href=https://dom.spec.whatwg.org/#concept-element-attributes-get-by-name>getting an attribute</a>
+  given <var>name</var> and <var>element</var>.
+
+ <li><p>Let <var>result</var> be the value of <var>attribute</var>
```
AutomatedTester: I don't think just returning the value here is good because `<input disabled>` is just going to return `""` where I really think that they want the result of `hasAttribute`. Since [boolean attributes](https://html.spec.whatwg.org/#boolean-attribute) don't actually need values. Most best practises recommend that people do `hasAttribute` for booleans I would rather have that as the result.

andreastt: That’s what calling `getAttribute` on a DOM element is supposed to return.  Having special WebDriver behaviour for this means you wouldn’t be able to return "incorrect" from `<input disabled=incorrect>`, and would deviate from DOM.

If you want to check if an element is disabled we should be encouraging the use of `Get Element Property`.  The concept of coerced boolean attributes is what the browser needs to determine if the input attribute is valid or not, in order to set the element’s `disabled` _property_.

andreastt:
@AutomatedTester r?

webdriver-spec.html
AutomatedTester: With Web Components there is no reflection between attributes and properties here so people will want to known if the attribute is disabled or not. 

I am ok with conflating `getAttribute` and `hasAttribute` in this call.

andreastt: I’m not really sure what you’re proposing here.  That **Get Element Attribute** duplicate the behaviour of **Get Element Property** _specifically_ when interacting with form control elements?  All elements?

I think my primary counterargument is that attributes is a DOM concept, and the DOM doesn’t say anything about disabledness/checkedness of HTML elements.  I don’t think it’s a particularly good idea if we rocked this idea, despite that HTML reflects changes to the property to the DOM.

Since it’s HTML that [defines the **disabled** concept](https://html.spec.whatwg.org/#concept-fe-disabled), what users want also isn’t `hasAttribute("disabeld")`, although in the particular case of calling that on form control elements it would _effectively_ give you the same result.

I think making the change I think you’re suggesting would lead to a situation where WebDriver has a special definition of what it means to be a content attribute, different to that of DOM and HTML.  It would effectively mean that if users actually wanted the actual, _correct_ value of the content attribute, they’d have to call **Execute Script**, asking for `arguments[0].getAttribute("disabled")`, which yet again proves why we shouldn’t overload the “attribute” concept.  It’s hard enough to explain what **Get Element Attribute** does in Selenium WebDriver as it is.

I’d rather we keep the primitives close to existing web platform technologies instead of inventing our own behaviour and _especially_ overload existing terminology.  If local ends wish to implement a behaviour that diverges from the web platform, they are able to do so with the primitives we provide them with.

jgraham: Per my comment on the Wires bug, I don't actually see how to implement the currently specified behaviour without a list of boolean attributes in the implementation. That's because it isn't possible to tell whether a specific content attribute is actually a boolean attribute through DOM APIs.

andreastt:
Rebased again.

AutomatedTester:
I have spoke to @andreastt about this and think that this will be a good point to discuss at TPAC. Leaving open for now.

AutomatedTester:
this was decided to remain as is

PR #188, opened 2015-08-05
Get Element Property: get the property, not the own property

andreastt:


AutomatedTester:
Landed 911ce45 with tweaks

PR #190, opened 2015-08-05
validate: use https://validator.w3.org/nu/ as our validator

andreastt:
According to Michael[tm] Smith this validator is more frequently updated
and also has better stability.


AutomatedTester:
landed in dfcab32

PR #191, opened 2015-08-05
Delete old working draft

andreastt:
This is archived at http://www.w3.org/TR/2013/WD-webdriver-20130312/
and following our change to become a living document, it is rendered
unnecessary to keep an extra copy in our source repository.


AutomatedTester:
landed in 81bdb39

PR #203, opened 2015-08-06
Invalid SSL Certificates: bring to the left

andreastt:
Although this chapter should probably be deleted.


AutomatedTester:
landed in 902889b

PR #204, opened 2015-08-06
Handling non-HTML Content: bring to the left

andreastt:


AutomatedTester:
To be honest, this stuff needs to removed and placed in the relevant parts of the spec. I am closing this one for now and we can revisit it at a later date

PR #214, opened 2015-08-24
Cookies: define a specific cookie store

andreastt:
Defines precisely which cookie store cookies are fetched from, and also
addresses the issue that we did not mandate to include HTTP-only cookies.

Thanks to Jim Evans for input.

[Review on Reviewable](https://reviewable.io/reviews/w3c/webdriver/214)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -4064,6 +4064,15 @@ URL, and command for each WebDriver <a>command</a>.
  <a href=http://www.w3.org/TR/html5/dom.html#dom-document-cookie>cookies</a>
  as described in the [[!html51]].
 
+<p>To get <dfn>all associated cookies</dfn> to
+ a <a href=https://dom.spec.whatwg.org/#concept-document>document</a>,
+ the user agent must return the enumerated set of cookies
+ that meet the requirements set out in the first step of the algorithm in [[RFC6265]] to
+ <a href=https://tools.ietf.org/html/rfc6265#section-5.4>compute <code>cookie-string</code></a>
+ for the <a href=https://dom.spec.whatwg.org/#concept-document>document</a>’s
+ <a href=https://dom.spec.whatwg.org/#concept-document-url>address</a>
+ for an ‘HTTP API’.
+
 <p>A <dfn>serialized cookie</dfn> is created with the following algorithm:
 
 <ol>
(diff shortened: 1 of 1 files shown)


andreastt:
@AutomatedTester Removed the storage mutex provision.

AutomatedTester:
Review status: 0 of 1 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

<sup>**[webdriver-spec.html, line 4068 \[r2\]](https://reviewable.io:443/reviews/w3c/webdriver/214#-Jxj9LeLWek7vxlWSmZs)** ([raw file](https://github.com/w3c/webdriver/blob/1b8ebfd2316e46ab1d780d0445fff858e9876872/webdriver-spec.html#L4068)):</sup>
this line is not needed as you have said that it is associated to the document address below.

---

<sup>**[webdriver-spec.html, line 4074 \[r2\]](https://reviewable.io:443/reviews/w3c/webdriver/214#-Jxj9mtaOh1EOr4xEIV7)** ([raw file](https://github.com/w3c/webdriver/blob/1b8ebfd2316e46ab1d780d0445fff858e9876872/webdriver-spec.html#L4074)):</sup>
what is an `HTTP API`

---

<sup>**[webdriver-spec.html, line 4161 \[r2\]](https://reviewable.io:443/reviews/w3c/webdriver/214#-JxjA49iH7Iv9m3Wzhgf)** ([raw file](https://github.com/w3c/webdriver/blob/1b8ebfd2316e46ab1d780d0445fff858e9876872/webdriver-spec.html#L4161)):</sup>
s/switching/matching/ so that it is more in line with the rest of the specification

---

<sup>**[webdriver-spec.html, line 4269 \[r2\]](https://reviewable.io:443/reviews/w3c/webdriver/214#-JxjAMYStr1wzztmRmcQ)** ([raw file](https://github.com/w3c/webdriver/blob/1b8ebfd2316e46ab1d780d0445fff858e9876872/webdriver-spec.html#L4269)):</sup>
I don't think this issue is valid since no UA implements `storage mutex` and the storage model in the RFC is an exercise for the reader

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/w3c/webdriver/214)

andreastt:
<sup>**[webdriver-spec.html, line 4074 \[r2\]](https://reviewable.io:443/reviews/w3c/webdriver/214#-Jxj9mtaOh1EOr4xEIV7)** ([raw file](https://github.com/w3c/webdriver/blob/1b8ebfd2316e46ab1d780d0445fff858e9876872/webdriver-spec.html#L4074)):</sup>
RFC 6265 makes the distinction between a ‘non-HTTP’ and ‘HTTP’ API.  Basically the difference is that when retrieving cookies via the HTTP API, cookie’s which `http-only-flag` is true are included.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/w3c/webdriver/214)

andreastt:
<sup>**[webdriver-spec.html, line 4068 \[r2\]](https://reviewable.io:443/reviews/w3c/webdriver/214#-Jxj9LeLWek7vxlWSmZs)** ([raw file](https://github.com/w3c/webdriver/blob/1b8ebfd2316e46ab1d780d0445fff858e9876872/webdriver-spec.html#L4068)):</sup>
The line below should probably say “for the current browsing context’s document”

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/w3c/webdriver/214)

AutomatedTester:
landed in 71bde4f

PR #215, opened 2015-08-25
Cookies: correct fields retrieved from a cookie

andreastt:
The previous version of "serialized cookie" used definitions from the
"receives a cookie" algorithm in RFC 6265 (cookie-attribute-list et
al.) which are only used when a new cookie is created.

Section 5.3 describes what fields the user agent is mandated to store
about a cookie, and we should reference these instead.

The introduction of the new Dependencies sub-chapter is priming Add
Cookie (and partly Delete Cookie) for some upcoming changes where it
would, I think, be tedious to keep extracting fields, especially when
the Cookies specification itself is vague on this topic.

[Review on Reviewable](https://reviewable.io/reviews/w3c/webdriver/215)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -166,10 +166,11 @@ var respecConfig = {
    <li><dfn>cookie</dfn>
    <li><dfn>cookie name</dfn> as a <a>cookie</a>’s <code>name</code> field
    <li><dfn>cookie value</dfn> as a <a>cookie</a>’s <code>value</code> field
-   <li><dfn>cookie path</dfn> as a a <a>cookie</a>’s <code>path</code> field
-   <li><dfn>cookie domain</dfn> as a <a>cookie</a>’s <code>domain</code> field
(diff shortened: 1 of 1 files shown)


AutomatedTester:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[webdriver-spec.html, line 172 \[r1\]](https://reviewable.io:443/reviews/w3c/webdriver/215#-JxjGmQxuRYr6fK3bI-z)** ([raw file](https://github.com/w3c/webdriver/blob/0789f88d511e216e88d6a908edfbfd94316ec158/webdriver-spec.html#L172)):</sup>
http-only was missing from the spec originally, we should add it in now.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/w3c/webdriver/215)

AutomatedTester:
landed in edf1559

PR #216, opened 2015-08-25
Elements: fix incorrect ways of retrieving the web element

andreastt:
Deserializing the web element is only appropriate when the element is
a JSON Object, which is the case if we find something that represents
a web element in the body for Execute Script or Execute Async Script.


AutomatedTester:
Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks pending.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/w3c/webdriver/216)

AutomatedTester:
landed in 5bfe875

PR #220, opened 2015-08-26
Remove custom Status of This Document

andreastt:
The status is auto-generated by ReSpec, which has the same information
listed here.


AutomatedTester:
landed in df8fd72

PR #222, opened 2015-08-26
Get Window Handle: change endpoint to GET /session/{session id}/window

andreastt:
The endpoint for getting a window’s handle is not RESTful as many of
the other commands are.

Thanks to Jim Evans for suggesting.


AutomatedTester:
landed in 6a92c1b

PR #229, opened 2015-08-28
Publish on before_deploy since after_deploy isn't run if deploy isn't…

andreastt:
… defined


AutomatedTester:
landed in fbc86b1

PR #234, opened 2015-08-31
Element Interaction: new chapter

andreastt:
Add preliminary steps for Element Click, Element Tap, Element Clear,
and Element Send Keys.


AutomatedTester:
landed in 8de4ab2

PR #235, opened 2015-09-01
Screen Capture: define Take Screenshot and Take Element Screenshot

andreastt:
A first stab at screen capture.  It’s missing some essential parts, but this provides the basic idea of what we’re trying to achieve which I believe is better than what we have currently.

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -1160,6 +1160,13 @@ URL, and command for each WebDriver <a>command</a>.
  </tr>
 
  <tr>
+  <td><dfn>unable to capture screen</dfn>
+  <td>500
+  <td><code>unable to capture screen</code>
+  <td>A screen capture was made impossible.
+ </tr>
+
+ <tr>
   <td><dfn>unexpected alert open</dfn>
   <td>500
   <td><code>unexpected alert open</code>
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -5213,18 +5224,85 @@ <h3 id="character-types">Character types</h3>
...
+  <a href=https://drafts.fxtf.org/geometry/#rectangle-width-dimension>width dimension</a>.
+
+ <li><p>Let <var>paint height</var> be
+  <var>viewport height</var> – <var>rect</var>’s
+  <a href=https://drafts.fxtf.org/geometry/#rectangle-height-dimension>height dimension</a>.
+
+ <li><p class=issue>Determine if all corners of <var>rect</var> is
+  inside the rectangle that is the viewport.
+
+ <li><p>Let <var>canvas</var> be a new
```
AutomatedTester: This should probably be `complete implementation specific steps to copy the framebuffer...`. Firefox has this ability but I know in the past the Chromium team have said that they don't want to implement anything like this (not sure why but ¯\_(ツ)_/¯)

andreastt: Added that prose.

The point of the algorithms is to give clarity to what the result or output of a command would look like, however.  Not necessarily to define exactly how a user agent might implement something.

AutomatedTester:
landed in 4ef1e6e

PR #236, opened 2015-09-02
Terminology: define UUID

andreastt:


webdriver-spec.html
```diff
@@ -400,7 +400,10 @@
 
 <p>An <dfn>ASCII string</dfn> is a string
  in the range U+0000 to U+007F, inclusive.
-</section>
+
+<p>A <dfn data-lt=uuid>Universally Unique IDentifier (UUID)</dfn> is
+ a 128 bits long URN that requires no central registration process. [[RFC4122]]
```
AutomatedTester: I wonder if we need to be need to explicit and say that we want a UUID Version 4

on page 6 of [RFC4122](https://www.ietf.org/rfc/rfc4122.txt) it says version 4 is
`The randomly or pseudo-randomly generated version specified in this document`

andreastt: I asked about this in #whatwg:

```
17:04 <ato> Does anyone know about UUID’s here?  There are different versions of it (1 though 5) and I’m wondering if it is sufficient to talk about “UUID” (as defined in RFC 4122) as a general concept, or if one should talk specifically about one version.
17:05 <gsnedders> ato: They're all UUIDs and all interchangeable
17:05 <gsnedders> ato: the type is encoded within the UUID
17:05 <gsnedders> ato: generally you pick what type depending on a number of factors (anonmity of the generator, etc.)
17:06 <ato> gsnedders: Does it make sense to pick the type?
17:06 <gsnedders> ato: in a spec? why?
17:07 <ato> gsnedders: To add some context, I don’t care about the interchange between UUIDs of different versions/types, as long as one type is inherently cohesive.
17:07 <gsnedders> ato: they aren't really different "versions", they're different ways to generate a UUID
17:08 <ato> gsnedders: Okay I see.
```

So it doesn’t make sense to talk about the version here unless we need to guarantee that the UUIDs are unique across different implementations, and even then they are supposed to be unique.

AutomatedTester:
landed in 654612e

PR #237, opened 2015-09-02
Terminology: delete prose about what "element" means throughout

andreastt:
It is by this point in time unambiguous what "element" means.


AutomatedTester:
I think that must have got caught up in a previous PR, git is acting like it is a noop

andreastt:
Yes, I appear to have fixed this in another PR.  Closing!

PR #238, opened 2015-09-02
Security: rewrite to normative prose

andreastt:


webdriver-spec.html
```diff
@@ -5346,42 +5346,28 @@ <h3 id="character-types">Character types</h3>
...
-  while the device is connected to a host.
-  An example of this is if the local end needed to go through
-  a remote debugger protocol.
-</aside>
-
-<p>If any of these fail then a
- <code><a href="#status-session-not-created">session not created</a></code>
- error MUST be thrown when the local end tries to create a new session.
+<p>If the <a>remote end</a> and the user agent
+ occupy the same binary executable,
```
AutomatedTester: The following really should happen whether or they use the same binary executable

andreastt: What I’m trying to say here is that if the WebDriver server lives inside the browser’s binary it shouldn’t ship with WebDriver enabled by default.  If the remote end binary is separate, no such restriction is needed.

I’m unsure how to make this clearer… suggestions?

webdriver-spec.html
```diff
@@ -5346,42 +5346,28 @@ <h3 id="character-types">Character types</h3>
...
+ only connections from loopback devices are allowed by default.
+ <a>Remote ends</a> do however support a scenario
+ where the <a>local end</a> runs on a different machine
+ from the <a>remote end</a>.
+
+<p>The <a>remote end</a> must include a configuration option
+ to limit the accepted IP range allowed to connect and make requests.
+ The default setting for this should be
+ to limit connections to the CIDR ranges <code>127.0.0.0/8</code> for IPv4
+ and <code>::1</code> for IPv6. [[RFC4632]]
```
AutomatedTester: This is lacking what should happen if it fails any of the above which was in the original prose.

andreastt: Do you mean the “if any of these fail then a _session not created_ error MUST be thrown” line?

I don’t think that makes any sense since all this talks about are the general security provisions a remote end must have.  If the httpd doesn’t limit connections to loopback by default, it is not in compliance.  If it does conform, the local end will not be able to connect since the subnet is limited.

webdriver-spec.html
AutomatedTester: I honestly think we can drop this since we say `A user agent may rely on a command-line flag` we can drop the `in this case`

andreastt: Fixed in 17ee27a.

AutomatedTester:
landed in d179b0a

PR #239, opened 2015-09-03
User Prompts: define commands and algorithms

andreastt:
This patch specifies the Dismiss Alert, Accept Alert, Get Alert Text,
and Send Alert Text commands, as well as the required infrastructure
for dealing with user prompts.

It also adds steps to handle unexpected alerts for invidiual commands.

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -1555,6 +1555,8 @@ URL, and command for each WebDriver <a>command</a>.
  which is one of <i>none</i>, <i>normal</i>, and <i>eager</i>.
  Unless stated otherwise, it is <i>normal</i>.
 
+<p>A <a>session</a> has an associated <a>unexpected user prompt handler</a>.
+
 <p>When asked to <dfn>close the session</dfn>,
  a <a>remote end</a> must take the following steps:
 
@@ -1618,6 +1620,9 @@ URL, and command for each WebDriver <a>command</a>.
 
  <li><p>If the <a>maximum active sessions</a> is equal to
   the length of the list of <a>active sessions</a>,
+  return <a>error</a> with <a>error code</a> <a>session not created</a>.
+
+ <li><p>If there is a <a>current user prompt</a>,
   return <a>error</a> with <a>error code</a> <a>session not created</a>.
 
  <li><p>Let <var>capabilities</var> be the result of <a>getting a property</a>
@@ -1824,6 +1829,8 @@ URL, and command for each WebDriver <a>command</a>.
  <li><p>If the <a>current top-level browsing context</a> is <a>no longer open</a>,
   return <a>error</a> with <a>error code</a> <a>no such window</a>.
 
+ <li><p><a>Handle any user prompts</a> and return its value if it is an <a>error</a>.
+
  <li><p>Let <var>url</var> be the result of <a>getting a property</a> named "url"
   from the <var>parameters</var> argument.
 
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -5340,26 +5402,146 @@ <h3 id="character-types">Character types</h3>
...
+
+<p class=issue>It’s not clear what interoperability looks like
+ for the ability to switch top-level browsing contexts
+ whilst a user prompt is present.
+
+<p>The <dfn>current user prompt</dfn> is said to be the active
+ <a href=https://html.spec.whatwg.org/#user-prompts>user prompt</a>,
+ which can be one of the entries on the <a>table of simple dialogues</a>,
+ the window.<a href=https://html.spec.whatwg.org/#dom-print><code>print</code></a> dialogue,
+ or a dialogue spawned by <a href=https://html.spec.whatwg.org/#dom-showmodaldialog><code>showModalDialog()</code></a>
```
AutomatedTester: We can drop this since all UAs are dropping support for showModalDialog()

webdriver-spec.html
```diff
@@ -5340,26 +5402,146 @@ <h3 id="character-types">Character types</h3>
...
+  <td>All <a href=https://html.spec.whatwg.org/#simple-dialogs>simple dialogues</a>
+   encountered should be <a>accepted</a>.
+ </tr>
+</table>
+
+<p>In order to <dfn>handle any user prompts</dfn>
+ a <a>remote end</a> must take the following steps:
+
+<ol>
+ <li><p>If there is no <a>current user prompt</a>,
```
AutomatedTester: If there is now alert and someone tries to interact we need to raise `no such alert`

AutomatedTester: re-reading this is it meant for "auto" handling of modals?

webdriver-spec.html
andreastt: Gecko hasn’t yet though, and it will keep working there for ESR 38 until May 2016.  According to MDN, it is still also available in Chrome.

webdriver-spec.html
andreastt: This whole definition is meant for auto-handling prompts, yes.  Basically the match below look for a _user prompt handler_ and if there isn’t one (_missing value default state_), it will return an _unexpected alert open_ error.

webdriver-spec.html
```diff
@@ -5340,26 +5402,146 @@ <h3 id="character-types">Character types</h3>
...
+<p>To <dfn data-lt="dismiss|dismissed|dismisses">dismiss</dfn>
+ the <a>current user prompt</a>,
+ do so as if the user would click the <b>Cancel</b> or <b>OK</b> button,
+ whichever is present, in that order.
+
+<p>To <dfn data-lt="accept|accepted|accepts">accept</dfn> the <a>current user prompt</a>,
+ do so as if the user would click the <b>OK</b> button.
+
+<p>A <dfn>user prompt handler</dfn> is
+ an <a href=https://html.spec.whatwg.org/#enumerated-attribute>enumerated attribute</a>
```
AutomatedTester: Why isn't this done via some capability?

andreastt: So the idea is to set this user prompt handler when we receive the capabilities, but I couldn’t figure out how to add it in a clean way.  This is the reason I’m using a so-called “enumerated attribute” from HTML and define a string keyword lookup for each handler.

AutomatedTester:
landed in 276a514

PR #245, opened 2015-09-04
Dependencies: define URI-TEMPLATE

andreastt:


webdriver-spec.html
```diff
@@ -636,7 +639,7 @@
  <li><p>Remove each entry in <var>endpoints</var>
   for which the concatenation of the <a>url prefix</a>
   and the production in the <i>url</i> column does not match <a>url</a>
-  according to the rules in [[!URI-Template]].
+  according to the rules in [[URI-TEMPLATE]].
```
AutomatedTester: did you mean to remove the normative reference?

andreastt: In retrospect I think probably most, if not all, of the `[[…]]` styled references we use should have a `!` and be normative.

I can fix this instance and will file a separate PR for the rest.

andreastt: Fixed in be9a336910094e5a8b24065bd0954c2ef11d5c38.

AutomatedTester:
landed in 45f3b24

PR #247, opened 2015-09-07
Get Current Url: return the URL of the current browsing context

andreastt:


webdriver-spec.html
```diff
@@ -1888,16 +1888,17 @@
 </table>
 
 <p>The <dfn>Get Current Url</dfn> <a>command</a>
- returns the URL of the <a>current top-level browsing context</a>.
+ returns the URL of the <a>current browsing context</a>.
```
AutomatedTester: This will return the src of a frame which is not what is intended by this call. This is supposed to return what is in the URL bar

andreastt: Did you see http://lists.w3.org/Archives/Public/public-browser-tools-testing/2015JulSep/0035.html?

AutomatedTester:
No I didnt. I thought we had decided on that but can't find anything saying that so will land as is

andreastt:
@AutomatedTester I was actually a bit surprised that this was Selenium’s behaviour.  I think it’s worth having a discussion about as I’d quite like to hear the arguments for why it should be like this.

AutomatedTester:
Closing as it is looking like the spec is right and the OSS project is wrong :tada:

PR #249, opened 2015-09-07
Dependencies: fix double definition of URI Templates

andreastt:


AutomatedTester:
respec picks the first item in the data-lt and checks it with the the innertext so if they are equal then everything is fine, if they are not then it errors.

I have updated your PR to have both and landed in a1b50b6

PR #255, opened 2015-09-09
User Prompts: rename to Set Alert Text and expect string input

andreastt:
There is no added value to simulating key presses to user prompt
dialogues since there are no generated key events from it.

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -1006,7 +1006,7 @@ var respecConfig = {
  <tr>
   <td>POST</td>
   <td>/session/{<var>session id</var>}/alert/text</td>
-  <td><a>Send Alert Text</a></td>
+  <td><a>Set Alert Text</a></td>
  </tr>
 
  <tr>
@@ -5608,7 +5608,7 @@ following must be <var>parameters</var> after the local end has made a request t
    <a>Dismiss Alert</a><br />
    <a>Accept Alert</a><br />
    <a>Get Alert Text</a><br />
-   <a>Send Alert Text</a>
+   <a>Set Alert Text</a>
  </tr>
 </table>
 
@@ -5788,7 +5788,7 @@ following must be <var>parameters</var> after the local end has made a request t
 </section>
 
 <section>
-<h3>Send Alert Text</h3>
+<h3>Set Alert Text</h3>
 
 <table class="simple jsoncommand">
  <tr>
@@ -5801,7 +5801,7 @@ following must be <var>parameters</var> after the local end has made a request t
  </tr>
 </table>
 
-<p>The <dfn>Send Alert Text</dfn> <a>command</a>
+<p>The <dfn>Set Alert Text</dfn> <a>command</a>
  sets the text field prompt to a given value,
  and only makes sense on <a>window.<code>prompt</code></a>.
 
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -969,7 +969,7 @@
  <tr>
   <td>POST</td>
   <td>/session/{session id}/alert/text</td>
-  <td><a>Send Alert Text</a></td>
```
AutomatedTester: `Send` fits in nicely with with `Send Keys`. Why change it to set?

andreastt: Because it no longer implies that ti should use the send keys algorithm that is used for input form controls for entering text into prompts.

This specification never outlined using that algorithm for prompts, but Marionette for example uses it for typing into prompts.  This is completely unnecessary and can’t be supported in other UAs.

andreastt:
@AutomatedTester Rebased

AutomatedTester:
So I don't see this as a we are sending over key events. We are sending over the `keys` that the user wants to type. What we do once it is in the UA shouldn't really influence the name of the API, it should show the user the intent of the call. If we have have `Set Alert Text` then to have symmetry for all other "typing" commands we should have `Set Text` and I dislike that. 

If you are really wanting this then I suggest that we discuss this at TPAC.

shs96c:
I agree with @AutomatedTester on this one. The idea with both send keys and set alert text is that the user is inputting a string of characters somehow. How that's implemented really doesn't matter from the local ends PoV.

andreastt:
It does matter if it implies that it’s entering the keys using the same mechanism used for sending keys to elements.  The only browser that can do this is Firefox because of XUL, but since we cannot standardise this behaviour the specification is currently saying “take this array of characters and join them together in a string”.  I wonder what benefit this added complexity gives us apart from confusion.

andreastt:
The backwards compatibility argument is definitely a strong one, but I wonder if we plan to do the same with other commands that receive textual input in the future?

Thinking about BASIC AUTH dialogues for example, if we decide to support them in the future, would they too receive username/password input as sequences of characters?

My point is that consistency matters and we should be aware of the implications of leaving this as it is.  For neither of the cases we can feasibly support character-by-character input emulation so it is a bit of a deception.

shs96c:
It depends on the platform. If you're using Windows and the SendKey Win32 API, then you can absolutely send each of the characters one at a time. The end point name was chosen to mirror that of the webelement "send keys" API because conceptually the two commands are doing the same thing: sending a stream of characters to the browser.

The skeleton API for handling auth dialogs doesn't assume only BASIC and DIGEST, but also things like ubikeys and fingerprints. It consists of a `Credentials` interface which is passed to an "authenticate" end point. Serialization of those credentials will depend on what the credential is. A username/password combination (which is what's in selenium webdriver right now) would be a JSON blob with those fields. Other credentials might just have a hash or a magic cookie.

andreastt:
@shs96c Thanks for outlining your thoughts on the auth API.  What you describe makes sense.

It’s true that both the Windows APIs and Gecko gives you the _ability_ to use the key-by-key entry **Send Keys** approach.  But it’s not clear what this gives us, and I don’t think anyone is arguing we _should_ use it, since the `window.confirm` API isn’t governed by the DOM and isn’t affected by DOM event listeners.  This is why the spec concatenates the elements of the array and hands it off.

So the two arguments in favour of keeping the API as-is is (1) backwards compatibility and (2) consistency with other commands, i.e. **Send Keys**.  What @shs96c says above is a good API for the authentication and I’m struggling to come up with other use cases where we take a single string field as input for a command.

Does anyone else have examples of where we do that, or examples of cases where we _might_ do that in the future?  I guess if the answer is “no” then the consistency argument gains more weight, and we can close this.

AutomatedTester:
since no examples were added for the future I am closing this out now.

jgraham:
I think what's currently in the spec is crazy. The API is literally "take a string (what any sane client blinding will provide), split at arbitrary points (because we don't make any effort to ensure that what's sent over the wire is a single codepoint/grapheme cluster/whatever), send over the wire, and rejoin to form the original string again".

Is there any reason that this should be character-by-character other than "some implementations might have to implement it that way because there is no other way to interact with an alert prompt"? If it does have to be character-by-character what does that actually mean (c.f. discussion about code units vs grapheme clusters). Also note that in the character-by-character case the spec is still wrong here because it doesn't enforce any notion of a character (i.e. sending `{"text": ["Hello Word"]}` and `{"text": ["H", "e", "l", "l", "o", " ", "W", "o", "r", "l", "d"]}` currently behave identically per spec.

PR #268, opened 2015-09-15
Element Clear: define steps for actually clearing element

andreastt:
Note that this leaves firing an event undefined and does not take
into account `<select>` which may require special treatment.

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -353,6 +353,11 @@ var respecConfig = {
    <!-- window.prompt --> <li><dfn>window.<a href=https://html.spec.whatwg.org/#dom-prompt><code>prompt</code></a></dfn>
   </ul>
 
+ <dd><p>The HTML Editing APIs specification defines the following terms: [[!EDITING]]
+  <ul>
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -2835,6 +2859,25 @@
...
 
+<p>An <a>element</a> is considered <dfn>editable</dfn> if it is
+ an <a><code>input</code> element</a>
+ whose <a><code>type</code> attribute</a> is not in the
+ <a>Button</a>,
+ <a>Hidden</a>,
+ <a>Image Button</a>,
+ <a>Reset Button</a>,
+ or <a>Submit Button</a> states,
+ is a <a><code>textarea</code> element</a>,
```
AutomatedTester: this reads to me that textarea is not an editable element

andreastt: Yes, it’s not very clear at the moment.  Tried to address this in fac6f9f.  Does it read better for you now?

AutomatedTester:
landed in 8432e45

PR #270, opened 2015-09-15
Screen Capture: drop issue about checking corner positions

andreastt:
Testing that one or more of the corners are inside the viewport is
faulty logic, since the bounding box of the element can be greater
and also unatainable by scrolling the element into view.

Using hit testing to verify that some part of the element is visible
is bordering on element visibility checks and is not the intention
of this command.


AutomatedTester:
landed in 1775744

PR #277, opened 2015-10-02
Fix a variable name. At some point in the past, this was renamed from…

juangj:
… "resultant capabilities" to "capabilities result", but this instance was missed.


shs96c:
Merged as cffbe83f3aa0b6fc348493e0ba2c42dd437eb923

PR #278, opened 2015-10-19
Bug 29157 Add /status

AutomatedTester:
This allows the local end to know that the remote end is ready to accept
commands.


juangj:
It still says "endpoint node" in "The name of the platform the <a>endpoint node</a> is currently running on."

But... does it really matter what platform an intermediary node is running on? It seems like all of that information could be optional. Or maybe the form of the response doesn't need to be specified at all, and the spec should only prescribe that the response have a 200 status.

AutomatedTester:
landed in 0e48b33

PR #279, opened 2015-10-28
Elements: define interactibility

andreastt:
Based on resolution from
http://www.w3.org/2015/10/25-webdriver-minutes.html.

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -2905,7 +2905,7 @@ var respecConfig = {
  of its rectangle that is inside the <a>viewport</a>,
  excluding the size of any rendered scrollbars.
 
-<p>In order to produce an <dfn>pointer-interactible elements tree</dfn>
+<p>In order to produce a <dfn>pointer-interactible elements tree</dfn>
  from an <var><a>element</a></var>:
 
 <ol>
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -2947,7 +2947,7 @@ var respecConfig = {
 </ol>
 
 <p>To <dfn>get the first pointer-interactible element</dfn>
- given an <a>pointer-interactible elements tree</a>,
+ given a <a>pointer-interactible elements tree</a>,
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -2882,6 +2888,68 @@
  whose <a><code>readOnly</code> attribute</a> is set.
 
 <section>
+<h3>Element Interactibility</h3>
+
+<p>In order to determine if an element can be interacted with,
+ WebDriver performs hit-testing to find
+ if the interaction will be able to reach the requested element.
```
AutomatedTester: This is only true for elements we are using pointer actions on.

webdriver-spec.html
```diff
@@ -2882,6 +2888,68 @@
...
+
+<p>An <dfn>interactible element</dfn> is defined to be
+ the first non-transparent <a>element</a>,
```
AutomatedTester: what is `non-transparent`?

webdriver-spec.html
```diff
@@ -2882,6 +2888,68 @@
...
+  because elementsFromPoint takes care of it
+
+  this is the same as
+
+    rect.width - (rect.right - innerWidth)
+    rect.height - (rect.bottom - innerHeight)
+ -->
+ <li><p>Let <var>visible width</var> be
+  the <var>element</var>’s <a>bounding rectangle</a>’s
+  (<a>max</a>(<a>x coordinate</a>, <a>x coordinate</a> + <a>width dimension</a>) –
```
AutomatedTester: what is `max`?

webdriver-spec.html
andreastt: This is just the non-normative description, but it means `opacity` set to != "1".

webdriver-spec.html
andreastt: See the definition under _Terminology_.  It’s a way to get the largest value of a variable list of values.

AutomatedTester:
landed in 29ce8d5

PR #281, opened 2015-12-01
Revised draft of interactions section

AutomatedTester:
There are issues in the prose for bits that are not complete but those can be completed as follow ups. The main feedback required is missing sections that have not been covered at all.

You can visit a rendered version of the section on my [gh-pages](http://oss.theautomatedtester.co.uk/webdriver/webdriver-spec.html#interactactions)


webdriver-spec.html
```diff
@@ -4624,636 +4646,562 @@
...
-<p class=issue><strong>Warning:</strong>
- This section has not yet been redefined to match the <a>routing requests</a> model,
- and uses old concepts and definitions.
- Please do not rely on it yet.
+<section><!-- Actions -->
+  <h2>Interactions</h2>
+  The Actions API gives you a mechanism to do atomic interactions by sending over
+  a JSON object that can then be used to <a>dispatch commands</a>.
+  The commands are then <a data-lt=command-processed>processed</a> by the user agent,
+  allowing complex interactions to be created.
```
andreastt: This feels a little unspecific.  I think a nicer phrasing would be that it allows you to _pipeline interaction actions to be performed atomically_ on the remote end, or something along those lines.

webdriver-spec.html
```diff
@@ -4624,636 +4646,562 @@
...
+  <aside class=example>
+    If a person wanted to emulate a user doing <code>click + drag + drop</code>
+    they would use the following section to generate an object that is
+    then sent over to the <a>remote end</a>
```
andreastt: Something missing?

webdriver-spec.html
```diff
@@ -4624,636 +4646,562 @@
...
+  </aside>
+
+  <p class='note'>
+    An <a>activation trigger</a> generated by the WebDriver API User SHOULD be
+    indistinguishable from those generated by a real user interacting with the
+    browser. In particular, the dispatched events will have the
```
andreastt: We can’t have normative text in a note.

webdriver-spec.html
```diff
@@ -4624,636 +4646,562 @@
...
+    <a>isTrusted</a> attribute set to true. The most robust way to dispatch
+    these events is by creating them in the browser implementation itself.
+    Sending OS-specific input messages to the browser's window has the
+    disadvantage that the browser being automated may not be properly isolated
+    from a user accidentally modifying input source state; use of an OS-level
+    accessibility API has the disadvantage that the browser's window must be
+    focused, and as a result, multiple tests cannot run in parallel.
```
andreastt: “Multiple tests in parallel” feels like the wrong vernacular to describe what WebDriver does, but I understand what you’re trying to say.

webdriver-spec.html
```diff
@@ -4624,636 +4646,562 @@
...
-              },
-              { "name": "pointerMove",
-                "ELEMENT": "element3"
-              },
-              { "name": "pointerUp"
-              }
-    ]
-  }
-]</pre>
+    <p>The <dfn>Add Actions</dfn> <a>command</a> allows you to create <dfn>sequential
```
andreastt: I think we should find a better term than _Add Actions_.  Maybe _Perform Actions_ or just _Actions_?

andreastt: _Sequential interactions_ doesn’t need to be a definition since it isn’t referenced anywhere.

webdriver-spec.html
```diff
@@ -4624,636 +4646,562 @@
...
+
+       <p><img src=graphics/note3actions.svg alt="">
+
+       <p>During tick 2, source 1 will have its pointerMove action dispatched,
+        while source 2 will do nothing.
+    </aside>
+
+    <p>
+
+    <p>The following <dfn>table of action commands</dfn> enumerates all supported
```
andreastt: I think calling actions “commands” is misleading because they are not the same type of commands we otherwise talk about in the specification.

What about just calling this the _table of actions_?

webdriver-spec.html
```diff
@@ -4624,636 +4646,562 @@
...
+      <tr>
+      <th>HTTP Method</th>
+        <th>Path Template</th>
+      </tr>
+      <tr>
+        <td>POST</td>
+        <td>/session/{<var>session id</var>}/actions</td>
+      </tr>
+    </table>
+    <p>The <a>remote end steps</a> are:
```
andreastt: First, check if the window is still open and whether there are any open user prompts.

webdriver-spec.html
```diff
@@ -4624,636 +4646,562 @@
...
+      <ol>
+        <li><p>Let <var>actions</var> be the result of <a>getting a property</a>
+          named "<code>actions</code>" from the <var>parameters</var> argument.
+
+        <li><p>If <var>actions</var> is not a JSON List, return <a>error code</a>
+          with <a>invalid argument</a>.
```
andreastt: Return _error_ with error code invalid argument.

webdriver-spec.html
```diff
@@ -4624,636 +4646,562 @@
...
+
+        <li><p>Let <var>length</var> be the length of <var>actions</var>.
+
+        <li><p>Let <var>k</var> be 0.
```
andreastt: I have a profound dislike for these type of for-loops because they are more prone to mistakes, and are more difficult to read, than a for…of style loop.

webdriver-spec.html
```diff
@@ -4624,636 +4646,562 @@
...
-          <li>let <var>modifier</var> be the modifier key represented by
-          <var>key</var></li>
-          <li>If <var>modifier</var> is currently held down:
-        <ol>
-      <li>Generate a <code>keyup</code> event of <var>modifier</var></li>
-    </ol>
-  </li>
-  <li>Else:
+
+            <li><p><a>Dispatch commands</a> with argument <var>action</var>.
```
andreastt: s/action/actions/

webdriver-spec.html
```diff
@@ -4624,636 +4646,562 @@
...
+      <th>Path Template</th>
+    </tr>
+    <tr>
+      <td>DELETE</td>
+      <td>/session/{<var>session id</var>}/actions</td>
+    </tr>
+  </table>
+
+    <p>The <dfn>Release Actions</dfn> <a>command</a> is used to cancel all current
+      <a>action chains</a>. This is the equivalent of releasing all input sources.
```
andreastt: We should point out here that modifier keys are released, etc.

webdriver-spec.html
```diff
@@ -4624,636 +4646,562 @@
...
+    <p>The <a>remote end steps</a> are:
+      <ol>
+
+        <li><p class=issue>This needs to be completed.
+
+        <li><p>Return <a>success</a> with data null.
+      </ol>
+    </section>
+    To be able to <dfn>dispatch commands</dfn> on the remote end we need
+    to perform the following steps:
```
andreastt: I feel this should be closer to the top of this chapter.

andreastt: _Dispatch command_ is an internal algorithm and by the looks of it we’re not passing it a JSON Object.  If we elsewhere say “dispatch command _action_”, we can just reference _action_ directly here I would think.

webdriver-spec.html
```diff
@@ -4624,636 +4646,562 @@
...
+      <p>The <dfn>pause</dfn> command:
+        <ol>
+          <li><p>If the <a>current top-level browsing context</a> is <a>no longer open</a>
+            , return <a>error</a> with <a>error code</a> <a>no such window</a>.
+
+          <li><p><a>Handle any user prompts</a> and return its value if it is an
+            <a>error</a>.
+
+          <li><p>Let <var>duration</var> be the result of <a>getting a property</a>
+            named "<code>duration</code>".
```
andreastt: Duration from where?

webdriver-spec.html
```diff
@@ -4624,636 +4646,562 @@
...
+          named "<code>value</code>".
+
+        <li><p>Let <var>active element</var> be the <a>active element</a> of
+          the <a>current browsing context</a>’s <a>document element</a>.
+
+        <li><p>Let <var>key to type</var> be the result of <a>processing keys</a>
+          with argument <var>key</var>.
+
+        <li><p>Generate <a data-lt=keyUp-Event>keyUp</a> <a>Keyboard Event</a> against
+          <var>active element</var> for <var>key to type</var>.
```
andreastt: If I understand correctly this is not how existing implementations work.  Certainly Marionette doesn’t dispatch the event on the active element.

webdriver-spec.html
```diff
@@ -4624,636 +4646,562 @@
...
+          <a>error</a>.
+
+        <li><p>Let <var>key</var> be the result of <a>getting a property</a>
+          named "<code>value</code>".
+
+        <li><p>Let <var>active element</var> be the <a>active element</a> of
+          the <a>current browsing context</a>’s <a>document element</a>.
+
+        <li><p>Let <var>key to type</var> be the result of <a>processing keys</a>
+          with argument <var>key</var>.
```
andreastt: The key to type needs to consider the current modifier keys pressed, if itself is a modifier key, uppercase if the shift modifier is pressed, compute the key code from the char code, etc.

webdriver-spec.html
```diff
@@ -4624,636 +4646,562 @@
...
-    <p>The state of the physical keyboard MUST NOT affect emulated user input.</p>
-    <h3>Internationalized input</h3>
-    <p>Non-latin symbols: TBD</p>
-    <p>Complex scripts using Input Method Editor (IME): TBD</p>
-  </section> <!-- sendkeys end -->
- </section> <!-- /High-Level APIs -->
-</section> <!-- /Interactions -->
+
+
+    <p>The steps for <dfn>processing keystrokes</dfn> are:
```
andreastt: This isn’t referenced?

andreastt:
I should point out that I haven’t compared this to existing implementations like Marionette yet.  I also haven’t finished reading it, but it’s overall a massive improvement to the status quo.

webdriver-spec.html
AutomatedTester: You are right, I have changed this to `multiple WebDriver instances` which is might still not be correct but is closer to where we want.

andreastt: 👍🏼

webdriver-spec.html
AutomatedTester: I have updated to *Perform Actions*

webdriver-spec.html
AutomatedTester: Updated

webdriver-spec.html
AutomatedTester: updated

webdriver-spec.html
AutomatedTester: WebDriver does this. The client APIs find the API and do some work before doing the action. See https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/common/action_chains.py#L74

However... having atomic actions might make this extra work impossible... will have another think about it...

webdriver-spec.html
AutomatedTester: I have corrected this and will be also used when `sendKeysToElement` is finished off.

AutomatedTester:
Landed in cd0fb1b

I have addressed all of Andreas' comments and since there are no other comments from other people I have felt it was better to land and then get people implementing it to report errors. 

It also creates the ability for others to submit PRs based on issues they are seeing when implementing.

PR #283, opened 2015-12-07
Return window size from getter

andreastt:
Fixes bug 29324.


andreastt:
@AutomatedTester I believe this is ready for merging.

AutomatedTester:
landed in 9dc352d

PR #285, opened 2016-01-11
Add in Get Page Source

AutomatedTester:
This adds in the ability to get the source of the current document.

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -222,6 +222,9 @@ var respecConfig = {
   the DOM Parsing and Serialisation specification: [[!DOM-PARSING]]
   <ul>
    <li><dfn><a href=https://w3c.github.io/DOM-Parsing/#widl-Element-innerHTML><code>innerHTML</code> IDL attribute</a></dfn>
+   <li><dfn><a href=https://w3c.github.io/DOM-Parsing/#widl-Element-outerHTML><code>outerHTML</code> IDL attribute</a></dfn>
+   <li><dfn><a data-lt="xmlserializer" href=https://w3c.github.io/DOM-Parsing/#the-xmlserializer-interface><code>XMLSerializer</code> Interface</a></dfn>
+   <li><dfn><a href=https://w3c.github.io/DOM-Parsing/#widl-XMLSerializer-serializeToString-DOMString-Node-root><code>serializeToString</code></a></dfn>
   </ul>
 
   <dd><p>The following attributes are defined in the UI Events
@@ -985,6 +988,12 @@ var respecConfig = {
   <td>GET</td>
   <td>/session/{<var>session id</var>}/element/{<var>element id</var>}/enabled</td>
   <td><a>Is Element Enabled</a></td>
+ </tr>
+
+ <tr>
+  <td>POST</td>
+  <td>/session/{<var>session id</var>}/source</td>
+  <td><a>Get Page Source</a></td>
  </tr>
 
  <tr>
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -222,6 +222,9 @@
   the DOM Parsing and Serialisation specification: [[!DOM-PARSING]]
   <ul>
    <li><dfn><a href=https://w3c.github.io/DOM-Parsing/#widl-Element-innerHTML><code>innerHTML</code> IDL attribute</a></dfn>
+   <li><dfn><a href=https://w3c.github.io/DOM-Parsing/#widl-Element-outerHTML><code>outerHTML</code> IDL attribute</a></dfn>
+   <li><dfn><a data-lt="xmlserializer" href=https://w3c.github.io/DOM-Parsing/#the-xmlserializer-interface><code>XMLSerializer</code> Interface</a></dfn>
+   <li><dfn><a href=https://w3c.github.io/DOM-Parsing/#widl-XMLSerializer-serializeToString-DOMString-Node-root><code>serializeToString</code></a></dfn>
```
andreastt: Should probably sort this alphabetically.

webdriver-spec.html
```diff
@@ -989,6 +992,12 @@
 
  <tr>
   <td>POST</td>
```
andreastt: GET?

andreastt: GET?

webdriver-spec.html
```diff
@@ -4010,7 +4019,43 @@
...
+  <th>HTTP Method</th>
+  <th>Path Template</th>
+ </tr>
+ <tr>
+  <td>GET</td>
+  <td>/session/{<var>session id</var>}/source</td>
+ </tr>
+</table>
+
+<p>The <dfn>Get Page source</dfn> <a>command</a> returns the HTML source of the
```
andreastt: s/source/Source/

andreastt: Also it doesn’t return the HTML source of the document, but a string serialisation of the DOM.

webdriver-spec.html
```diff
@@ -4010,7 +4019,43 @@
...
+  document as it is at the time of the call.
```
andreastt: Of the _current browsing context_, and I’m not sure we need to say “at the time of the call” since this is implicit for all WebDriver commands.

webdriver-spec.html
```diff
@@ -4010,7 +4019,43 @@
...
+
+<p>The <a>remote end steps</a> are:
+
+<ol>
+ <li><p>If the <a>current top-level browsing context</a> is <a>no longer open</a>,
+  return <a>error</a> with <a>error code</a> <a>no such window</a>.
+
```
andreastt: Do we want to handle user prompts here?  We do for _Get Current URL_, but it might be we could remove it from there.

webdriver-spec.html
```diff
@@ -4010,7 +4019,43 @@
...
+ <li><p>Let <var>source</var> be the result returned from the
+   <a><code>outerHTML</code> IDL Attribute</a> of the <a>Document Element</a>.
```
andreastt: … of the current browsing context’s active document.

andreastt: Also, if I do `document.documentElement.outerHTML` in a content-less document I get a _TypeError: document.documentElement is null_.  Not sure if it’s supposed to throw in that case, and this could be a Gecko bug.

webdriver-spec.html
```diff
@@ -4010,7 +4019,43 @@
...
+
+ <li><p>If <var>source</var> is <a>null</a>:
+   <p>Let <var>source</var> be the result of calling <code>serializeToString()</code>
+     on the <code>XMLSerializer</code> Interface.
```
andreastt: Here I think you could just say to [produce an XML serialisation](https://w3c.github.io/DOM-Parsing/#dfn-concept-serialize-xml) and be done with it.

webdriver-spec.html
```diff
@@ -4398,7 +4443,7 @@
 </ol>
 </section> <!-- /Execute Async Script -->
 </section> <!-- /Executing Script -->
-
+</section> <!-- /Document Handling -->
```
andreastt: This is in the wrong place.  Since your new chapter _Document Handling_ comes directly after _Element Interaction_, it should be closed **before** _Executing Script_ begins.

webdriver-spec.html
AutomatedTester: Unfortunately I can't use that since that requires `well formed` to be true where `serialiseToString` sets that to false.

webdriver-spec.html
AutomatedTester: done

webdriver-spec.html
AutomatedTester: corrected

webdriver-spec.html
AutomatedTester: corrected

webdriver-spec.html
AutomatedTester: True, I was just thinking about the amount of users who I have helped understand what Page Source returns since its not the HTML that is initially sent over the wire. They forget that their JS could be mutating the DOM.

Updated as suggested

webdriver-spec.html
AutomatedTester: added it in for symmetry with other calls. I dont think it is necessary here but lets keep calls similar

webdriver-spec.html
AutomatedTester: changed it in a way you suggested but using what I had originally.

AutomatedTester:
landed in 0bece03da3b699fe620ff98a5472036b36c879f7

PR #289, opened 2016-02-09
Complete interaction steps for element.sendKeys()

AutomatedTester:
This documents what events need to be fired and when during the process
of sending keys to an element.

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -174,6 +174,7 @@ var respecConfig = {
       <li><dfn><a href="https://w3c.github.io/uievents/#glossary-activation-trigger">Activation Trigger</a></dfn>
       <li><dfn><a href="https://w3c.github.io/uievents/#interface-KeyboardEvent">Keyboard Event</a></dfn>
       <li><dfn data-lt="keyUp-Event"><a href="https://w3c.github.io/uievents/#event-type-keyup">keyUp Event</a></dfn>
+      <li><dfn data-lt="keyPress-Event"><a href="https://w3c.github.io/uievents/#event-type-keypress">keyPress Event</a></dfn>
       <li><dfn data-lt="keyDown-Event"><a href="https://w3c.github.io/uievents/#event-type-keydown">keyDown Event</a></dfn>
       <li><dfn><a href="https://w3c.github.io/uievents/#events-keyboard-event-order">Keyboard Event Order</a></dfn>
       <li><dfn><a href="https://w3c.github.io/uievents/#interface-MouseEvent">Mouse Event</a></dfn>
@@ -263,6 +264,7 @@ var respecConfig = {
    <!-- Event loop --> <li><dfn><a href=https://html.spec.whatwg.org/#event-loop>Event loop</a></dfn>
    <!-- Focusing steps --> <li><dfn><a href="https://html.spec.whatwg.org/#focusing-steps">Focusing steps</a></dfn>
    <!-- Hidden --> <li><dfn><a href="https://html.spec.whatwg.org/#hidden-state-%28type=hidden%29">Hidden</a></dfn> state
+   <!-- File --> <li><dfn><a href="https://html.spec.whatwg.org/#file-upload-state-%28type=file%29">File</a></dfn> state
    <!-- Image Button --> <li><dfn><a href="https://html.spec.whatwg.org/#image-button-state-%28type=image%29">Image Button</a></dfn> state
    <!-- Joint session history --> <li><dfn><a href=https://html.spec.whatwg.org/multipage/browsers.html#joint-session-history>Joint session history</a></dfn>
    <!-- Missing value default state --> <li><dfn><a href=https://html.spec.whatwg.org/#missing-value-default>Missing value default state</a></dfn>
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -4105,7 +4108,46 @@
 
  <li><p><a>Scroll into view</a> the <var>element</var>.
 
- <li><p class=issue>Interact with the element.
+ <li><p>Run the <a>focusing steps</a> for the <a>element</a>.
+
+ <li><p>Set the cursor position to after the last character in the element
```
lukeis: This doesn't feel like a good definition of where the cursor is placed. I'm at a bit of a loss for figuring out a good alternative. Because we need to imply either the end of the element's Value or the end of the Text in the element (for contenteditable elements).

webdriver-spec.html
```diff
@@ -4105,7 +4108,46 @@
...
+
+      <li><p>Generate <a data-lt=keyDown-Event>keyDown</a> <a>Keyboard Event</a> against
+        <var>active element</var> for <var>key to type</var>.
+
+      <li><p>Generate <a data-lt=keyPress-Event>keyPress</a> <a>Keyboard Event</a> against
+        <var>active element</var> for <var>key to type</var>.
+
+      <li><p>Generate <a data-lt=keyUp-Event>keyUp</a> <a>Keyboard Event</a> against
+        <var>active element</var> for <var>key to type</var>.
+    </ol>
```
lukeis: I think we also should talk about the focus / blur and change events that occur (or should be fired here too)

AutomatedTester: We handle the `blur` and `focus` in the `focusing steps`

webdriver-spec.html
andreastt: First of all, we probably want to reconcile sending key input from this command and from an action chain to a single definition.

Second, you only want to move the cursor to the end for this command, as the action API is supposed to enter keys wherever the current cursor is.

OTOH this algorithm needs to take into account not only `<input>` and `contenteditable` elements, but also `<textarea>`.  For `<input>` and `<textarea>` the following can be used:

```
let t = el.type;
if (t && (t == "text" || == "textarea")) {
  let len = el.value.length;
  el.setSelectionRange(len, len);
}
el.focus();
```

`contenteditable` shares some of the same challenges as when `document.designMode` is enabled.  I would suggest documenting the other things first, and file a bug to address editable elements later.

webdriver-spec.html
```diff
@@ -4105,7 +4108,46 @@
...
 
- <li><p class=issue>Interact with the element.
+ <li><p>Run the <a>focusing steps</a> for the <a>element</a>.
+
+ <li><p>Set the cursor position to after the last character in the element
+
+ <li><p>Let <var>character array</var> be the result of <a>getting a property</a>
+    <code>text</code> from the <var>parameters</var> argument.
+
+ <li><p>If <var>element</var> is a <a><code>input</code></a> whose
```
andreastt: s/a/an/

webdriver-spec.html
```diff
@@ -4105,7 +4108,46 @@
...
+   <a><code>type</code> attribute</a> is <a>File</a>:
+   <ol>
+     <li><p>Let <var>text</var> be the result of
+      joining together each element in <var>character array</var> to a string.
+
+     <li><p>Set the <a><code>value</code> Attribute</a> to <var>text</var>.
```
andreastt: This isn’t what current implementations are doing.  We need to take into account that the HTML specification doesn’t allow you to set the value of `<input type=file>`.

This necessitates implementation-specific steps and we ought to call out this out.

For example the Gecko implementation does the following in privileged chrome space:

```
let val = "file.txt";
let fs = Array.prototype.slice.call(fileInputElement.files);
let file;
try {
  file = new File(val);
} catch (e) {
  throw new InvalidArgumentError("File not found: " + val);
}
fs.push(file);
el.mozSetFileArray(fs);
```

So:

* Add test for file existence
* Allow multiple files to be appended to existing array of files on element

webdriver-spec.html
```diff
@@ -4105,7 +4108,46 @@
...
+   </ol>
+
+  <li><p>Let <var>i</var> be 0.
+
+  <li><p>Let <var>length</var> be the length of <var>character array</var>.
+
+  <li><p>While <var>i</var> &lt; <var>length</var>:
+
+    <ol>
+      <li><p>Let <var>char</var> be the entry in <var>character array</var> at
```
andreastt: Can we please not use iterators with indexes?  It’s so hard to read for no apparent purpose.

burg:
Some miscellaneous comments:

* The algorithm refers to a 'cursor' but this doesn't make sense when sending keys. Please don't call it a cursor. It's a text insertion caret, or shorter, an insertion caret or just 'caret'. Maybe its behavior should be delegated to other specs, like the Selection API, Contenteditable API, or CSS-UI. Also the spec needs to be careful when saying things like the 'end' of text since this verbiage is ambiguous when dealing with RTL or BIDI text.

* The use of 'character' is ambiguous. Is it a Unicode code point? A grapheme cluster? Should UAs subject this key input to IME handling or not? An example of a complex text case would be really helpful to clear the water here, but the text needs to be unambiguous too. Would you like some test cases?

AutomatedTester:
landed in de28aa7

PR #290, opened 2016-03-10
Rename Get command to Go

andreastt:


AutomatedTester:
what is the reason behind this?

andreastt:
https://www.w3.org/Bugs/Public/show_bug.cgi?id=29520

AutomatedTester:
landed in 891bfd1, d46720d

shs96c:
We should roll this back. The command is named after the HTTP verb "GET", and is widespread use throughout existing webdriver implementations.

andreastt:
@burg’s argument is that it’s confusing exactly for that reason, because it’s actually a POST.

shs96c:
But it results in the browser executing a GET request, as opposed to a POST. For implementations of webdriver that don't use HTTP as the transport mechanism the ambiguity doesn't exist.

Admittedly, the spec is defining an HTTP protocol, but that's to ensure interop. I can imagine later levels of the spec using (for example) a standardised debugging protocol.

uchuugaka:
Personally, I always thought it should be named request.
Because it is a request.

andreastt:
Personally I feel perhaps the most descriptive name would be _Navigate_, as this is the [algorithm we’re calling down to](https://html.spec.whatwg.org/#navigate) in HTML. But as I said earlier on the bug, I really don’t have any strong opinions on the matter.

andreastt:
Also this change has already landed, so I’m closing the PR. We can consider reverting the change.

PR #293, opened 2016-04-05
protocol: prevent response result error fallthrough

andreastt:
The current wording lets it send two responses when the response result
fails: First the error, and then directly afterwards a success with the
same data.

This change prevents the algorithm from doing this by jumping back to
step 1 direclty afterwards sending the error, as above.


AutomatedTester:
f1c2cbd

PR #294, opened 2016-04-07
Frame same tree check

andreastt:
Resolves one issue and adds a couple of dependencies.


AutomatedTester:
landed in 9e09ab3, 9cefe74, f16f6d4

PR #295, opened 2016-04-07
Window serialisation

andreastt:
https://lists.w3.org/Archives/Public/public-browser-tools-testing/2016AprJun/0005.html

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -2301,10 +2301,36 @@ session := new_session(capabilities)</code></pre>
 <p>The <a>top-level browsing context</a> is said
  to be <dfn>no longer open</dfn> if it has been <a>discarded</a>.
 
-<p>Each <a>top-level browsing context</a>
- has an associated <dfn data-lt="window handle|window handles">window handle</dfn>,
- which is a string uniquely identifying that browsing context.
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -2306,6 +2298,40 @@
...
+
+<p>Each <a>browsing context</a> has an associated
+ <dfn data-lt="window handle|window handles">window handle</dfn> (string)
+ uniquely identifying it.
+ It must not be "<code>current</code>".
+
+<p>The <dfn>web window identifier</dfn>
+ is the string constant "<code>window-fcc6-11e5-b4f8-330a88ab9d7f</code>".
+
+<p>The <dfn>web frame identifier</dfn>
```
AutomatedTester: How would this be used? Since we can't switch to a frame like this it seems like it's not needed.

andreastt: As I [explained in my email](https://lists.w3.org/Archives/Public/public-browser-tools-testing/2016AprJun/0005.html) it lets clients infer type information so they would know not to accept a frame as argument to *Switch To Window*.

AutomatedTester:
landed in c84ab45, 484a922

PR #298, opened 2016-04-08
element retrieval: improve intro wording

andreastt:


webdriver-spec.html
```diff
@@ -3039,16 +3040,18 @@
 <section>
 <h2>Element Retrieval</h2>
 
-<p>The <a>Find Element</a> and <a>Find Elements</a> <a>commands</a>
- allows look up of individual and a collection of <a>elements</a>, respectively.
+<p>The <a>Find Element</a> and <a>Find Elements</a> commands
+ allow look up of individual- and collections of <a>elements</a>.
```
burg: I find this construction less clear than 'individual elements and collections of elements, respectively'.

webdriver-spec.html
```diff
@@ -3039,16 +3040,18 @@
...
+ Element retrieval searches are performed
```
burg: You have not tied the term 'Element Retrieval' to a definition/command. So this does not follow from the previous sentence.

webdriver-spec.html
```diff
@@ -3039,16 +3040,18 @@
...
+ using pre-order traversal of the document’s nodes
```
burg: Is there a reason to specify the traversal order, as opposed to a sort order on the results? Specifying a traversal seems overly restrictive to how selector matching might be performed.

andreastt: This is just a description of how querySelectorAll works, but I’m happy to leave this undescribed if you think that’s a better idea?

webdriver-spec.html
andreastt: It’s the name of the chapter, not a definition as such?

webdriver-spec.html
andreastt: I agree.  Fixed in 926aa874ce2a195fab4cf4c2ea34c9cd595a2daf.

AutomatedTester:
landed in fe1c2d5

PR #299, opened 2016-04-21
added new error code for element intercepting click

Jarob22:
Creating a pull request for this new error

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/299)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -3946,6 +3946,9 @@ following must be <var>parameters</var> after the local end has made a request t
  <li><p>Let <var>centre point</var> be a pair of <var>rect</var>’s
   (<a>rectangle x coordinate</a> + <var>horizontal centre offset</var>)
   and (<a>rectangle y coordinate</a> + <var>vertical centre offset</var>)
+  
+ <li><p>If <a>command</a> is intercepted by an <var>element</var> with a higher 
+  z-index, an <a>element intercepting click</a> <a>error</a> is returned.
 
  <li><p class=issue>Run interaction steps on coordinate.
 
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -1136,6 +1136,15 @@ var respecConfig = {
    e.g. attempting to click an element
    that is no longer attached to the <a>document</a>.
  </tr>
+ 
+ <tr>
+  <td><dfn>element intercepting click</dfn>
+  <td>400
+  <td><code>element intercepting click</code>
+  <td>An element <a>command</a> could not be completed
+   because there is an element in front of it intercepting
+   the command.
+ </tr>
 
  <tr>
   <td><dfn>invalid selector</dfn>


lukeis:
I think we need to add prose into the Element Click section about raising this exception -

https://github.com/w3c/webdriver/blob/master/webdriver-spec.html#L3937-L3941

Jarob22:
@lukeis Done. :)

webdriver-spec.html
```diff
@@ -3946,6 +3946,9 @@
  <li><p>Let <var>centre point</var> be a pair of <var>rect</var>’s
   (<a>rectangle x coordinate</a> + <var>horizontal centre offset</var>)
   and (<a>rectangle y coordinate</a> + <var>vertical centre offset</var>)
+  
+ <li><p>If <a>command</a> is intercepted by an <var>element</var> with a higher 
+  z-index, an <a>element intercepting click</a> <a>error</a> is returned.
```
burg: Stacking order is way more complicated than "compare z-index". A click could be "intercepted" by an element with lower z-index. I'm not sure what intercept means, could you define it more precisely? Is the "intercept" test based on DOM event handling or is it unrelated to existing hit testing/event dispatch mechanisms?

lukeis: We might want to reference elementsFromPoint. Where this exception would get thrown if the originating element is not in the set returned from that function provided the coordinates just calculated. 

http://www.w3.org/TR/cssom-view-1/#extensions-to-the-document-interface

webdriver-spec.html
```diff
@@ -1136,6 +1136,15 @@
...
    that is no longer attached to the <a>document</a>.
  </tr>
+ 
+ <tr>
+  <td><dfn>element intercepting click</dfn>
+  <td>400
+  <td><code>element intercepting click</code>
+  <td>An element <a>command</a> could not be completed
+   because there is an element in front of it intercepting
+   the command.
```
andreastt: It’s not clear to me what “intercepting the command” means. Also I think we should point out explicitly _when_ this error would occur, e.g. for which specific commands.

webdriver-spec.html
andreastt: Basically `elementsFromPoint` gets the paint ordering in which each element has been drawn. This seems like a more sound basis to use when determining if an element that is attempted clicked is actually clickable.

However, as @burg points out, whether an element is immediately clickable doesn’t say anything about whether the click bubbles down to an underlying element.

I think it is a conceptually good change to have a separate error for when another element is on top and a click won’t fulfil, but I’m unsure how we can effectively describe this.

Jarob22: I think this conversation has gotten a bit out of hand when putting it in the perspective of what I'm actually doing here. I'm not creating a new end point or a new way to detect whether an element click has been intercepted, I am merely wanting to change the output error of what happens when an element click is supposed to succeed after passing all the checks in https://w3c.github.io/webdriver/webdriver-spec.html#element-click but before point 14 it fails because another element was in the way. At the moment, it is a very unhelpful "unknown error" which cannot be caught specifically. 

I do not know the ins and outs of the stacking order and element visibility/presence algorithms and am not sure why I'm being asked all these questions about them. I don't mean to be snappy or brusque, and I do want to get this change through, I just don't know the answers to half these questions and I'm quite confused as to why there is a much larger discussion around this than I'd thought initially. 

Cheers,
James

shs96c:
I've started a [discussion](https://lists.w3.org/Archives/Public/public-browser-tools-testing/2017JanMar/0017.html) about this on the mailing list.

AutomatedTester:
This was redone in #525

PR #302, opened 2016-05-05
Locator strategy

AutomatedTester:
This pull request moves over all the remaining locator strategies to the current style of the specification.

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -3139,17 +3139,6 @@ session := new_session(capabilities)</code></pre>
     <li><p>Return <var>elements</var>.
   </ol>
 
-<p class=issue><strong>Warning:</strong>
- This rest of this section has not yet been redefined to match the <a>routing requests</a> model,
- and uses old concepts and definitions.
- Please do not rely on it yet.
-
-<section>
-<h3>ECMAScript</h3>
-
-<p>Finding elements by ecmascript is covered in the <a href="#ecmascript">ecmascript part of this spec</a>.
-</section>
-
 <section>
 <h3>Link Text</h3>
 
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -3120,7 +3120,6 @@ session := new_session(capabilities)</code></pre>
 
 
 
-<section>
 <h3>CSS Selectors</h3>
 
 <p>To find a <a>WebElement</a> with the <dfn>CSS Selector</dfn> <a>strategy</a>
@@ -3138,9 +3137,7 @@ session := new_session(capabilities)</code></pre>
           with the <a>context object</a> equal to the <var>start node</var>.
       </dl>
     <li><p>Return <var>elements</var>.
-
   </ol>
-</section> <!-- /CSS Selector -->
 
 <p class=issue><strong>Warning:</strong>
  This rest of this section has not yet been redefined to match the <a>routing requests</a> model,
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -460,7 +460,15 @@ var respecConfig = {
   <ul>
    <li><dfn data-lt="supported property index"><a href=http://heycam.github.io/webidl/#dfn-supported-property-indices>Supported property indices</a></dfn>
    <li><dfn><a href="https://heycam.github.io/webidl/#syntaxerror"><code>SyntaxError</code></a></dfn></li>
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -459,7 +460,15 @@
   <ul>
    <li><dfn data-lt="supported property index"><a href=http://heycam.github.io/webidl/#dfn-supported-property-indices>Supported property indices</a></dfn>
    <li><dfn><a href="https://heycam.github.io/webidl/#syntaxerror"><code>SyntaxError</code></a></dfn></li>
+   <!-- DOMException --><li><a href="https://heycam.github.io/webidl/#dfn-DOMException"><dfn><code>DOMException</code></dfn></a>
   </ul>
+
+  <dt>XPATH
+  <dd><p>The following terms are defined in the Document Object Model XPath standard [[!XPATH]]
```
andreastt: This whole section should go under the DOM section further up.

webdriver-spec.html
```diff
@@ -3138,96 +3148,75 @@
...
-
   </ol>
-</section> <!-- /CSS Selector -->
 
-<p class=issue><strong>Warning:</strong>
- This rest of this section has not yet been redefined to match the <a>routing requests</a> model,
- and uses old concepts and definitions.
- Please do not rely on it yet.
+<h3>Link Text</h3>
+<p>To find a <a>WebElement</a> with the <dfn data-lt="link text selector">Link Text</dfn>
```
andreastt: WebElement → web element

webdriver-spec.html
```diff
@@ -3138,96 +3148,75 @@
...
+   <a>strategy</a> the following steps need to be completed:
 
-<section>
-<h3>ECMAScript</h3>
+<ol>
+  <li><p>let <var>elements</var> be the result of calling <a>querySelectorAll</a>,
+    with argument <a>a elements</a>, with the <a>context object</a> equal to the
+    <var>start node</var>.
```
andreastt: let → Let, also applies in the CSS selector section above

webdriver-spec.html
```diff
@@ -3138,96 +3148,75 @@
...
-            result = []
-            for element in elements:
-              text = element.text
-              if text == search_term:
-                result.append(element)
-          </pre>
-          <p>Where "search_term" is the link text being searched for, and "result" contains the list of elements to return.</p>
-    </div>
-</section> <!-- /Link Text -->
+<p>The <dfn data-lt="partial link text selector">Partial Link text</dfn> <a>strategy</a>
```
andreastt: Link → link

webdriver-spec.html
```diff
@@ -3138,96 +3148,75 @@
...
+  is very similar to the <a>Link Text</a> <a>strategy</a>, but rather than matching
+  the entire string, only a subset needs to match. That is, this MUST return all
```
andreastt: subset → substring

webdriver-spec.html
```diff
@@ -3138,96 +3148,75 @@
...
+  <a>a elements</a> with rendered text that partially or completely and case
```
andreastt: I think we should codify this to `<a><code>a</code> elements</a>` for clarity throughout.

webdriver-spec.html
andreastt: No normative text in descriptions. This is what the algorithms are for.

webdriver-spec.html
andreastt: “Partially or completely and case sensitively” is a bit much I think. It’s enough to say “… with rendered text that contains the selector expression”.

webdriver-spec.html
```diff
@@ -3138,96 +3148,75 @@
...
-            result = []
-            for element in elements:
-              text = element.text
-              if text.search(search_term) != -1:
-                result.append(element)
-          </pre>
-          <p>Where "search_term" is the link text being searched for, and "result" contains the list of elements to return.</p>
-    </div>
-</section> <!-- /Partial Link Text -->
+        <li><p>If the result of <a>indexOf</a> on <var>rendered text</var> with argument
```
andreastt: I don’t think we should mandate the use of _indexOf_. Just say it’s a string that must contain something else. Implementations may not be in JS.

webdriver-spec.html
AutomatedTester: These are not in the DOM spec, the DOM Spec doesnt deal with XPath as far as I can see.

AutomatedTester:
fixed all comments and have landed.

PR #303, opened 2016-05-10
cookies: error on attempting to delete non-existing cookie

andreastt:


AutomatedTester:
Could you expand on the reasoning for this? This has webdriver acting very differently from the web platform.

If I do 
```js
document.cookie = "foo=bar; expires=Thu, 01 Jan 1970 00:00:00 GMT";
```

I would not get an error as we just hand off to the cookie jar which just sees this as something that it shouldnt care about. In [delete cookies](http://w3c.github.io/webdriver/webdriver-spec.html#dfn-delete-cookies) the UA *could* short cut the process and just do something similar without needing to get all the cookies associated with the active document.

AutomatedTester:
@andreastt can you comment or should I close?

andreastt:
I think what you say makes sense. I have a sneaking feeling that the count of deleted cookies is not something we can reliably track.

You also correctly point out that implementations may not even follow the exact steps in this context. (Marionette for example uses the Gecko cookie service in chrome space.)

PR #306, opened 2016-06-16
Fix up some wording about processing capabilities.

juangj:
In steps say to "append a string", specify that the string is appended
to the 'unmet capabilities' list.

Move the check for mismatched required capabilities and server
capabilities earlier. The result of the algorithm should be unchanged,
but it avoids the awkward run-on sentence.

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -1491,6 +1491,8 @@ session := new_session(capabilities)</code></pre>
 
  <li><p>Let <var>length</var> be the length of <var>required capabilities</var>.
 
+ <li><p>Let <var>unmet capabilities</var> be an empty JSON List.
+
  <li><p>Let <var>k</var> be 0.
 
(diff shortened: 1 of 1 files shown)


AutomatedTester:
This part of the spec is going to be discussed at the next F2F so would like to leave this PR until then

juangj:
Were any decisions about this made at the last F2F? I recall the topic coming up in the minutes, but it wasn't clear to me that any decisions were actually made. Was it just deferred until TPAC, where Simon could discuss in person?

AutomatedTester:
defered to the next face to face

shs96c:
#327 addresses the same area in a way that plenty of people have been looking at. Perhaps provide some input on that?

juangj:
Yes, I will -- this PR wasn't really about changing the substance but just fixing up awkward wording. But no point, if the text is being rewritten regardless.

PR #307, opened 2016-06-27
Define set- and get window position commands

andreastt:
These are implemented in most drivers, exists in Selenium’s JSON wire protocol, but appears to have been lost in the transition to the spec.


AutomatedTester:
r+ could you rebase please?

andreastt:
Thanks!

PR #308, opened 2016-06-28
Make abstraction for window size

andreastt:
Set Window Size returns the size of the window, but does so by calling
Get Window Size, which steps duplicate many of those already run. This
patch addresses this by abstracting the window size concept.


AutomatedTester:
r+ could you rebase please?

andreastt:
Thanks!

PR #310, opened 2016-06-29
Split Get Cookie into two distinct endpoints

andreastt:
Get Cookie previously accepted an optional URL parameter, that if
set would return a list of a single cookie if a matching cookie was
found. This patch changes it so that there is a separate endpoint for
getting the single cookie to avoid confusion as our URL matching does
not really support optional parameters. This is not a functional change.

Additionally this patch introduces a breaking chance: Instead of returning
a list with a single cookie or an empty list, it introduces a "no such
cookie" error when no matching cookie is found.

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -1012,8 +1012,14 @@ var respecConfig = {
 
  <tr>
   <td>GET</td>
+  <td>/session/{<var>session id</var>}/cookie</td>
+  <td><a>Get All Cookies</a></td>
+ </tr>
+
+ <tr>
+  <td>GET</td>
   <td>/session/{<var>session id</var>}/cookie/{<var>name</var>}</td>
-  <td><a>Get Cookie</a></td>
+  <td><a>Get Named Cookie</a></td>
  </tr>
 
  <tr>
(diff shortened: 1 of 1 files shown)


andreastt:
@AutomatedTester If we think breaking backwards compatibility here is problematic, though it is an easy client-side fix, it is possible to change this so that both _Get Named Cookie_ and _Get All Cookies_ returns a JSON List.

Regardless the change to split these into separate endpoints is sound.

webdriver-spec.html
```diff
@@ -1231,6 +1237,15 @@
  </tr>
 
  <tr>
+  <td><dfn>no such cookie</dfn>
```
AutomatedTester: Unfortunately the cookie stuff never errors, unless you are doing cross domain, why are we introducing this when it not how the platform works and breaking change?

andreastt: In Gecko we typically exhaust the iterator in the [cookie manager](https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XPCOM/Reference/Interface/nsICookieManager) to serialise a list of cookies. Since we for the suggested _Get Named Cookie_ match on a user-given path we will know if our sequence of cookies has a cookie with that path or not. Based on that we can choose to error.

AutomatedTester:
landed in 83b4464

PR #311, opened 2016-06-29
Define the property name for the PAC URL.

juangj:
The variable "proxy autoconfiguration url" is referenced but never
defined. Marionette is currently looking for "pacUrl", which seems like
a fine name.

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -1595,9 +1595,13 @@ session := new_session(capabilities)</code></pre>
 
   <dl class=switch>
    <dt>"<code>pac</code>"
-   <dd><p>If the implementation supports <a>proxy autoconfiguration</a>,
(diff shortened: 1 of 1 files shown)


juangj:
On the other hand, the OSS dialect uses "proxyAutoconfigUrl". *shrug*

andreastt:
@juangj I think `proxyAutoconfigUrl` is fine too. Happy to change Marionette. Filed [bug 1283459](https://bugzilla.mozilla.org/show_bug.cgi?id=1283459).

AutomatedTester:
could you update to have `proxyAutoconfigUrl` instead of pacURL

juangj:
Done.

andreastt:
This is now fixed in Nightly for Marionette. Will uplift to Beta and Aurora.

PR #312, opened 2016-07-07
Trim leading and trailing white space from rendered text when looking…

AutomatedTester:
… for link text

This allows WebDriver users to do
```
driver.find_element(By.LINK_TEXT "a link with a trailing space")
```
and have it return back
```
<a href=#>a link with a trailing space </a>
```


andreastt:
LGTM.

andreastt:
@AutomatedTester Should we also do this for partial link text selectors?

njasm:
Hi, it make's sense to me.

AutomatedTester:
> @AutomatedTester Should we also do this for partial link text selectors?

Since we are essentially doing `indexOf` I dont think we really need it there.

PR #314, opened 2016-07-18
Error on serialising window of discarded browsing context

andreastt:
This patch resolves an action from the WG F2F in July 2016.

When attempting to serialise a WindowProxy exotic object which associated
browsing context has been discarded, we return an error instead.


andreastt:
r? @AutomatedTester 

Open questions:

* Should we also do this for web elements that are stale?
* Is _stale element reference_ the correct error?

AutomatedTester:
>    * Should we also do this for web elements that are stale?

We should really if we can see that it is a stale element reference

>    * Is stale element reference the correct error?

Yes.

PR #315, opened 2016-07-18
Make normative reference to HTTP

andreastt:
Resolves an action from the WebDriver F2F WG meeting in July 2016.

Define what it means to provide an HTTP frontend for the protocol.

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -361,6 +361,10 @@ var respecConfig = {
   </ul>
 
  <dt>HTTP and related specifications
+ <dd><p>To be <dfn>HTTP compliant</dfn>,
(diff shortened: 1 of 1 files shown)


andreastt:
r? @AutomatedTester 

I’m not confident this is the right or best way to do it. Perhaps it’s enough to reference RFC 7230 directly in the paragraph? However, I notice GET/POST/HEAD et al. are defined in RFC 7231. Other relevant bits could be defined in the other RFCs.

I like how the WHATWG spec can [make their own reference pointers](https://fetch.spec.whatwg.org/#references) to encompass more than one specification text.

AutomatedTester:
r+

PR #316, opened 2016-07-18
Fix SSL certificate handling

andreastt:
This PR resolves an action item from the WG F2F in July 2016.

It removes the chapter on SSL certificate handling and incorporates the logic directly with the _Go_ command.

Please do not squash the commits.


AutomatedTester:
landed as 51b9bbf cadf6b0 c94f390 377c2d6

PR #317, opened 2016-07-19
Return error on malformed URL

andreastt:
As decided in the WG’s F2F meeting in July 2016.

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -455,9 +455,11 @@ var respecConfig = {
  <dt>URLs
  <dd>The following terms are defined in the WHATWG URL standard: [[!URL]]
   <ul>
+   <!-- Absolute URL --> <li><dfn><a href=https://url.spec.whatwg.org/#syntax-url-absolute>Absolute URL</a></dfn>
+   <!-- Absolute URL with fragment --> <li><dfn><a href=https://url.spec.whatwg.org/#syntax-url-absolute-with-fragment>Absolute URL with fragment</a></dfn>
    <!-- Domain --><li><dfn><a href="https://url.spec.whatwg.org/#concept-domain">Domain</a></dfn>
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -2024,12 +2024,17 @@
...
+
  <li><p>Let <var>url</var> be the result of
-  <a>getting a property</a> named "<code>url</code>"
-  from the <var>parameters</var> argument.
+  <a>getting a property</a> named <code>url</code>
+   from the <var>parameters</var> argument.
 
- <li><p>If <var>url</var> is undefined,
-  return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
+ <li><p>If <var>url</var> is a malformed URL,
```
AutomatedTester: We should change this to error when it is not an [Absolute URL with fragement](https://url.spec.whatwg.org/#syntax-url-absolute-with-fragment).  At the moment there is no explanation of what a malformed URL is.

andreastt: Presumably it should also be allowed to navigate to a URL _without_ a fragment.

webdriver-spec.html
```diff
@@ -2049,6 +2049,13 @@
   <a>getting a property</a> named <code>url</code>
   from the <var>parameters</var> argument.
 
+ <li><p>If <var>url</var> is a malformed URL,
```
AutomatedTester: Yay github for losing context...

A fragment must be 0 or more Url units (see [url spec](https://url.spec.whatwg.org/#syntax-url-fragment). From the URL spec `A URL
must be either a relative URL with fragment or an absolute URL with fragment`. Since we only want the the absolute url we should describe it as such.

andreastt: I was on an aeroplane, and the wi-fi stopped working just as I was trying to push my fixup after rebasing. Should be fixed now.

andreastt:
Ping @AutomatedTester

AutomatedTester:
landed in 561fb5e and d455bf9

PR #319, opened 2016-07-22
Fix the value returned by Get Title.

jgraham:
This now refers directly to the IDL Attribute definition in HTML, which
covers all the cases which the previous text missed (zero or multiple
<title> elements, SVG documents, etc.).


andreastt:
https://www.w3.org/Bugs/Public/show_bug.cgi?id=29750

andreastt:
Ping @AutomatedTester

AutomatedTester:
landed in cefc45f

PR #320, opened 2016-08-02
Add algorithms for decoding an action sequence.

jgraham:


webdriver-spec.html
```diff
@@ -558,6 +566,13 @@
  which encapsulates the value returned,
  whereas an <a>error</a> response has an associated <a>error code</a>.
 
+<p>When calling a fallible algorithm, the construct
+  "Let <var>result</var> be the result of <dfn>trying</dfn> to
```
AutomatedTester: `trying`? what is the definition in this case?

PR #321, opened 2016-08-08
Share definition of in-view centre point

andreastt:
The Element Click and pointer-interactable concepts both need to calculate
an element's visible centre point.


AutomatedTester:
landed in 76852ec

PR #324, opened 2016-09-02
Use non-normative language in the Security appendix.

shs96c:
This address issue #24832

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -6880,24 +6880,21 @@ session := new_session(capabilities)</code></pre>
 <section class=appendix>
 <h2>Security</h2>
 
-<p>The following suggestions are made to implementors to increase the
- security of users of WebDriver:
+<p>A user agent can rely on a command-line flag or a configuration
+ option to test whether to enable WebDriver, or alternatively make the
+ user agent initiate or confirm the connection through a privileged
+ content document or control widget, in case the user agent does not
+ directly implement the HTTP endpoints.
 
- <ul>
-  <li>Provide a command-line flag or configuration option to enable
-   WebDriver.
-  <li>Only listen on loopback devices for requests for
-   new <a>sessions</a>.
-  <li>Provide a configuration option to limit the accepted IP range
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -6878,23 +6878,24 @@
...
+
+ <ul>
+  <li>Provide a command-line flag or configuration option to enable
+   WebDriver.
+  <li>Only listen on loopback devices for requests for
+   new <a>sessions</a>.
+  <li>Provide a configuration option to limit the accepted IP range
+   allowed to connect and make requests, such as the IPv4 localhost
+   CIDR range <code>127.0.0.0/8</code> and the IPv6 localhost
+   address <code>::1</code>. [[!RFC4632]]
```
AutomatedTester: We should include what the default is here.

uchuugaka: Seems like this should be default but there should also be the ability to designate an address and and to do so over https and or http2 with TLS. 
This is common practice with databases. 
Ultimately lots of folks will want to use or implement a grid of systems.

shs96c:
What "default" are we talking about here? This is in a non-normative part of the spec, and these are just suggestions for implementors to increase security.

webdriver-spec.html
```diff
@@ -6878,23 +6878,24 @@
 <section class=appendix>
 <h2>Security</h2>
 
-<p>A user agent may rely on a command-line flag
- or a configuration option to test whether to enable WebDriver.
- Another option is to make the user agent initiate or confirm
- the connection through a privileged content document or control widget,
- in case the user agent does not directly implement the HTTP endpoints.
```
andreastt: The latter part of this is left out in the changes.

andreastt:
Generally I don’t really see the benefit of making this into a list, as one could just change the word “must” in the last paragraph to “can” or something similar.

webdriver-spec.html
shs96c: Now modified to follow your previous suggestion about wording.

shs96c:
If there are no more comments, I'll merge this in.

PR #325, opened 2016-09-02
It is possible for end nodes to have multiple active sessions.

shs96c:
It is up to implementors of the spec to define their own
value for "maximum active sessions"

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/325)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -1761,8 +1761,6 @@ session := new_session(capabilities)</code></pre>
  <dfn>maximum active sessions</dfn> (an integer)
  that defines the number of <a>active sessions</a>
  that are supported.
- This may be “unlimited” for <a>intermediary nodes</a>,
- but must be exactly one for a <a>remote end</a> that is an <a>endpoint node</a>.
 
 <p>A <a>session</a> has an associated <dfn>session ID</dfn>
  (a <a>UUID</a>) used to uniquely identify this session.  Unless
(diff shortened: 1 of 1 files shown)


jgraham:
I think there is possibly some misunderstanding here. It is not intended to forbid multiple sessions with entirely different browser instances (e.g. with different profiles). It is intended to forbid multiple simultaneous connections to the same browser. I haven't heard any use case for allowing this so I think the restriction makes sense.

AutomatedTester:
Do you have a use case where someone would want to have mutliple sessions in a [remote end](http://w3c.github.io/webdriver/webdriver-spec.html#dfn-endpoint-node). It feels like people will just set this to 1 already so might as well put it in the spec.

shs96c:
Just because we've not thought of the use case yet doesn't seem like a strong reason to forbid it. I'd prefer to do something like suggesting that each end-node only accept a single session. For example, I could imagine someone starting a session, hitting a bug, and starting a new session to the same browser to investigate what's going on.

andreastt:
I don’t think that’s entirely unreasonable, but it opens up a wormhole of race condition aspects we need to consider throughout the spec.

Some examples:

* What happens when an _Execute Async Script_ command is running and a second session is requested? The way the spec is written right now is that the main loop is synchronous (that’s what makes the API blocking), so the _New Session_ request wouldn’t be processed until _Execute Async Script_ times out.
* If we solve the above problem and introduce separate threads for each session to operate independently, how will we deal with concurrent access to DOM or races in commands competing for focus?

shs96c:
I can reword to suggest that the value should be 1 for endpoint nodes as the spec makes no effort to deal with concurrency. Would that work?

andreastt:
But there’s nothing stopping an additional client to re-use a session from a different thread/location. Because we don’t describe the characteristics of the local end, it’s possible to imagine two clients with internal synchronisation reusing the same HTTP session endpoints, e.g. both calling to the same `/session/<the same session id here>/…`.

Because a WebDriver session is often synonymous with the lifetime of the browser process, I’m finding it hard to imagine how allowing multiple sessions would work in practice… What would happen when one of the session’s calls quit? Wouldn’t they also compete over the same window resources?

shs96c:
Dropping this. I don't think the additional complexity is worth the effort.

PR #326, opened 2016-09-02
Suggest browser specific extension in Protocol Extensions.

shs96c:
This addresses the concern raised in #25418


andreastt:
This seems like a good clarification to me. In Mozilla’s case we’ve already exposed `/session/{session id}/moz/xbl/*` as a sub-hierarchy for interacting with XBL attributes in anonymous content (the less you know about this the better!), but this need not be in conflict with the way this change is worded, as far as I understand it.

andreastt:
This all looks fine to me.

PR #327, opened 2016-09-06
Overhaul the section on processing capabilities.

jleyba:
There are several goals with these changes:

1.  The endpoint node should either be able to satisfy all capabilities
    requested by the local end, or fail to create a session.

2.  The local end should be able to submit multiple sets of acceptable
    capabilities with a clearly defined order of preference. This eliminates
    the need for extra, potentially expensive, new session requests when one
    set of capabilities cannot be satisfied.

3.  Permit the definition and processing of custom capabilities, either by
    the endpoint node or an intermediate node.

4.  Preserve wire compatibility with existing implementations (i.e. the changes
    do not depend on the "requiredCapabilities" and "desiredCapabilities"
    parameters).

----

When the local end truly has no preference for the features supported by a
new session, it can simply request an empty set of capabilities.

    POST /session
    {"capabilities": {}}

The local end requires a specific browser (Firefox), but has no preference
for the exact value:

    POST /session
    {"capabilities": {"alwaysMatch":{"browserName":"firefox"}}}

The local end requires Firefox on Linux, or Chrome on Windows: the order
specified dictates in which order the endpoint should try to satisfy the
request:

    POST /session
    {"capabilities": {
      "firstMatch": [
        {"browserName": "firefox", "platformName": "linux"},
        {"browserName": "chrome", "platformName": "windows"}
      ]}}

Capabilities that are always required may be specified as "alwaysMatch". These
are merged with each "firstMatch" entry, eliminating the need to specify
expensive capabilities (like an encoded profile) multiple times:

    POST /session
    {"capabilities": {
       "alwaysMatch": {
         "browserName": "firefox",
         "profile": "base64-encoded-blob"
       },
       "firstMatch": [
         {"platformName": "linux"},
         {"platformName": "windows"}
       ]}}

Version strings are complicated. The exact matching algorithms are explicitly
left for each vendor, however, implementors are encouraged to support basic
operators to define a range of versions. That is, to request a version of
Firefox older than 48, the local end would specify:

    POST /session
    {"capabilities": {
       "alwaysMatch": {
         "browserName": "firefox",
         "browserVersion": "< 48"
       }}}

If an intermediate node requires additional information (i.e. auth credentials),
it MAY define custom capabilities, but it SHOULD require this information be
passed as a top level parameter:

    POST /session
    {"username": "test-user",
     "password": "abc123",
     "capabilities": {}}

Capabilities may never be ignored. If the endpoint does not recognize a key, it
must fail to match on those capabilities:

    POST /session
    {"capabilities": {
       "alwaysMatch": {
         "browserName": "firefox",
         "unrecognized-capability": 1234
       }}}

Given that unrecognized capabilities will cause new session requests to fail,
the spec requires that intermediate nodes that use capabilities for routing
remove those capabilities before forwarding the request to the selected remote
end.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/327)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -2180,6 +2180,9 @@ entries:
          "<code>script</code>". Otherwise, set the
          <a>session script timeout</a> to 30,000 milliseconds.
       </ol>
+
+    <li><p>Apply changes to the user agent for any implementation defined
+      capabilities selected during the <a>processing capabilities</a> step.
   </ol>
 
  <li><p>Set the <a>webdriver-active flag</a> to true.</p>
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -1823,8 +1823,8 @@ entries:
   <dd>The platform version, as a string.
 
   <dt>"<code>acceptSslCerts</code>"
-  <dd>False, indicating the session <em>will not</em> implicitly trust untrusted
-  or self-signed SSL certificates on <a data-lt=go>navigation</a>.
+  <dd>Boolean false, indicating the session <em>will not</em> implicitly trust
+  untrusted or self-signed SSL certificates on <a data-lt=go>navigation</a>.
   </dl>
 
 <li><p>Let <var>capabilities length</var> be the number of entries in <var>capabilities</var>.
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -1792,8 +1792,11 @@ named "<code>firstMatch</code>" from <var>capabilities request</var>.
   <li><p>Let <var>entry</var> be the entry in <var>secondary</var> at index
   <var>k</var>.
 
-  <li><p>If the name of <var>entry</var> is <em>not</em> the name of an entry
-  in <var>result</var>, add <var>entry</var> to <var>result</var>.
+  <li><p>If the name of <var>entry</var> is also the name of an entry in
+  <var>result</var>,
+  return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
+
+  <li><p>Otherwise, add <var>entry</var> to <var>result</var>.
 
   <li><p>Increment <var>k</var> by 1.
   </ol>
@@ -1851,7 +1854,7 @@ entries:
       <dt>"<code>browserVersion</code>"
       <dd><p>Compare <var>capability value</var> to the
       "<code>browserVersion</code>" entry in <var>matched capabilities</var>
-      using an implementation defined comparison algorithm. The comparison
+      using an implementation-defined comparison algorithm. The comparison
       <em>SHOULD</em> accept a <var>capability value</var> that places
       constraints on the version using the "<code>&lt;</code>",
       "<code>&lt;=</code>", "<code>&gt;</code>", and "<code>&gt;=</code>"
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -1850,7 +1850,8 @@ entries:
 
     <dl class=switch>
       <dt>"<code>browserName</code>"
-      <dd><p>If <var>capability value</var> is <a>null</a>, <a>undefined</a>,
+      <dd>
+        <p>If <var>capability value</var> is <a>null</a>, <a>undefined</a>,
         or equal to the string "<code>*</code>", it <em>MUST</em> be considered
         a match for the entry in <var>matched capabilities</var>.
 
@@ -1859,30 +1860,45 @@ entries:
         return null.
 
       <dt>"<code>browserVersion</code>"
-      <dd><p>Compare <var>capability value</var> to the
-      "<code>browserVersion</code>" entry in <var>matched capabilities</var>
-      using an implementation-defined comparison algorithm. The comparison
-      <em>SHOULD</em> accept a <var>capability value</var> that places
-      constraints on the version using the "<code>&lt;</code>",
-      "<code>&lt;=</code>", "<code>&gt;</code>", and "<code>&gt;=</code>"
-      operators.
+      <dd>
+        <p>If <var>capability value</var> is <a>null</a>, <a>undefined</a>,
+        or equal to the string "<code>*</code>", it <em>MUST</em> be considered
+        a match for the entry in <var>matched capabilities</var>.
 
-      <p>If the two values do not match, return null.
+        <p>Otherwise, compare <var>capability value</var> to the
+        "<code>browserVersion</code>" entry in <var>matched capabilities</var>
+        using an implementation-defined comparison algorithm. The comparison
+        <em>SHOULD</em> accept a <var>capability value</var> that places
+        constraints on the version using the "<code>&lt;</code>",
+        "<code>&lt;=</code>", "<code>&gt;</code>", and "<code>&gt;=</code>"
+        operators.
 
-      <p class=note>Version comparison is left as an implementation detail since
-      each user agent will likely have conflicting methods of encoding the
-      user agent version and standardizing these schemes is beyond the scope of
-      this specification.
+        <p>If the two values do not match, return null.
+
+        <p class=note>Version comparison is left as an implementation detail
+        since each user agent will likely have conflicting methods of encoding
+        the user agent version and standardizing these schemes is beyond the
+        scope of this specification.
 
       <dt>"<code>platformName</code>"
-      <dd><p>If <var>capability value</var> is not equal to
-      "<code>platformName</code>" entry in <var>matched capabilities</var>,
-      return null.
+      <dd>
+        <p>If <var>capability value</var> is <a>null</a>, <a>undefined</a>,
+        or equal to the string "<code>*</code>", it <em>MUST</em> be considered
+        a match for the entry in <var>matched capabilities</var>.
+
+        <p>Otherwise, if <var>capability value</var> is not a string equal to
+        the "<code>platformName</code>" entry in
+        <var>matched capabilities</var>, return null.
 
       <dt>"<code>platformVersion</code>"
-      <dd><p>If <var>capability value</var> is not equal to
-      "<code>platformVersion</code>" entry in <var>matched capabilities</var>,
-      return null.
+      <dd>
+        <p>If <var>capability value</var> is <a>null</a>, <a>undefined</a>,
+        or equal to the string "<code>*</code>", it <em>MUST</em> be considered
+        a match for the entry in <var>matched capabilities</var>.
+
+        <p>Otherwise, if <var>capability value</var> is not a string equal to
+        the "<code>platformVersion</code>" entry in
+        <var>matched capabilities</var>, return null.
 
       <dt>"<code>acceptSslCerts</code>"
       <dd>
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -2067,20 +2067,20 @@ entries:
  a <a>session not created</a> <a>error</a> is returned.
 
 <p>If the <a>remote end</a> is an <a>intermediary node</a>,
- it may use the result of the <a>processing capabilities</a> algorithm to route
- the new session request to the appropriate <a>endpoint node</a>.
- An <a>intermediary node</a> may also define custom <a>capabilities</a> to
- assist in this process, however, these capabilities  <em>MUST NOT</em> be
+ it <em>MAY</em> use the result of the <a>processing capabilities</a> algorithm
+ to route the new session request to the appropriate <a>endpoint node</a>.
+ An <a>intermediary node</a> <em>MAY</em> also define custom <a>capabilities</a>
+ to assist in this process, however, these capabilities  <em>MUST NOT</em> be
  forwarded to the <a>endpoint node</a>. Furthermore, if the
  <a>intermediary node</a> requires additional information unrelated to user
  agent features, it is <em>RECOMMENDED</em> that this information be passed as
  top-level parameters and not part of the requested <a>capabilities</a>.
 
 <aside class=example>
- <p>An <a>intermediary node</a> may require authentication on all new session
- requests. This authentication is an argument to the new session command itself
- and not the user agent's capabilities. Therefore, the authentication should be
- passed as a top-level parameter and not embedded in capabilities:
+ <p>An <a>intermediary node</a> <em>MAY</em> require authentication on all new
+ session requests. This authentication is an argument to the new session command
+ itself and not the user agent's capabilities. Therefore, the authentication
+ should be passed as a top-level parameter and not embedded in capabilities:
 
  <pre class=highlight>{
     "user": "alice",
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -1435,288 +1435,513 @@ var respecConfig = {
 <section>
 <h2>Capabilities</h2>
 
-<p>WebDriver <dfn>capabilities</dfn> allow the <a>local end</a>
- to specify what features it desires or requires the <a>remote end</a> to fulfill
- to be able to create a <a>new session</a>.
+<p>WebDriver <dfn>capabilities</dfn> are used to communicate the features
+supported by a given implementation. The <a>local end</a> may use capabilities
+to define which features it requires the <a>remote end</a> to satisfy when
+creating a <a>new session</a>. Likewise, the <a>remote end</a> uses capabilities
+to describe the full feature set for a session.
 
-<p>The following table provides an overview
- of recognised capabilities, and their usage and effect:
+<p>The following table provides an overview of the standard capabilities that
+each implementation <em class="rfc2119" title="MUST">MUST</em> support.
+An implementation <em class="rfc2119" title="MAY">MAY</em> define additional
+capabilities, although it is
+<em class="rfc2119" title="RECOMMENDED">RECOMMENDED</em> that custom
+capabilities prepend the capability key with a vendor prefix, as outlined in
+[[CSS21]].
+
+<aside class="example">
+<p>As an example, Mozilla could elect to hide new features behind capabilities
+with a "-moz-" prefix:
+
+<pre class=highlight>{
+  "browserName": "firefox",
+  "browserVersion": "1234",
+  "-moz-experimental-webdriver": true
+}</pre>
+</aside>
 
 <table class=simple>
- <tr>
+<thead>
+  <tr>
   <th>Capability
   <th>Key
   <th>Value Type
   <th>Description
- </tr>
+  </tr>
+</thead>
 
- <tr>
+<tbody>
+  <tr>
   <td><dfn>Browser name</dfn>
   <td>"<code>browserName</code>"
   <td>string
   <td>Identifies the user agent.
- </tr>
+  </tr>
 
- <tr>
+  <tr>
   <td><dfn>Browser version</dfn>
   <td>"<code>browserVersion</code>"
   <td>string
   <td>Identifies the version of the user agent.
- </tr>
+  </tr>
 
- <tr>
+  <tr>
   <td><dfn>Platform name</dfn>
   <td>"<code>platformName</code>"
   <td>string
   <td>Identifies the system or operating system
-   of the <a>endpoint node</a>.
- </tr>
+  of the <a>endpoint node</a>.
+  </tr>
 
- <tr>
+  <tr>
   <td><dfn>Platform version</dfn>
   <td>"<code>platformVersion</code>"
   <td>string
   <td>Identifies the system version of the <a>endpoint node</a>.
- </tr>
+  </tr>
 
- <tr>
+  <tr>
   <td><dfn>Accept SSL certificates</dfn>
   <td>"<code>acceptSslCerts</code>"
   <td>boolean
-  <td>Indicates that untrusted and self-signed SSL certificates
-   are trusted implicitly on <a data-lt=go>navigation</a>
-   for the duration of the session.
- </tr>
+  <td>Indicates whether untrusted and self-signed SSL certificates are
+  implicitly trusted on <a data-lt=go>navigation</a> for the duration of the
+  session.
+  </tr>
+
+  <tr>
+  <td><dfn>Page Load Strategy</dfn>
+  <td>"<code>pageLoadStrategy</code>"
+  <td>string
+  <td>Defines the <a>current session</a>’s <a>page load strategy</a>.
+  </tr>
+
+  <tr>
+  <td><dfn>Proxy Configuration</dfn>
+  <td>"<code>proxy</code>"
+  <td>JSON Object
+  <td>Describes a session's proxy configuration.
+  </tr>
+
+  <tr>
+  <td><dfn>Session Timeouts Configuration</dfn>
+  <td>"<code>timeouts</code>"
+  <td>JSON Object
+  <td>Describes the timeouts imposed on certain session operation.
+  </tr>
+</tbody>
 </table>
 
-<p>Implementations are free to use the <a>capabilities</a> dictionary
- passed in the request to create a <a>new session</a>
- to provide additional feature requests
- from any of the <a>remote ends</a>, including the <a>endpoint node</a>.
- Likewise the <a>remote end</a> may supply additional entries
- on the <a>capabilities</a> object returned
- from creating a <a>new session</a>.
+<section>
+<h3>Proxy Configurations</h3>
+<p>The <a>Proxy Configuration</a> capability is a JSON Object nested within the
+primary <a>capabilities</a>. Implementations
+<em class="rfc2119" title="MAY">MAY</em> define additional proxy configuration
+options, but they <em class="rfc2119" title="MUST NOT">MUST NOT</em> alter the
+semantics of those listed below.
 
-<aside class=example>
- <p>Capabilities may for example be used
-  to provide a serialised profile for a <a>remote end</a> to use
-  during startup of the user agent binary.
+<table class=simple>
+<thead>
+<tr>
+<th>Key
+<th>Value Type
+<th>Description
+</tr>
+</thead>
 
- <p>A fictive example of this:
+<tbody>
+<tr>
+<td>"<code>proxyType</code>"
+<td>string
+<td>Indicates the type of proxy configuration. This value
+<em class="rfc2119" title="SHOULD">SHOULD</em> be one of the following:
+"<code>pac</code>", "<code>noproxy</code>", "<code>autodetect</code>",
+"<code>system</code>", or "<code>manual</code>".
+</tr>
 
- <pre><code>capabilities := {}
-profile := serialise_profile("/path/to/profile")
-capabilities["profile"] = profile
-session := new_session(capabilities)</code></pre>
-</aside>
+<tr>
+<td>"<code>proxyAutoconfigUrl</code>"
+<td>string
+<td>Defines the URL for a proxy auto-config file. [URL]
+  This property <em class="rfc2119" title="SHOULD">SHOULD</em> only be set when
+  the "<code>proxyType</code>" is equal to "<code>pac</code>".
+</tr>
 
-<p>When <dfn>processing capabilities</dfn> with argument <var>parameters</var>
- a <a>remote end</a> must run the following steps:
+<tr>
+<td>"<code>ftpProxy</code>"
+<td>string
+<td>Defines the proxy <em>hostname</em> for FTP traffic. Alternatively, an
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -1435,288 +1435,513 @@
...
-<p>The following table provides an overview
- of recognised capabilities, and their usage and effect:
+<p>WebDriver <dfn>capabilities</dfn> are used to communicate the features
+supported by a given implementation. The <a>local end</a> may use capabilities
+to define which features it requires the <a>remote end</a> to satisfy when
+creating a <a>new session</a>. Likewise, the <a>remote end</a> uses capabilities
+to describe the full feature set for a session.
+
+<p>The following table provides an overview of the standard capabilities that
+each implementation <em class="rfc2119" title="MUST">MUST</em> support.
```
shs96c: No need for the class or title.

webdriver-spec.html
```diff
@@ -1435,288 +1435,513 @@
...
+  <td><dfn>Page Load Strategy</dfn>
+  <td>"<code>pageLoadStrategy</code>"
+  <td>string
+  <td>Defines the <a>current session</a>’s <a>page load strategy</a>.
+  </tr>
 
- <p>A fictive example of this:
+  <tr>
+  <td><dfn>Proxy Configuration</dfn>
+  <td>"<code>proxy</code>"
```
shs96c: We should probably thoroughly describe the proxy object, but that could wait for a follow edit.

webdriver-spec.html
```diff
@@ -1435,288 +1435,513 @@
...
+  <td>JSON Object
+  <td>Describes the timeouts imposed on certain session operation.
+  </tr>
+</tbody>
+</table>
 
-<p>When <dfn>processing capabilities</dfn> with argument <var>parameters</var>
- a <a>remote end</a> must run the following steps:
+<section>
+<h3>Proxy Configurations</h3>
```
shs96c: I see you've done the description already. :)

webdriver-spec.html
```diff
@@ -1435,288 +1435,513 @@
...
-    and the values do not match do the following:
+  <li><p>If <var>dc</var> is not a JSON Object,
+  return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
 
-    <ol>
-     <li><p>Append a string containing the property name
-      and the differences between the values.
-    </ol>
+  <li><p>Let <var>merged capabilities</var> be the result of
+  <a>merging capabilities</a> with <var>required capabilities</var> and
```
shs96c: Which takes precedence: the capabilities from "required" or "desired" capabilities. I'd suggest those from "required", but returning an error when keys clash would also be valid.

shs96c:
I need to finish the review, but broadly in agreement with this change. I think it addresses concerns raised on IRC within the #selenium project, at face to face meetings, and in personal discussions.

webdriver-spec.html
```diff
@@ -1435,288 +1435,513 @@
...
 
-     <li><p>Let <var>socks proxy</var> be the result of <a>getting a property</a>
-      named <code>socksProxy</code> from <var>proxy</var>.
+  <dt>"<code>platformVersion</code>"
+  <dd>The platform version, as a string.
 
-     <li><p>Let <var>socks proxy port</var> be the result of <a>getting a property</a>
-      named <code>socksProxyPort</code> from <var>proxy</var>.
+  <dt>"<code>acceptSslCerts</code>"
+  <dd>False, indicating the session <em>will not</em> implicitly trust untrusted
```
shs96c: "Boolean false" to differentiate from the string "false"

webdriver-spec.html
```diff
@@ -1435,288 +1435,513 @@
...
+  <var>k</var> in <var>capabilities</var>.
 
-   <dt>Otherwise
-   <dd><p>Return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
-  </dl>
+  <li><p>Switch on <var>capability name</var>:
+
+    <dl class=switch>
+      <dt>"<code>browserName</code>"
+      <dd><p>If <var>capability value</var> is not equal to
```
shs96c: Or if the capability value has not been set.

webdriver-spec.html
```diff
@@ -1435,288 +1435,513 @@
...
+      <p>If the two values do not match, return null.
+
+      <p class=note>Version comparison is left as an implementation detail since
+      each user agent will likely have conflicting methods of encoding the
+      user agent version and standardizing these schemes is beyond the scope of
+      this specification.
+
+      <dt>"<code>platformName</code>"
+      <dd><p>If <var>capability value</var> is not equal to
+      "<code>platformName</code>" entry in <var>matched capabilities</var>,
```
shs96c: The empty platform capability value matches all platforms.

webdriver-spec.html
```diff
@@ -1435,288 +1435,513 @@
...
+      return null.
+
+      <dt>"<code>platformVersion</code>"
+      <dd><p>If <var>capability value</var> is not equal to
```
shs96c: If the capability value is empty, there's an implicit match

webdriver-spec.html
```diff
@@ -1435,288 +1435,513 @@
...
+      "<code>platformVersion</code>" entry in <var>matched capabilities</var>,
```
shs96c: Do we want to support ranges? If so, we can handle in a follow up change.

webdriver-spec.html
```diff
@@ -1435,288 +1435,513 @@
...
+
+        <p>Let <var>processed value</var> be the result of implementation
+        specific steps to match on <var>capability name</var> with
+        <var>capability value</var>.
+
+        <p>If <var>processed value</var> is an error,
+        return <var>processed value</var>.
+
+        <p>Otherwise, create a new entry in <var>matched capabilities</var> with
+        name <var>capability name</var> and value <var>processed value</var>.
```
shs96c: I'm guessing that for things like "proxy" and "profile", `processed value` might be something like `true`, to indicate that the configured value was used?

webdriver-spec.html
```diff
@@ -1844,19 +2069,34 @@
...
 
 <p>If the <a>remote end</a> is an <a>intermediary node</a>,
- they are allowed to freely use the <a>capabilities</a> data
- e.g. to select a specific browser to test based on
- a combination of the <a data-lt="required-capabilities">required</a>
- and <a>desired capabilities</a>.
- Typically the <a>New Session</a> response from the <a>endpoint node</a>
- selected in this process will then be relayed directly to the <a>local end</a>,
- as outlined in step 1 of the <a>remote end steps</a>.
+ it may use the result of the <a>processing capabilities</a> algorithm to route
```
shs96c: s/may/MAY/

webdriver-spec.html
```diff
@@ -1844,19 +2069,34 @@
...
+ the new session request to the appropriate <a>endpoint node</a>.
+ An <a>intermediary node</a> may also define custom <a>capabilities</a> to
+ assist in this process, however, these capabilities
+ <em class="rfc2119" title="MUST NOT">MUST NOT</em> be forwarded to the
+ <a>endpoint node</a>. Furthermore, if the <a>intermediary node</a> requires
+ additional information unrelated to user agent features,
+ it is <em class="rfc2119" title="RECOMMENDED">RECOMMENDED</em> that this
+ information be passed as top-level parameters and not part of the requested
+ <a>capabilities</a>.
```
shs96c: Worth noting that endpoint nodes MUST ignore top-level parameters that they don't process.

webdriver-spec.html
```diff
@@ -1435,288 +1435,513 @@
...
+      <dd>
+        <p>If <var>capability name</var> is not recognized by the <a>endpoint node</a>,
+        return null.
+
+        <p>Let <var>processed value</var> be the result of implementation
+        specific steps to match on <var>capability name</var> with
+        <var>capability value</var>.
+
+        <p>If <var>processed value</var> is an error,
+        return <var>processed value</var>.
```
shs96c: You'd need to return "session not created", but have the value explain the error.

webdriver-spec.html
jleyba: This is already covered in the "merging capabilities" steps (required silently takes precedence over desired)

webdriver-spec.html
jleyba: Why don't we remove platformName and platformVersion from the standard set of capabilities?

1.  This frees us up from having to define canonical platform names
2.  This frees us from defining a version matching scheme (note I punted on that for browsers since versions are vendor specific)
3.  The local end should only specify the platform as routing information for intermediate nodes. In this case, the intermediate node can define its own mappings for platform name and version.
4.  In the absence of an intermediate node, the local end should already know what platform the remote end is on.

Having the endpoint node record the platform name and version is nice for debugging purposes, but coming up with a consistent naming scheme seems out of scope for this spec (especially since `navigator.userAgent` strings don't even have an established format)

webdriver-spec.html
jleyba: This step is within an iteration over the set capabilities, so capability value _will_ be set. It _may_ be set to `null`  (JSON does not support `undefined`), but that is implicitly handled by the "not equal to" comparison.

webdriver-spec.html
jleyba: > The empty platform capability value matches all platforms.

As with above, in this context, _capability value_ will always be set. If the requested capabilities do not specify a value for `platformName`, the capability is not considered when checking for a match (e.g. unset implies match all)

webdriver-spec.html
jleyba: > If the capability value is empty, there's an implicit match

_capability value_ is always set in this context

webdriver-spec.html
jleyba: On `browserVersion`? Sure.

On `platformVersion`? See my comments above about whether this should be included for endpoint nodes at all.

webdriver-spec.html
jleyba: I was trying to keep this two pass:

1.  Match and select session capabilities. That is what is described here.
2.  Initialize the session with matched capabilities. This is covered below.

The idea is matching and pre-checks are cheaper than trying to fully initialize a session. In other words, don't bother unpacking the profile in step (1) if there is another requested capability that the endpoint cannot satisfy.

webdriver-spec.html
jleyba: Why? If the error is "invalid argument" that is sufficient to return to a local end. The new session command failed because of an invalid argument error.

webdriver-spec.html
shs96c: The value is set, but null matches everything. This isn't spelt out.

jleyba: > The value is set, but null matches everything. This isn't spelt out.

Really? You think `{"browserName": null}` should match anything? Match anything should just be unset. If you set the value to a garbage value, it shouldn't match on anything. Garbage in, garbage out.

webdriver-spec.html
```diff
@@ -1435,288 +1435,506 @@
...
+<td>Defines the proxy <em>port</em> for FTP traffic.
+  This property <em>SHOULD</em> only be set when the "<code>proxyType</code>" is
+  equal to "<code>manual</code>".
+</tr>
+
+<tr>
+<td>"<code>httpProxy</code>"
+<td>string
+<td>Defines the proxy <em>hostname</em> for HTTP traffic. Alternatively, an
+  implementation <em>MAY</em> accept a hostname and port through this property.
```
juangj: This is nice for backwards compatibility. If a hostname:port pair is given here, AND httpProxyPort is set, which takes precedence?

webdriver-spec.html
```diff
@@ -1844,19 +2062,32 @@
...
- selected in this process will then be relayed directly to the <a>local end</a>,
- as outlined in step 1 of the <a>remote end steps</a>.
+ it <em>MAY</em> use the result of the <a>processing capabilities</a> algorithm
+ to route the new session request to the appropriate <a>endpoint node</a>.
+ An <a>intermediary node</a> <em>MAY</em> also define custom <a>capabilities</a>
+ to assist in this process, however, these capabilities  <em>MUST NOT</em> be
+ forwarded to the <a>endpoint node</a>. Furthermore, if the
+ <a>intermediary node</a> requires additional information unrelated to user
+ agent features, it is <em>RECOMMENDED</em> that this information be passed as
+ top-level parameters and not part of the requested <a>capabilities</a>.
```
juangj: Just to confirm/be clear: The intermediary node is permitted to leave that top-level information in the request, correct? Endpoint nodes are supposed to only examine the capabilities object.

webdriver-spec.html
```diff
@@ -1435,288 +1435,506 @@
...
 
   <ol>
-   <li><p>Let <var>capability</var> be the entry
-    in <var>required capabilities</var> at index <var>k</var>.
-
-   <li><p>If the name of the <var>capability</var> is among
-    the names of entries in <var>desired capabilities</var>,
-    remove the corresponding entry from <var>desired capabilities</var>.
+  <li><p>If <var>desired capabilities</var> is <a>null</a> or <a>undefined</a>,
+  set the value to an empty JSON List.
```
juangj: This should set the value to `[{}]`, rather than `[]`, so that there is at least one candidate capability set to consider.

webdriver-spec.html
```diff
@@ -1435,288 +1435,506 @@
...
-    If this can not be set during this process return <a>error</a>.
+  <ol>
+  <li><p>Let <var>entry</var> be the entry in <var>secondary</var> at index
+  <var>k</var>.
 
-   <dt>"<code>autodetect</code>"
-   <dd><p>Set the proxy to Auto-Detect proxy setting from the network
-    using implementation defined steps.
-    If this can not be set during this process return <a>error</a>.
+  <li><p>If the name of <var>entry</var> is <em>not</em> the name of an entry
```
juangj: If the values of the entries in *primary* and *secondary* are both JSON objects, should they be merged recursively?

For example, in a situation like this it might be nice to be able to specify additional values to add to the `firefoxOptions` without having to re-specify the entire `firefoxOptions` object.
```
{
  "alwaysMatch": {
    "browserName": "Firefox",
    "firefoxOptions": {
      "pref": { "some.pref", true" },
      "profile": "enormousstring=="
    }
  },
  "firstMatch": [
    {
      "firefoxOptions": {
        "prefs": { "some.other.pref", true }
      }
    }
  ]
}
```

It might be useful to merge those

webdriver-spec.html
shs96c: I'd suggest raising an error, explaining the clash. Let the user make up their mind rather than guessing.

webdriver-spec.html
shs96c: Not sure if that's the best idea. I can imagine two cases where people overwrite the "required" capability:

1. They're not aware it's been set, and so don't know what they're asking for.
2. They want to override a particular setting if this desired capability matches (eg. "always use this firefox profile, unless you pick version 50, in which case, use this one instead")

The first suggests that an error be thrown. The second suggests that desired takes precedence.

webdriver-spec.html
shs96c: This would cause havoc with the `proxy` setting.

webdriver-spec.html
shs96c: Yes, I think a `null` `browserName` should match everything (`"*"` too). Languages like Java don't have a separate "undefined" and "null" value, and a simple implementation of an intermediary node would just call `get` on deserialised JSON object to try and figure out which browser to choose.

```
Object value = capabilities.get("browserName");
Browser browser = value == null ? Browser.CHROME : Browser.from(value);
```

webdriver-spec.html
shs96c: We should canonicalise the common platform names (Windows, Linux, and OS X), otherwise there's less hope of interop between competing services implementing this spec. That is CompanyX could define a linux node as "linux", CompanyY as "gnu/linux", and then someone wishing to switch services would need to recode their tests. We should also indicate the name of the property used for determining the desired platform.

shs96c: See the reasoning for `browserName` as to why I think the empty platform capability should match

webdriver-spec.html
shs96c: But it may be `null`. As above :)

webdriver-spec.html
shs96c: We should discuss this in the face to face in Lisbon. For now, we can leave this as it's defined here.

webdriver-spec.html
shs96c: Partly because the effect is that the session has not been created, thus making this a more descriptive error to return to the user, partly because existing local end code is written to expect "session not created" to be converted to something like `NoSuchSessionException`, and this is what frameworks catch, and partly because it makes implementing the protocol handshake between OSS and W3C easier.

But mainly because it's more descriptive to the user. The error description should indicate that there was a value that didn't make sense in this particular field.

webdriver-spec.html
shs96c: Gotcha. 👍

webdriver-spec.html
shs96c: Yes! Endpoint nodes should only be looking at parameters and fields that they care about.

webdriver-spec.html
jleyba: Ok, I've updated the text to return an error in the presence of duplicate keys.

webdriver-spec.html
jleyba: Done.

webdriver-spec.html
jleyba: Updated so browserName, browserVersion, platformName, and platformVersion all support wildcard matching.

webdriver-spec.html
jleyba: Leaving the language in this section as is, but I've updated the new session command to indicate it should always return a "session not created" error that wraps w/e this routine failed with

webdriver-spec.html
jleyba: Updated text to clarify that this is permitted

webdriver-spec.html
jleyba: Done

webdriver-spec.html
jleyba: Matching fails on key conflicts at the top level. Recursion adds too much complexity.

webdriver-spec.html
```diff
@@ -1435,288 +1435,532 @@
...
+supported by a given implementation. The <a>local end</a> may use capabilities
+to define which features it requires the <a>remote end</a> to satisfy when
+creating a <a>new session</a>. Likewise, the <a>remote end</a> uses capabilities
+to describe the full feature set for a session.
+
+<p>The following table provides an overview of the standard capabilities that
+each implementation <em>MUST</em> support. An implementation <em>MAY</em> define
+additional capabilities, although it is <em>RECOMMENDED</em> that custom
+capabilities prepend the capability key with a vendor prefix, as outlined in
+[[CSS21]].
```
shs96c: We have a section on protocol extension in the spec. Refer to that instead?

webdriver-spec.html
```diff
@@ -1435,288 +1435,532 @@
...
-<p>Implementations are free to use the <a>capabilities</a> dictionary
- passed in the request to create a <a>new session</a>
- to provide additional feature requests
- from any of the <a>remote ends</a>, including the <a>endpoint node</a>.
- Likewise the <a>remote end</a> may supply additional entries
- on the <a>capabilities</a> object returned
- from creating a <a>new session</a>.
+  <td>Indicates whether untrusted and self-signed SSL certificates are
+  implicitly trusted on <a data-lt=go>navigation</a> for the duration of the
+  session.
```
shs96c: Worth pointing out that this SHOULD default to `true` here.

webdriver-spec.html
jleyba: The section on protocol extension talks about vendor prefixs used in the URI path. There's no mention of extending parameters. Should that be added?

webdriver-spec.html
```diff
@@ -1435,288 +1435,532 @@
...
- <pre><code>capabilities := {}
-profile := serialise_profile("/path/to/profile")
-capabilities["profile"] = profile
-session := new_session(capabilities)</code></pre>
-</aside>
+  <tr>
+  <td><dfn>Session Timeouts Configuration</dfn>
+  <td>"<code>timeouts</code>"
+  <td>JSON Object
+  <td>Describes the timeouts imposed on certain session operation.
```
shs96c: Refer to the `Set Timeout` section.

webdriver-spec.html
jleyba: Eww no. Users should explicitly expose themselves to the risks that come with insecure certs

webdriver-spec.html
```diff
@@ -1435,288 +1435,532 @@
...
 
-  <dl>
-   <dt>"<code>browserName</code>"
-   <dd>Lowercase name of the user agent
+<section>
+<h3>Session Timeouts Configuration</h3>
+<p>The <a>Session Timeouts Configuration</a> capability is a JSON Object nested
+within the primary capabilities. It consists of the properties listed in the
+table below. These values may be changed after creating a <a>new session</a>
+using the <a>Set Timeout</a> command.
```
shs96c: These timeouts are all in the `Set Timeout` section. Refer to these rather than duplicate the information?

webdriver-spec.html
```diff
@@ -1435,288 +1435,532 @@
...
-  that contains all entries from <var>required capabilities</var>
-  and all entries from <var>desired capabilities</var>.
+<li><p>Let <var>length</var> be the length of <var>desired capabilities</var>.
 
- <li><p>Let <var>j</var> be 0.
+<li><p>Let <var>k</var> be 0.
 
- <li><p>Let <var>capabilities length</var> be
-  the length of <var>unprocessed capabilities</var>.
+<li><p>Let <var>matched capabilities</var> be <a>null</a>.
```
shs96c: The way this is written, if both `firstMatch` and `alwaysMatch` are empty, then `matched capabilities` will by `null` when exiting the loop, meaning an weird value will be returned. ATM, selenium will return a default driver implementation (probably firefox) if this is the case. I'd suggest using a default value of an empty JSON Object instead of `null` in this section.

webdriver-spec.html
```diff
@@ -1435,288 +1435,532 @@
...
+  <dd>Lowercase name of the current platform.
 
-     <li><p>Let <var>socks proxy port</var> be the result of <a>getting a property</a>
-      named <code>socksProxyPort</code> from <var>proxy</var>.
+  <dt>"<code>platformVersion</code>"
+  <dd>The platform version, as a string.
 
-     <li><p>If <var>socks proxy</var> and <var>socks proxy port</var> is defined:
+  <dt>"<code>acceptSslCerts</code>"
+  <dd>Boolean false, indicating the session <em>will not</em> implicitly trust
```
shs96c: Boolean true, since the largest audience for this spec are QA engineers who may not be able to control their environment enough to be able to have a valid SSL cert, and who also are most likely to be inexperienced enough developers to not be able to set this parameter. *sigh*

webdriver-spec.html
```diff
@@ -1435,288 +1435,532 @@
...
+
+      <dt>"<code>platformName</code>"
+      <dd>
+        <p>If <var>capability value</var> is <a>null</a>, <a>undefined</a>,
+        or equal to the string "<code>*</code>", it <em>MUST</em> be considered
+        a match for the entry in <var>matched capabilities</var>.
+
+        <p>Otherwise, if <var>capability value</var> is not a string equal to
+        the "<code>platformName</code>" entry in
+        <var>matched capabilities</var>, return null.
```
shs96c: We need to add a TODO to specify the common platform names.

webdriver-spec.html
AutomatedTester: When discussing self-signed certificates at the last [F2F](https://www.w3.org/2016/07/14-webdriver-minutes.html#item02) there was talk from vendors that they/their infosec teams would not be too happy. We did discuss a mechanism for white listing domains for self signing, can't see it in the minutes, and @andreastt is looking to implement it in Marionette (we can hide how it works in geckodriver if need be). I think this is a good balance since @seleniumhq can overwrite it if it wants vendors still keep users safe

webdriver-spec.html
shs96c: LGTM

webdriver-spec.html
```diff
@@ -1844,19 +2088,34 @@
...
- selected in this process will then be relayed directly to the <a>local end</a>,
- as outlined in step 1 of the <a>remote end steps</a>.
+ it <em>MAY</em> use the result of the <a>processing capabilities</a> algorithm
+ to route the new session request to the appropriate <a>endpoint node</a>.
+ An <a>intermediary node</a> <em>MAY</em> also define custom <a>capabilities</a>
+ to assist in this process, however, these capabilities  <em>MUST NOT</em> be
+ forwarded to the <a>endpoint node</a>. Furthermore, if the
+ <a>intermediary node</a> requires additional information unrelated to user
+ agent features, it is <em>RECOMMENDED</em> that this information be passed as
+ top-level parameters and not part of the requested <a>capabilities</a>. An
```
shs96c: If they are passed this way, the parameters MAY be passed to the endpoint node.

webdriver-spec.html
```diff
@@ -1905,20 +2165,62 @@
...
+    <li><p>If <var>proxy</var> is a <a>Proxy Configuration</a> object, take
+      implementation-defined steps to set the user agent proxy using the
+      extracted <var>proxy</var> configuration. If the defined proxy cannot be
+      configured,
+      return <a>error</a> with <a>error code</a> <a>session not created</a>.
+
+    <li><p>Let <var>timeouts</var> be the result of getting property
+      "<code>timeouts</code>" from <var>capabilities</var>.
+
+    <li><p>If <var>timeouts</var> is a <a>Session Timeouts Configuration</a>
```
shs96c: And if it's not?

webdriver-spec.html
```diff
@@ -1435,288 +1435,532 @@
...
-    append a string saying that a JSON Object is required,
-    else call <a>set the proxy</a> passing in <var>proxy</var>.
-
-   <li><p>Let <var>page load strategy</var> be the result of <a>getting a property</a>
-    named <code>pageLoadStrategy</code> from <var>unprocessed capability</var>.
-    If <var>page load strategy</var> is <a>undefined</a>,
-    then set the entry <code>pageLoadStrategy</code>
-    in <var>server capabilities</var> to <a>normal</a>.
-  </ol>
+  <li><p>If <var>merged capabilities</var> is an <a>error</a>,
```
AutomatedTester: This can be merged with the point above to make

```
<li><p>Let <var>merged capabilities</var> be the result of
 <a>merging capabilities</a> with <var>required capabilities</var> and
<var>dc</var> as arguments. 
 <p>Otherwise, return <var>merged capabilties</var>
```

webdriver-spec.html
```diff
@@ -1435,288 +1435,532 @@
...
+  <li><p>Let <var>entry</var> be the entry in <var>secondary</var> at index
+  <var>k</var>.
 
-   <dt>"<code>autodetect</code>"
-   <dd><p>Set the proxy to Auto-Detect proxy setting from the network
-    using implementation defined steps.
-    If this can not be set during this process return <a>error</a>.
+  <li><p>If the name of <var>entry</var> is also the name of an entry in
+  <var>result</var>,
+  return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
```
AutomatedTester: should we process all the entries and keep track of errors and if `errors.length > 0` then raise the errors?

Doing that might catch multiple issues rather than people fix one thing retry, fix the next thing, retry, fix the next thing and so on.

webdriver-spec.html
shs96c: In Selenium, we'll just set the capability to be as user friendly as possible, and require people who care to change it to `false` if they want to.

andreastt:
@jleyba At the F2F we decided to adopt this proposal, congratulations!

There were noted several technical problems with the algorithm in the patch, but it was suggested that we would go ahead and merge it and fix these later. There is general agreement on the approach.

shs96c:
Adding @AutomatedTester as an assignee, since the action item from the Lisbon F2F was for him to make some edits and then land this patch.

AutomatedTester:
landed, with some updates (and I expect there to be a lot more) in d0ef240f535d00413403d4006f2b6dd7a3298eac

PR #334, opened 2016-09-21
Allow status command to include optional arbitrary JSON fields

andreastt:
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/334)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -90,39 +90,6 @@ var respecConfig = {
  or specific steps may be implemented in any manner,
  so long as the end result is equivalent.
 
-<p>The standard defines four broad conformance classes:
-
-<dl>
- <dt><dfn data-lt="local end|local ends|local">Local End</dfn>
- <dd><p>The local end represents the client side of the protocol,
-  which is usually in the form of language-specific libraries
-  providing an API on top of the WebDriver <a href=#protocol>protocol</a>.
-  This specification does not place any restrictions on the details of those libraries
-  above the level of the wire protocol.
-
-  <p class=issue>Define the requirements on the local end somewhere.
-
- <dt><dfn data-lt="remote end|remote ends|remote">Remote End</dfn>
- <dd>The remote end hosts the server side of the <a href=#protocol>protocol</a>.
-  Defining the behaviour of a remote end in response to the WebDriver protocol
-  forms the largest part of this specification.
-
- <dt><dfn data-lt="intermediary node|intermediary nodes">Intermediary Node</dfn>
- <dd>Intermediary nodes are those that act as proxies,
-  implementing both the client and server sides of the <a href=#protocol>protocol</a>.
-  Intermediary nodes must be black-box indistinguishable from a <a>remote end</a>
-  from the point of view of <a>local end</a>,
-  and so are bound by the requirements on a <a>remote end</a> in terms of the wire protocol.
-  However they are not expected to implement <a>remote end steps</a> directly.
-
- <dt><dfn>Endpoint Node</dfn>
- <dd>An endpoint node is the final <a>remote end</a>
-  in a chain of nodes that is not an <a>intermediary node</a>.
-  The endpoint node is implemented by a user agent or a similar program.
-  An endpoint node must be, like <a>intermediary nodes</a>,
-  indistinguishable from a <a>remote end</a>.
-</dl>
-
 <section>
 <h3>Dependencies</h3>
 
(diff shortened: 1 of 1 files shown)


jgraham:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[webdriver-spec.html, line 1995 at r1](https://reviewable.io:443/reviews/w3c/webdriver/334#-KSfXb3YjtdqrAH3mss_:-KSfXb3ZovcL0MF-l9y0:b4oc4kx) ([raw file](https://github.com/w3c/webdriver/blob/958b4711a7313d1c91c3b6e87e137c5b0b5cd6c7/webdriver-spec.html#L1995)):*
> ```HTML
>  <li><p>Let <var>body</var> be a new JSON <a>Object</a>.
> 
>  <li><p>Optionally run implemention-specific steps
> ```

I think this is generally true of any command, especially for intermediary nodes (e.g. for authentication information)

---

*[webdriver-spec.html, line 2010 at r1](https://reviewable.io:443/reviews/w3c/webdriver/334#-KSfXpwhc4r_BVg-1WnK:-KSfXpwhc4r_BVg-1WnL:brnxm3a) ([raw file](https://github.com/w3c/webdriver/blob/958b4711a7313d1c91c3b6e87e137c5b0b5cd6c7/webdriver-spec.html#L2010)):*
> ```HTML
>     <p>Otherwise true.
>  
>    <dt><a>endpoint node</a>
> ```

I don't think this is really clearer than the vauge handwavy text since you have to say "let it be true (oh except it might not be really).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/webdriver/334)*

andreastt:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[webdriver-spec.html, line 1995 at r1](https://reviewable.io:443/reviews/w3c/webdriver/334#-KSfXb3YjtdqrAH3mss_:-KSfbzM3o6IqpIbCQIvl:b-xaht9n) ([raw file](https://github.com/w3c/webdriver/blob/958b4711a7313d1c91c3b6e87e137c5b0b5cd6c7/webdriver-spec.html#L1995)):*
<details><summary><i>Previously, jgraham wrote…</i></summary>
> I think this is generally true of any command, especially for intermediary nodes (e.g. for authentication information)
</details>
The point is that it needs to happen before the `ready` and `message` fields overwrite whatever extra data the implementation throws in there.

I can reword it in a less step-by-step way to say that these two fields are reserved, but in this particular case I think it’s worth pointing out that we fully expect more data to be added to the body.

---

*[webdriver-spec.html, line 2010 at r1](https://reviewable.io:443/reviews/w3c/webdriver/334#-KSfXpwhc4r_BVg-1WnK:-KSfctdK9kMfv7OGkgGT:b-9lal51) ([raw file](https://github.com/w3c/webdriver/blob/958b4711a7313d1c91c3b6e87e137c5b0b5cd6c7/webdriver-spec.html#L2010)):*
<details><summary><i>Previously, jgraham wrote…</i></summary>
> I don't think this is really clearer than the vauge handwavy text since you have to say "let it be true (oh except it might not be really).
</details>
I’m not sure what you’re referring to. The text for the endpoint node case or the intermediary node?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/webdriver/334)*

webdriver-spec.html
```diff
@@ -1968,42 +1968,72 @@
...
+
+<p>The readiness state is represented
+ by the <code>ready</code> property of the body,
+ which is false if an attempt to <a data-lt="new session">create a session</a>
+ at the current time would fail.
+ However, the value true does not guarantee
+ that a new session request will succeed.
+
+<p>Implementations may optionally include
+ arbitrary other meta information as part of the body,
```
shs96c: Nit: No need for "arbitrary" here, as you've already said that this is the case above.

webdriver-spec.html
```diff
@@ -1968,42 +1968,72 @@
...
+    <p>Otherwise true.
+ 
+   <dt><a>endpoint node</a>
+   <dd><p>False when the <a>maximum active sessions</a>
+    is equal to the length of the list of <a>active sessions</a>
+    or there is an <a>active user prompt</a>.
+
+    <p>Otherwise true.
+
+    <p>However, there may be further
```
shs96c: Technically never reached, and incredibly hand-wavey. It would be better to default `readiness state` to `true`, and then let the unhandled case be dealt with implicitly.

webdriver-spec.html
```diff
@@ -1968,42 +1968,72 @@
...
+     implementation-specific scenarios
+     in which the value is false.
+   </dl>
 
-  <li><p><a>Set a property</a> of <var>body</var> with
-  name <code>"ready"</code> and value <var>body</var>.
+ <li><p>Let <var>readiness state message</var>
+  be an implementation-defined string
+  explaining the value given to <var>readiness state</var>.
```
shs96c: Indicate that the empty string is a legitimate response, but is only advised for when `readiness state` is `true`.

AutomatedTester:
@andreastt ping to finish up PR (when you have time)

webdriver-spec.html
andreastt: This is a reasonable point. I believe I have addressed it now.

andreastt:
@AutomatedTester Addressed code review issues. Removed some of the unnecessary stuff pointed out by jgraham and a logic inconsistency pointed out by simonstewart.

PR #335, opened 2016-09-21
Suggest visual element to identify automation windows and other security clarifications

andreastt:
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/335)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -7455,6 +7455,16 @@ session := new_session(capabilities)</code></pre>
  to limit connections to the IPv4 localhost
  CIDR range <code>127.0.0.0/8</code>
  and the IPv6 localhost address <code>::1</code>. [[!RFC4632]]
+
+<p>It is also suggested that user agents
+ make an effort to visually distinguish
+ a user agent session that is under control of WebDriver
+ from those used for normal browsing sessions.
+ This can be done through a chrome UX element
+ such as a “door hanger”,
+ colorful decoration of the OS window,
+ or some widget element that is prevalent in the window
+ so that it easy to identify automation windows.
 </section>
 
 <section class=appendix>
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -7429,21 +7429,42 @@
...
+ to test whether to enable WebDriver,
+ or alternatively make the user agent initiate
+ or confirm the connection through
+ a privileged content document or control widget,
+ in case the user agent does not
  directly implement the HTTP endpoints.
 
-<p>To prevent arbitrary machines from connecting and
- creating <a>sessions</a>, it is suggested that only connections from
+<p>It is strongly suggested that user agents
```
shs96c: Not mentioned in the commit message, but makes sense.

webdriver-spec.html
```diff
@@ -7429,21 +7429,42 @@
...
+ CIDR range <code>127.0.0.0/8</code>
+ and the IPv6 localhost address <code>::1</code>. [[!RFC4632]]
+
+<p>It is also suggested that user agents
+ make an effort to visually distinguish
+ a user agent session that is under control of WebDriver
+ from those used for normal browsing sessions.
+ This can be done through a chrome UX element
+ such as a “door hanger”,
+ colourful decoration of the OS window,
```
shs96c: I believe we use US English in the spec. "colorful"

shs96c:
I have no idea if it's possible to edit this inline.

webdriver-spec.html

webdriver-spec.html
andreastt: Good point. Added it to the commit message, but I’m happy to split it out as a separate commit if you think it is best.

AutomatedTester:
landed in a8ac274 et al

PR #337, opened 2016-09-21
Refactor waiting for navigation and response checks

andreastt:
This patch creates two new definitions to _wait for navigation to
complete_ and for _post-navigation checks_ that can be referenced and
reused in commands where we expect navigation to occur, such as Element
Click, Go, and Refresh.

It references the Web App Security specification, as HTML does, when it
looks at the outcome of the response from navigating.

The patch furthermore defines page loading strategies (and fixes a lot
of references to them throughout) and makes sure theyare taken into
consideration when waiting for the document's ready state to change.

It also adds a few examples.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/337)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -135,8 +135,16 @@ var respecConfig = {
 -->
 
 <dl>
+ <dt>Web App Security
+ <dd><p>The following terms are defined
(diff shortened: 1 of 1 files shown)


AutomatedTester:
landed in 321fff4

PR #339, opened 2016-09-26
Set the current browsing context when switching windows.

### LINKED ISSUE #338: Switch to Window doesn't specify what to do with the current browsing context
It sets the current top-level browsing context, but not the current browsing context.

The only real ambiguity comes when "Switch to Window" is called with the handle of the current top-level browsing context (i.e., is a no-op):
* Marionette does not change the current browsing context.
* All other driver implementations I'm aware of (including Microsoft WebDriver and Apple SafariDriver) set the current browsing context to the top-level context, which seems to me like the right approach.

juangj:
The current top-level browsing context is changed implicitly by doing this.

Fixes #338.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/339)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -2609,7 +2609,7 @@ session := new_session(capabilities)</code></pre>
 
  <li><p>If <var>handle</var> is equal to the associated <a>window handle</a>
   for some <a>top-level browsing context</a> in the <a>current session</a>,
-  set the <a>session</a>’s <a>current top-level browsing context</a>
+  set the <a>session</a>’s <a>current browsing context</a>
   to that browsing context, and return <a>success</a> with data null.
 
   <p>Otherwise, return <a>error</a> with <a>error code</a> <a>no such window</a>.
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -259,7 +259,6 @@ var respecConfig = {
    <!-- A browsing context is discarded --> <li><dfn data-lt=discarded><a href=https://html.spec.whatwg.org/#a-browsing-context-is-discarded>A browsing context is discarded</a></dfn>
    <!-- Active document --> <li><dfn><a href=https://html.spec.whatwg.org/#active-document>Active document</a></dfn>
    <!-- Active element --> <li><dfn>Active element</dfn> being the <a href=https://html.spec.whatwg.org/#dom-document-activeelement><code>activeElement</code></a> attribute on <a href=https://html.spec.whatwg.org/#document><code>Document</code></a>
-   <!-- Ancestor browsing context --> <li><dfn><a href=https://html.spec.whatwg.org/#ancestor-browsing-context>Ancestor browsing context</a></dfn>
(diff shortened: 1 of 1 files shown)


andreastt:
Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[webdriver-spec.html, line 2612 at r1](https://reviewable.io:443/reviews/w3c/webdriver/339#-KSfnYMgVL0BWLnalW1A:-KSfnYMgVL0BWLnalW1B:b-s7b60g) ([raw file](https://github.com/w3c/webdriver/blob/e7806f8adb40d661b9d4c28d8ee88ceef979f29f/webdriver-spec.html#L2612)):*
> ```HTML
>  <li><p>If <var>handle</var> is equal to the associated <a>window handle</a>
>   for some <a>top-level browsing context</a> in the <a>current session</a>,
>   set the <a>session</a>’s <a>current browsing context</a>
> ```

We need to set both the _current top-level browsing context_ and the _current browsing context_ because these are WebDriver-internal concepts.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/webdriver/339)*

juangj:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[webdriver-spec.html, line 2612 at r1](https://reviewable.io:443/reviews/w3c/webdriver/339#-KSfnYMgVL0BWLnalW1A:-KSh5a-hZzvUwVpqP0cR:b-9av3rn) ([raw file](https://github.com/w3c/webdriver/blob/e7806f8adb40d661b9d4c28d8ee88ceef979f29f/webdriver-spec.html#L2612)):*
I had thought that, at first, but I changed my mind because "current top-level browsing context" is currently defined in terms of "current browsing context":

>A session has an associated **current top-level browsing context**, which is the current browsing context if it itself is a top-level browsing context, or the top-level browsing context for which the current browsing context is an ancestor browsing context.

This implies that it's not sensible to set "current top-level browsing context" directly.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/webdriver/339)*

andreastt:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[webdriver-spec.html, line 2612 at r1](https://reviewable.io:443/reviews/w3c/webdriver/339#-KSfnYMgVL0BWLnalW1A:-KShKsSF54kSKo-7aSu8:b76qqen) ([raw file](https://github.com/w3c/webdriver/blob/e7806f8adb40d661b9d4c28d8ee88ceef979f29f/webdriver-spec.html#L2612)):*
<details><summary><i>Previously, juangj (Jason Juang) wrote…</i></summary>
> I had thought that, at first, but I changed my mind because "current top-level browsing context" is currently defined in terms of "current browsing context":
> 
> >A session has an associated **current top-level browsing context**, which is the current browsing context if it itself is a top-level browsing context, or the top-level browsing context for which the current browsing context is an ancestor browsing context.
> 
> This implies that it's not sensible to set "current top-level browsing context" directly.
</details>
I stand corrected, sir :smile: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/webdriver/339)*

juangj:
Actually, looking at the definition that I pasted again, the ancestor relationship there is backwards, so there's another commit to reword that.

andreastt:
This looks fine to me.

PR #342, opened 2016-09-30
Remove introduction section header in actions and surface issues

andreastt:
This change makes no functional changes to the actions chapter apart
from removing the ‘Introduction’ subchapter, and surfacing issues
that were previously in the comments.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/342)


webdriver-spec.html
```diff
@@ -5316,106 +5316,110 @@
...
-
-  </section> <!-- /Introduction -->
+<section>
+<h2>Actions</h2>
+
+<p class=issue>Expand these examples
+
+<p>The Actions API provides a low-level interface
+ for providing virtualised device input to the web browser.
+ Conecputally, the actions API divides time into a series of <a>ticks</a>.
```
shs96c: "Conceptually"

webdriver-spec.html
```diff
@@ -5316,106 +5316,110 @@
...
+ The local end sends a series of actions
+ which correspond to the change in state, if any,
+ of each input device during each tick.
+ For example, pressing a key
```
shs96c: Query: how are you formatting this? The line lengths seem terribly uneven

webdriver-spec.html
```diff
@@ -5316,106 +5316,110 @@
...
+ is represented by an action chain
+ consisting of a single key input device and two ticks,
+ the first containing a <a>keyDown</a> action,
+ and the second a <a>keyUp</a> action,
+ whereas a pinch-zoom input is represented
+ by an action chain consisting of a three ticks
+ and two pointer input devices of type touch,
+ each performing a sequence of actions <a>pointerDown</a>,
+ followed by <a>pointerMove</a>, and then <a>pointerUp</a>.
```
shs96c: A diagram would be incredibly useful here. Probably in a later revision

webdriver-spec.html
```diff
@@ -5316,106 +5316,110 @@
...
+
+<p class=issue>Move this note:
+
+<p class=note>An <a>activation trigger</a>
+ generated by the WebDriver API user
+ need to be indistinguishable from those generated
```
shs96c: s/need/needs/

webdriver-spec.html
```diff
@@ -5316,106 +5316,110 @@
...
+ by a real user interacting with the browser.
+ In particular, the dispatched events will have
+ the <a>isTrusted</a> attribute set to true.
+ The most robust way to dispatch these events
+ is by creating them in the browser implementation itself.
+ Sending OS-specific input messages to the browser's window
+ has the disadvantage that the browser being automated
+ may not be properly isolated from a user
+ accidentally modifying input source state;
+ use of an OS-level accessibility API has the disadvantage
```
shs96c: This is a new sentence.

webdriver-spec.html
```diff
@@ -5316,106 +5316,110 @@
...
+ that the browser's window must be focused,
+ and as a result, multiple WebDriver instances cannot run in parallel.
```
shs96c: Might be worth calling out that the advantage of a OS-level accessibility API is that it guarantees inputs correctly mirror user input, and allows interaction with the host OS if necessary. Though using it is a terrible idea from a machine utilisation perspective.

webdriver-spec.html
```diff
@@ -5316,106 +5316,110 @@
...
+
+<aside class="example">
```
shs96c: Once the note above is moved, you'll have an example outside of an example section, followed by this example. Consider removing the example in the paragraphs above.

webdriver-spec.html
```diff
@@ -5316,106 +5316,110 @@
...
+
+ <p>When the <a>remote end</a> receives this,
+  it will look at each input source's action lists.
+  It will dispatch the first action of each source together,
+  then the second actions together, and lastly, the final actions together.
+
+ <p>The diagram below displays when each action gets executed.
+  "Source 1" is the first finger, and "source 2" is the second.
+
+ <p><img src=graphics/note1actions.svg alt="">
```
shs96c: Fantastic.

webdriver-spec.html
```diff
@@ -5316,106 +5316,110 @@
...
+
+ <p>There is no limit to the number of input sources,
+  and there is no restriction regarding the length of each input's action list.
+  Meaning, there is no requirement that all action lists have to be the same length.
```
shs96c: Nit: "This means" reads better than "meaning". If you want to use "meaning", use it in the middle of the sentence.

webdriver-spec.html
andreastt: By [semantic linefeeds](http://rhodesmill.org/brandon/2012/one-sentence-per-line/), which I find particularly useful for maintaining prose in VCSes. I don’t expect you to do the same.

webdriver-spec.html
andreastt: I didn’t write this.

webdriver-spec.html
andreastt: I agree. And probably some of this should be rewritten. This PR is surfacing the known problems.

webdriver-spec.html
andreastt: All good points. Addressed.

webdriver-spec.html
andreastt: I will leave this for a later PR.

webdriver-spec.html
andreastt: Again, I didn’t write this, but I’ve addressed it.

AutomatedTester:
landed in 4df527f

PR #343, opened 2016-09-30
Define extension capabilities

andreastt:
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/343)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -1405,7 +1405,7 @@ var respecConfig = {
  <a data-lt="extension command URL">URL</a>.
 
 <p>The <a>extension command</a>’s <dfn>extension command URL</dfn>
- is a <a>URL</a> composed of the <a>extension prefix</a>,
+ is a <a>URL</a> composed of the <a>extension command prefix</a>,
  followed by "<code>/</code>",
  and the <a>extension command</a>’s <a data-lt="extension command name">name</a>.
  The <a>extension command URL</a>,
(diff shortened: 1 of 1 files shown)


shs96c:
I need to do a review, but capabilities should be in the firstMatch or alwaysMatch fields, not as new top level properties of the new session command.

webdriver-spec.html
```diff
@@ -1426,16 +1426,68 @@
...
+<p>As with <a>extension commands</a>,
+ it is suggested that the key used to denote
+ the <a>extension capability</a> namespace
+ is based on the <a href=http://www.w3.org/TR/CSS21/syndata.html#vendor-keywords>vendor keywords</a>
+ listed in [[!CSS21]]
+ and preceeds the first "<code>:</code>" character in the string.
+
+<aside class=example>
+ <p><a>Extension capabilities</a> are typically used
+  to provide UA- or <a>intermediary node</a> specific configuration
```
shs96c: s/UA-/UA/

webdriver-spec.html
```diff
@@ -1426,16 +1426,68 @@
...
+  that is not handled by the <a>table of capabilities</a>.
+
+ <p>An example <a>new session</a> request body
+  might look like this:
+
+ <pre>{
+	"alwaysMatch": …,
+	"firstMatch": …,
```
shs96c: Missing opening brace.

andreastt: There is an opening brace on the `<pre>` line.

webdriver-spec.html

andreastt:
> I need to do a review, but capabilities should be in the firstMatch or alwaysMatch fields, not as new top level properties of the new session command. 

I think I will fix this by not saying explicitly where it should be.

shs96c:
LGTM. Merging.

PR #345, opened 2016-10-04
Always return response data in a value field

shs96c:
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/345)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -734,9 +734,8 @@ var respecConfig = {
   to the description of <var>status</var> in the <a>status code registry</a>.
 
  <li><p>If <var>data</var> is not null, let <var>response</var>’s
-  <a>body</a> be a JSON object with a top-level property
-  of <var>value</var>, the value of which is the result
-  of <a>serializing as JSON</a> with <var>data</var> as the argument.
+  <a>body</a> be a JSON <a>Object</a> with a key <code>value</code>
+  set to the <a>JSON serialisation</a> of <var>data</var>.
 
  <li><p>Let <var>response bytes</var> be the byte sequence resulting
   from serializing <var>response</var> according to the rules in [[!RFC7230]].
@@ -2249,8 +2248,7 @@ session := new_session(capabilities)</code></pre>
   of the <a>current top-level browsing context</a>’s
   <a>active document</a>’s <a>document URL</a>.
 
- <li><p>Return <a>success</a> with data set
-  to <var>url</var>.</p></li>
+ <li><p>Return <a>success</a> with data <var>url</var>.</p></li>
 </ol>
 </section>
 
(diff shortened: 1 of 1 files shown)


andreastt:
For the benefit of posterity, there is relevant discussion about this PR in https://lists.w3.org/Archives/Public/public-browser-tools-testing/2016OctDec/0001.html.

(At some point we should also move discussions onto here.)

webdriver-spec.html
```diff
@@ -734,8 +734,9 @@
   to the description of <var>status</var> in the <a>status code registry</a>.
 
  <li><p>If <var>data</var> is not null, let <var>response</var>’s
-  <a>body</a> be the result of <a>serializing as JSON</a>
-  with <var>data</var> as the argument.
+  <a>body</a> be a JSON object with a top-level property
```
andreastt: `JSON <a>Object</a>`

andreastt: I also think saying “top-level property” is redundant. You can just say `… JSON <a>Object</a> with a key <code>value</code> set to the <a>JSON serialisation</a> of <var>data</var>`.

webdriver-spec.html
```diff
@@ -734,8 +734,9 @@
...
+  of <var>value</var>, the value of which is the result
```
andreastt: This should be `<code>value</code>` as it’s not a variable in the algorithm, but a key on the JSON Object.

webdriver-spec.html
```diff
@@ -2248,10 +2249,8 @@
   of the <a>current top-level browsing context</a>’s
   <a>active document</a>’s <a>document URL</a>.
 
- <li><p>Let <var>body</var> be a JSON Object
-  with the "<code>value</code>" property set to <var>url</var>.
-
- <li><p>Return <a>success</a> with data <var>body</var>.</p></li>
+ <li><p>Return <a>success</a> with data set
+  to <var>url</var>.</p></li>
```
andreastt: `with data <var>url</var>` to be consistent with the other algorithms.

webdriver-spec.html
```diff
@@ -2441,12 +2440,8 @@
...
  context</a>'s <a>active document</a>.
 
- <li><p>Let <var>body</var> be a new JSON Object.
-
- <li><p>Set the property "<code>value</code>" of <var>body</var>
-  to the value of <var>title</var>.
-
- <li><p>Return <a>success</a> with data <var>body</var>.
+ <li><p>Return <a>success</a> with data being the value
+ of <var>title</var>.
```
andreastt: Same here. Please let us just say “Return success with data _title_” throughout this patch.

andreastt:
Also, I don’t believe you are handling the return value from finding single- and multiple elements correctly. This requires a change too.

shs96c:
Updated the patch. 

If we were handling finding elements correctly before, we will be doing so now. The change to how we encode responses should mean that where we used to return the raw element or array of elements we now return a JSON object with the value set properly.

webdriver-spec.html

webdriver-spec.html
shs96c: Fixed by the above

webdriver-spec.html

webdriver-spec.html
shs96c: Done and done.

andreastt:
Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/webdriver/345)*

PR #428, opened 2016-10-18
Make small corrections to Actions

mjzffr:
Little things I noticed while implementing the spec in Marionette. @jgraham r?

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/428)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -5857,7 +5857,7 @@ session := new_session(capabilities)</code></pre>
         <li><p>If <var>type</var> is equal to <code>"pointer"</code>,
             let <var>parameters data</var> be the result of getting a
             property named <var>parameters</var>
-            from <var>action</var>. Then let <var>parameters</var> be
+            from <var>action sequence</var>. Then let <var>parameters</var> be
             the result of <a>trying</a> to <a>process pointer
             parameters</a> with argument <var>parameters
             data</var>.</li>
@@ -5951,7 +5951,7 @@ session := new_session(capabilities)</code></pre>
           </ol>
       </li>
 
-      <li><p>Return <a>success</a> with data <var>primary</var>.</li>
+      <li><p>Return <a>success</a> with data <var>parameters</var>.</li>
 
     </ol>
 
@@ -5978,11 +5978,13 @@ session := new_session(capabilities)</code></pre>
           constructed with arguments <var>id</var>,
           <code>"none"</code>, and <var>subtype</var>.</li>
 
-      <li><p>Let <var>result</var> be the result of <a>trying</a>
-          to <a>process a pause action</a> with arguments <var>action
-          item</var>, and <var>action object</var>.</li>
+      <li><p>If <var>subtype</var> is <code>"pause"</code>;
+          let <var>result</var> be the result of <a>trying</a> to
+          <a>process a pause action</a> with arguments <var>action
+          item</var>, and <var>action</var>. If doing so results in
+          an <a>error</a>, return that error.</li>
 
-      <li><p>Return <var>result</var>.</li>
+      <li><p>Return <a>success</a> with data <var>action</var>.</li>
 
     </ol>
 
@@ -6006,8 +6008,9 @@ session := new_session(capabilities)</code></pre>
       <li><p>If <var>subtype</var> is <code>"pause"</code>;
           let <var>result</var> be the result of <a>trying</a> to
           <a>process a pause action</a> with arguments <var>action
-          item</var>, and <var>action object</var>, and
-          return <var>result</var>.</li>
+          item</var>, and <var>action</var>. If doing so results in
+          an <a>error</a>, return that error, otherwise return
+          <var>action</var>.</li>
 
       <li><p>Let <var>key</var> be the result of getting a property
       named <code>value</code> from <var>action item</var>.</li>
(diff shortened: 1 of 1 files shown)


jgraham:
Review status: 0 of 1 files reviewed at latest revision, 4 unresolved discussions.

---

*[webdriver-spec.html, line 5984 at r1](https://reviewable.io:443/reviews/w3c/webdriver/428#-KUNzZYrSsOoQs1S19hz:-KUNzZYsjwgWMwBqKWBG:b6xdjur) ([raw file](https://github.com/w3c/webdriver/blob/5f3ce807a26fe286d79ad8ccd694fa05016120bf/webdriver-spec.html#L5984)):*
> ```HTML
>           let <var>result</var> be the result of <a>trying</a> to
>           <a>process a pause action</a> with arguments <var>action
>           item</var>, and <var>action</var>. If doing so results in
> ```

"trying" is magic for "do x and unwrap the result if it's a success or otherwise return the error".

---

*[webdriver-spec.html, line 5987 at r1](https://reviewable.io:443/reviews/w3c/webdriver/428#-KUOJJcPFQLut7AxvLpf:-KUOJJcQMSJkzIpnCI-j:bgwx3rg) ([raw file](https://github.com/w3c/webdriver/blob/5f3ce807a26fe286d79ad8ccd694fa05016120bf/webdriver-spec.html#L5987)):*
> ```HTML
>           an <a>error</a>, return that error.</li>
> 
>       <li><p>Return <a>success</a> with data <var>action</var>.</li>
> ```

I think this is supposed to return success with data *result/*  It should perhaps be more explicit that *subtype* having some other value would be a bug in the spec e.g. stating *subtype has value `"pause"`*.

---

*[webdriver-spec.html, line 6011 at r1](https://reviewable.io:443/reviews/w3c/webdriver/428#-KUOIslwJ-Ezm-o6ZciO:-KUOIsly6w9fQ1yDDdQE:b-euwvd4) ([raw file](https://github.com/w3c/webdriver/blob/5f3ce807a26fe286d79ad8ccd694fa05016120bf/webdriver-spec.html#L6011)):*
> ```HTML
>           let <var>result</var> be the result of <a>trying</a> to
>           <a>process a pause action</a> with arguments <var>action
>           item</var>, and <var>action</var>. If doing so results in
> ```

See above about "trying" and note that this has a type error (if one return option is an error the other must be success with data).

---

*[webdriver-spec.html, line 6058 at r1](https://reviewable.io:443/reviews/w3c/webdriver/428#-KUOJyfeM8HC7Uw0XF4J:-KUOJyffQVoMu5Po7n1r:bribihm) ([raw file](https://github.com/w3c/webdriver/blob/5f3ce807a26fe286d79ad8ccd694fa05016120bf/webdriver-spec.html#L6058)):*
> ```HTML
>           let <var>result</var> be the result of <a>trying</a> to
>           <a>process a pause action</a> with arguments <var>action
>           item</var>, and <var>action</var>. If doing so results in
> ```

As above.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/webdriver/428)*

jgraham:
Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/webdriver/428)*

mjzffr:
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[webdriver-spec.html, line 5883 at r1](https://reviewable.io:443/reviews/w3c/webdriver/428#-KUSK83WEGMpM0mieMij:-KUSK83WEGMpM0mieMik:b-p9mwkb) ([raw file](https://github.com/w3c/webdriver/blob/5f3ce807a26fe286d79ad8ccd694fa05016120bf/webdriver-spec.html#L5883)):*
> ```HTML
>               <li><p>If <var>type</var> is <code>"none"</code>
>                 let <var>action</var> be the result of <a>trying</a>
>                 to <a>process a null action</a> with
> ```

My changes below with the use of "process a pause action" aim to get this statement (among others) to type-check: here, we expect "process a null action" to return an action, but it actually returns "success with data null" from "process a pause action".

---

*[webdriver-spec.html, line 5981 at r1](https://reviewable.io:443/reviews/w3c/webdriver/428#-KUSIZKUcBOzMUWdTH9u:-KUSIZKUcBOzMUWdTH9v:b3vpiht) ([raw file](https://github.com/w3c/webdriver/blob/5f3ce807a26fe286d79ad8ccd694fa05016120bf/webdriver-spec.html#L5981)):*
> ```HTML
>           <code>"none"</code>, and <var>subtype</var>.</li>
> 
>       <li><p>If <var>subtype</var> is <code>"pause"</code>;
> ```

Whoops, I pasted in this line of text unnecessarily. 

---

*[webdriver-spec.html, line 5987 at r1](https://reviewable.io:443/reviews/w3c/webdriver/428#-KUOJJcPFQLut7AxvLpf:-KUSJ0gEREIlLq7oQ648:b-igxcok) ([raw file](https://github.com/w3c/webdriver/blob/5f3ce807a26fe286d79ad8ccd694fa05016120bf/webdriver-spec.html#L5987)):*
<details><summary><i>Previously, jgraham wrote…</i></summary>
> I think this is supposed to return success with data *result/*  It should perhaps be more explicit that *subtype* having some other value would be a bug in the spec e.g. stating *subtype has value `"pause"`*.
</details>
The return value of "process a pause action" is either *error* or *success with data null*, with a side-effect of modifying *action*, so here in case of success I think we actually want to return *action*.  (See how "process a null action" is used above in "process an input source action sequence" -- it's expecting an action.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/webdriver/428)*

mjzffr:
Comments addressed. @jgraham r?

jgraham:
Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/webdriver/428)*

jgraham:
@AutomatedTester r+ please merge

PR #432, opened 2016-10-25
Describe what happens when a selector is equal to "". Fixes #420

AutomatedTester:
With findElements we return an empty list e.g. []
With findElement we return elementNotFoundError

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/432)

PR #434, opened 2016-10-25
Correct start nodes for element retrieval and error on missing args

andreastt:
These two patches corrects the start nodes for element retrieval so that they don’t use the `<html>` element directly but instead the current document’s document element, and return the appropriate errors if the `value` and `using` fields are missing from the request bodies.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/434)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -4090,9 +4090,17 @@ entries:
  <li><p>Let <var>start node</var> be
   the <a>current browsing context</a>’s <a>document element</a>.
 
- <li><p>Let <var>location strategy</var> be the result of <a>getting a property</a> called "<code>using</code>".
+ <li><p>Let <var>location strategy</var> be the result
+  of <a>getting a property</a> called "<code>using</code>".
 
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -4078,20 +4078,29 @@
...
+ <li><p>If <var>location strategy</var> is <a>undefined</a>,
+  return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
 
- <li><p>Let <var>location strategy</var> be the result of <a>getting a property</a> called "<code>using</code>".
+ <li><p>Let <var>selector</var> be the result
+  of <a>getting a property</a> called "<code>value</code>".
 
- <li><p>Let <var>selector</var> be the result of <a>getting a property</a> called "<code>value</code>".
+ <li><p>If <var>selector</var> is <a>undefined</a>,
+  return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
```
AutomatedTester: This is divergent from what Selenium does, see PR #432

andreastt: The patch in #432 checks for an empty string and null, not that they are missing.

PR #441, opened 2016-11-14
Check length of DOMRect sequence on pointer interactability

andreastt:
The DOMRect sequence returned from `getClientRects()` can be of length
0, which happens when the element’s `display` style property has the
computed value of `none`, amongst other things.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/441)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -3797,13 +3797,11 @@ named "<code>firstMatch</code>" from <var>capabilities request</var>.
  is the centre point of the area
  of the first <a data-lt="getClientRects">DOM client rectangle</a>
  that is inside the <a>viewport</a>.
- It can be calculated this way:
+ It can be calculated this way,
+ given an element, here called <var>rectangle</var>,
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -3840,8 +3838,20 @@
...
+  An element which style property `display` is "none"
+  returns an empty DOMRect sequence.
+ -->
+ <li><p>If <var>rectangles</var> has the length of 0,
+  return an empty sequence.
+
  <li><p>Let <var>centre point</var> be
-  the <a>in-view centre point</a> of <var>element</var>.
+  the <a>in-view centre point</a> of <var>element</var>
+  given the first indexed element of <var>rectangles</var>.
```
andreastt: Maybe we should do something more sophisticated here, but I feel we should land this change and file an issue about the edge case.

AutomatedTester: agreed, please raise a bug when you get a moment

PR #442, opened 2016-11-15
Correct math for finding in-view centre point

### LINKED ISSUE #425: Wrong math in the element center calculation
In chapter [9. Elements
    ](https://www.w3.org/TR/webdriver/#h-elements), section [9.1 Element Interactability
      ](https://www.w3.org/TR/webdriver/#h-element-interactability)::

> An element’s in-view centre point is the centre point of the area of the first DOM client rectangle that is inside the viewport. It can be calculated this way:
> 
> Let rectangle be the first element of the DOMRect sequence returned by calling getClientRects on element.
> 
> Let visible width be the rectangle’s (max(x coordinate, x coordinate + width dimension)).
> 
> Let visible height be the rectangle’s (max(y coordinate, y coordinate + height dimension)).
> 
> Let vertical centre offset be (visible width / 2.0).
> Let horizontal centre offset be (visible height / 2.0).
> Let centre point be a pair of the rectangle’s (x coordinate + horizontal centre offset) and (y coordinate + vertical centre offset).
> Return centre point.

Let's omit typos vertical<-> horizontal.

let x = 1000, let width = 200.
Visible width = max(1000, 1000 + 200) = 1200.
Horizontal centre offset = 1200 / 2 = 600.
Horizontal centre point = 1000 + 600 = 1600.
Which is out of the element rectangle boundaries.
Such a math seems to be wrong.

I think the x center should be calculated as:

0.5 \* ( max(0, min(x, x + width)) + min(window.innerWidth, max(x, x + width)) )

andreastt:
Thanks to @Dzenly for fixing this!

Fixes #425.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/442)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -3805,32 +3805,33 @@ named "<code>firstMatch</code>" from <var>capabilities request</var>.
  off a <a><code>DOMRect</code></a> sequence:
 
 <ol>
- <!--
-  no need to care for scroll bars here
-  because elementsFromPoint takes care of it
+ <li><p>Let <var>rectangle</var> be
+  the first element of the <a><code>DOMRect</code></a> sequence
+  returned by calling <a>getClientRects</a> on <var><a>element</a></var>.
 
-  this is the same as
+ <li><p>Let <var>left</var> be
+  (<a>max</a>(0, <a>min</a>(<a>x coordinate</a>,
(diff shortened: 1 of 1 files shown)


AutomatedTester:
can you rebase and land please?

PR #475, opened 2017-01-09
Add dfn for plural of tick

vkatsikaros:
Fix respec warning.

Found linkless <a> element with text 'ticks' but no matching <dfn>.   respec-w3c-common:1:26900

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/475)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -6030,7 +6030,7 @@ named "<code>firstMatch</code>" from <var>capabilities request</var>.
         </tr>
       </table>
 
-    <p>A <dfn data-lt="tick|ticks">tick</dfn> is the basic unit of time over which actions
+    <p>A <dfn data-lt="ticks">tick</dfn> is the basic unit of time over which actions
       can be performed. During a tick, each input source has an assigned
       action — possibly a noop <a>pause</a> action — which may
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -6030,7 +6030,7 @@
         </tr>
       </table>
 
-    <p>A <dfn>tick</dfn> is the basic unit of time over which actions
+    <p>A <dfn data-lt="tick|ticks">tick</dfn> is the basic unit of time over which actions
```
andreastt: The first `tick` is implicit here and can be dropped.

vkatsikaros: Ah I see, I used as example `<dt><dfn data-lt="remote end|remote ends|remote">Remote End</dfn>` https://github.com/w3c/webdriver/blob/master/webdriver-spec.html#L541 

@andreastt in that case is the `remote end` in `data-lt` is due to upper/lower case?

andreastt: It doesn’t care about upper- or lower case, but “remote end” in that list was probably added there before ReSpec supported taking the inner text implicitly. It’s safe to remove it, but that should probably be done in a separate commit.

PR #478, opened 2017-01-09
Add DOM spec dependency 'fire an event'

vkatsikaros:
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/478)


webdriver-spec.html
```diff
@@ -2558,9 +2559,9 @@
  the point at which the command returns
  is determined by the session’s <a>page loading strategy</a>.
  The <a>normal</a> state causes it
- to return after the load event fires on the new page,
+ to return after the load event <a>fires</a> on the new page,
```
AutomatedTester: I think `<a>event fires</a>` would be better. Could you add that to the `data-lt`

PR #487, opened 2017-01-10
Alter pointerMove algorithm to support relative moves.

shs96c:
As discussed on the [mailing list](https://lists.w3.org/Archives/Public/public-browser-tools-testing/2017JanMar/0004.html).

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/487)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -6559,12 +6559,12 @@ named "<code>firstMatch</code>" from <var>capabilities request</var>.
            if <a href="#dfn-get-element-rect" class="internalDFN"
            data-link-type="dfn">Get Element Rect</a> had been called.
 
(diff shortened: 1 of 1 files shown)


jgraham:
This doesn't seem to work how I expected. I expected that one would allow origin to either be `"viewport"`, `"pointer"` or a web element reference. Then I expected the `x` and `y` properties to remain unchanged except the viewport `x` and `y` would be calculated relative to the provided origin, which would either be the top left of the viewport, the current internal pointer location, or the visible centre of the provided element. 

In addition this change is partially in the wrong place; you can't work out the actual x,y coordinates until you dispatch the action (since other actions may have changed things in the interim).

shs96c:
@jgraham has an alternative proposal for this. Let's work off that.

Issue #438, opened 2016-11-04
Clarify whether Element Retrieval should always respect the implicit wait timeout


qmfrederik:
When reading the WebDriver standard, it is not clear to us whether [11.2 Find Element](https://www.w3.org/TR/webdriver/#find-element) and [11.3 Find Elements](https://www.w3.org/TR/webdriver/#find-elements) should both respect the [session implicit wait timeout](https://www.w3.org/TR/webdriver/#dfn-session-implicit-wait-timeout).

Both Find Element and Find Elements reference the [Find](https://www.w3.org/TR/webdriver/#dfn-find) method, which merely references the element location strategy but does not reference the timeout.

We would _naively_ expect that Find Element respects the timeout and Find Elements does not, but this is not made explicit in the standard.

Would it be possible to make it explicit in the [Find](https://www.w3.org/TR/webdriver/#dfn-find) section that the implicit timeout should be respected, and to clarify whether this is the case for both Find Element and Find Elements or Find Element only?

Thanks!

andreastt:
Note that there is an [ongoing discussion](https://lists.w3.org/Archives/Public/public-browser-tools-testing/2016OctDec/thread.html) about whether implicit waits should also apply to commands that do element interaction.

qmfrederik:
@andreastt Thanks! Followed the discussion & and submitted feedback to the poll :)

shs96c:
The result of the discussion were that implicit waits should be in the spec, and that they should be used for waiting for an element to be interactable. They should not be used for "find" and I'll update the spec accordingly.

shs96c:
I've opened PR #488 to address this.

InstyleVII:
I should note the spec does define that element retrieval should respect implicit wait already here:
https://w3c.github.io/webdriver/webdriver-spec.html#dfn-session-implicit-wait-timeout

> A session has an associated session implicit wait timeout that specifies a time to wait for the implicit element location strategy when retreiving elements. Unless stated otherwise it is zero milliseconds.

qmfrederik:
@InstyleVII The problem with that paragraph is that the spec does not define what the "implicit element location strategy" is, nor does it give any details about how the implicit wait should work, especially in the context of wait - by polling? At what frequency?

@shs96c would it make sense to remove the paragraph @InstyleVII refers to from the spec as implicit wait no longer applies to "find"?

InstyleVII:
[...]

PR #489, opened 2017-01-10
Make clear that element retrieval waits for no-one

shs96c:
Closes issue #438.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/489)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -3920,9 +3920,8 @@ named "<code>firstMatch</code>" from <var>capabilities request</var>.
  Element retrieval searches are performed
  using pre-order traversal of the document’s nodes
  that match the provided selector’s expression, and ignore the value
- of the <a href="#dfn-page-load-strategy" class="internalDFN"
- data-link-type="dfn">page load strategy</a>, acting as if it were set
- to <code>"none"</code>.
+ of the <a>page load strategy</a>, acting as if it were set
+ to <code>"none"</code>. 
  Elements are <a data-lt="JSON serialisation of an element">serialised</a>
  and returned as <a>web elements</a>.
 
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -3919,7 +3919,10 @@
  allow lookup of individual elements and collections of elements.
  Element retrieval searches are performed
  using pre-order traversal of the document’s nodes
- that match the provided selector’s expression.
+ that match the provided selector’s expression, and ignore the value
+ of the <a href="#dfn-page-load-strategy" class="internalDFN"
```
AutomatedTester: this just needs to be `<a>page load strategy</a>`.

webdriver-spec.html
```diff
@@ -3919,7 +3919,10 @@
  allow lookup of individual elements and collections of elements.
  Element retrieval searches are performed
  using pre-order traversal of the document’s nodes
- that match the provided selector’s expression.
+ that match the provided selector’s expression, and ignore the value
```
AutomatedTester: Why does this need to be mentioned? Find Elements is not, as currently written, influenced by Page load strategy.

webdriver-spec.html
andreastt: Ignoring the value of the page load strategy makes no sense in this context. The page load strategy only applies when, well, actually loading a document.

webdriver-spec.html
andreastt: +1

shs96c: Because of issue #438, which is why I linked to it in the description.

webdriver-spec.html
shs96c: See #438. If someone found this unclear, and it's not obfuscating things to clarify, it seems wise to help clear things up.

andreastt: I did see #438, which is about _implicit wait_ strategy not being mentioned in element retrieval. The element retrieval algorithm doesn’t explain what the driver should do when implicitly waiting for elements to appear in the DOM.

shs96c:
Thanks to @andreastt and @AutomatedTester for gently pointing out that it's the implicit wait strategy that should be used.

Issue #419, opened 2016-10-12
Add back Find Element by Tag Name


AutomatedTester:
This was agreed back in Redmond that we would add it back.

shs96c:
Added

PR #490, opened 2017-01-10
Add back the "tag name" element location strategy.

shs96c:
Closes issue #419.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/490)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -128,6 +128,7 @@ var respecConfig = {
    <!-- Fire an event --> <li><dfn data-lt="fire|fires|fired|event fires|event fired"><a href=https://dom.spec.whatwg.org/#concept-event-fire>Fire an event</a></dfn>
    <!-- Get an attribute by name --> <li><dfn data-lt="getting an attribute by name"><a href=https://dom.spec.whatwg.org/#concept-element-attributes-get-by-name>Get an attribute by name</a></dfn>
    <!-- getAttribute --> <li><dfn><a href=https://dom.spec.whatwg.org/#dom-element-getattribute>getAttribute</a></dfn>
+   <!-- getElementsByTagName --> <li><dfn><a href="https://dom.spec.whatwg.org/#dom-element-getelementsbytagname"><code>getElementsByTagName</code></a></dfn>
    <!-- hasAttribute --> <li><dfn data-lt="has the attribute"><a href=https://dom.spec.whatwg.org/#dom-element-hasattribute>hasAttribute</a></dfn>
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -4098,6 +4103,23 @@
...
 
 <section>
+<h4>Tag Name</h4>
+
+<p>To find a <a>web element</a> with the <dfn>Tag
+ Name</dfn> <a>strategy</a> the following steps need to be completed:
+
+<ol>
+ <li><p><p>Let <var>elements</var> be the result of
+ calling <a>getElementsByTagName</a> with
```
andreastt: Missing reference to https://dom.spec.whatwg.org/#dom-element-getelementsbytagname.

webdriver-spec.html
```diff
@@ -4098,6 +4103,23 @@
...
+ argument <var>selector</var>, with the <a>context object</a> equal to
+ the <var>start node</var>.
+
+ <li><p>Return <var>elements</var>.
```
andreastt: I don’t think we need the steps here at all: all this does is assign to _elements_, then returning it.

The above can be shortened to:

> Fo find an element with the Tag Name strategy, return the result of calling _start node_’s _getElementsByTagName_ with the argument _selector_.

webdriver-spec.html
shs96c: Fixed.

webdriver-spec.html

shs96c:
Apparently, rebasing and force pushing causes GH to not realise that reviews have been acted upon.

webdriver-spec.html
```diff
@@ -128,6 +128,7 @@
    <!-- Fire an event --> <li><dfn data-lt="fire|fires|fired|event fires|event fired"><a href=https://dom.spec.whatwg.org/#concept-event-fire>Fire an event</a></dfn>
    <!-- Get an attribute by name --> <li><dfn data-lt="getting an attribute by name"><a href=https://dom.spec.whatwg.org/#concept-element-attributes-get-by-name>Get an attribute by name</a></dfn>
    <!-- getAttribute --> <li><dfn><a href=https://dom.spec.whatwg.org/#dom-element-getattribute>getAttribute</a></dfn>
+   <!-- getElementsByTagName --> <li><dfn><a href="https://dom.spec.whatwg.org/#dom-element-getelementsbytagname">getElementsByTagName</a></dfn>
```
AutomatedTester: can you wrap `getElementsByTagName` in `<code>` tag e.g. `<code>getElementsByTagName</code>` and then update the text below to wrap it there too.

PR #495, opened 2017-01-11
Allow specifying an origin for pointer moves.

jgraham:
Add an origin property which allows pointer moves to either be relative to the viewport,
relative to the current pointer location, or relative to the centre point of an element.
This replaces the old element property and permits moveTo sementics from the current Selenium
API to be implemented.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/495)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -6517,30 +6517,32 @@ named "<code>firstMatch</code>" from <var>capabilities request</var>.
           item</var>.</li>
 
       <li><p>If <var>duration</var> is not <a>undefined</a>
-       and <var>duration</var> is not an <a>Integer</a> greater than or equal to 0,
-       return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
+          and <var>duration</var> is not an <a>Integer</a> greater than or equal to 0,
+          return <a>error</a> with <a>error code</a> <a>invalid argument</a>.</li>
 
       <li><p>Set the <code>duration</code> property
-       of <var>action</var> to <var>duration</var>.
+          of <var>action</var> to <var>duration</var>.</li>
 
-      <li><p>Let <var>element</var> be the result of getting a
-       property named <code>element</code> from <var>action item</var>.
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -7391,23 +7392,54 @@
...
+                      + <var>x offset</var>, and <var>y</var>
+                      equal <var>y element</var> + <var>y
+                      offset</var>.</li>
+                </ol>
+              </dd>
+            </dl>
         </li>
 
+        <li><p>If <var>x</var> is less than 0 or greater than the width
+            of the viewport in <a>CSS reference pixels</a>, then
```
shs96c: Do we need to update Get Window Size to also return the size of the viewport?

jgraham:
> How do handle the case where the user wants to drag element A on to element B but can't get both elements on to the viewport at the same time? When using xvfb or similar, the default screen size is normally tiny.

I feel like 3 is the right option here because it gives us a useful low-level primitive that composes with the other low level primitives to give the behaviour we want. Option 2 seems to be adding special magic for one use case that doesn't necessarily help with other potential cases (e.g. an infinite scrolling site might want to scroll down 1000px and check that some new content appears. That wouldn't correspond to targeting a specific element because the element wouldn't exist at the time of the action being initiated).

I also feel that we should defer adding this until we have a solid implementation of the existing API.

webdriver-spec.html
jgraham: I'm happy to do that (it seems way more useful to me in general), but I think it should go via the list.

shs96c:
:lgtm:

---

Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/webdriver/495#-:-KaI5vu2pXebxPBHzGFV:bg7igk9)*

shs96c:
Discussion taken on to the list. Merging this in.

PR #496, opened 2017-01-11
Set the header in the response sent back to local end. Fixes #349.

### LINKED ISSUE #349: Content Type header should be specified on the response
James Graham:

> Presumably application/json. Probably some other headers are also needed.

AutomatedTester:
When sending responses back to the local end we need to set the header
name 'Content-Type' with value 'application/json'.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/496)


webdriver-spec.html
```diff
@@ -778,6 +782,10 @@
   and <a>status message</a> to the string corresponding
   to the description of <var>status</var> in the <a>status code registry</a>.
 
+ <li><p><a data-lt='set header'></a>Set</a> the <var>response</var>'s <a>header</a>
```
andreastt: Stray end tag `</a>`.

AutomatedTester: fixed.

PR #503, opened 2017-01-12
Unify the return values for window related commands

vkatsikaros:
Commands "Get Window Size" and "Set Window Size" at their last success step return

> Return success with the JSON serialisation of the current top-level browsing context’s window size.

Whereas "Maximize Window" and "Fullscreen Window" at their last success step return

> Return success with the a new JSON Object:
>
> width
>    Current top-level browsing context’s window dimension width.
> height
>    Current top-level browsing context’s window dimension height.

which is effectively the same as the first one.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/503)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -3591,18 +3591,8 @@ named "<code>firstMatch</code>" from <var>capabilities request</var>.
   containing the <a>current top-level browsing context</a>
   to the maximum available size allowed by the window manager.
 
- <li><p>Return <a>success</a>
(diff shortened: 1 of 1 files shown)


shs96c:
:lgtm:

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/webdriver/503#-:-KaHtA1P_LgNzMr4anfM:bnfp4nl)*

PR #511, opened 2017-01-13
Correct <a> 'input cancel actions'

vkatsikaros:
The concept "input cancel actions" is undefined and from reading the
algorthm it looks this step should be emptying the well defined
"input cancel list".

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/511)


vkatsikaros:
If the above is correct this raises a question about step: chapter [16. Actions](https://w3c.github.io/webdriver/webdriver-spec.html#release-actions#h-actions), section [16.6 Release Actions](https://w3c.github.io/webdriver/webdriver-spec.html#release-actions#h-release-actions)::
> Dispatch tick actions with arguments undo actions and duration 0.

but I'll file this as separate issue.

Issue #449, opened 2016-12-14
Standard capabilities should be lower-cased


andreastt:
In chapter [6. Capabilities](https://w3c.github.io/webdriver/webdriver-spec.html#capabilities#h-capabilities), the standard capabilities _Page Load Strategy_, _Proxy Configuration_ and _Session Timeouts Configuration_ should all be lower cased to match the typography used elsewhere in the document.

PR #518, opened 2017-01-16
Fix case of standard capabilities in the document to be consistent. F…

AutomatedTester:
…ixes #449

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/518)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -1636,21 +1636,21 @@ with a "-moz-" prefix:
   </tr>
 
   <tr>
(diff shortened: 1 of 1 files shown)


shs96c:
I'm not sure about this. Either the casing refers to the name of something (in which case we should capitalise the entire name, David burns) or we should lower case the whole thing.

shs96c:
LGTM.

AutomatedTester:
landed this already

PR #520, opened 2017-01-16
Add The Unicode Standard and dependency for 'codepoint'

vkatsikaros:
Also added 'unicode codepoint' links

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/520)


webdriver-spec.html
```diff
@@ -6421,7 +6427,7 @@
       named <code>value</code> from <var>action item</var>.</li>
 
       <li><p>If <var>key</var> is not a <a>String</a> containing a
-          single unicode codepoint <span class="issue">or grapheme
+          single <a>unicode codepoint</a> <span class="issue">or grapheme
```
shs96c: The unicode spec appears to use "code point" rather than "codepoint". We should correct our usage to match.

PR #524, opened 2017-01-16
Fix case of standard capabilities in the document to be consistent. F…

AutomatedTester:
…ixes #449

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/524)


AutomatedTester:
Not sure what we should put for browserName,platform etc so thoughts welcome

shs96c:
This appears to be a duplicate of #518. Closing this PR without merging and using that one instead.

PR #525, opened 2017-01-16
Return error is Element Click will fire events on different element. …

### LINKED ISSUE #395: Specific error code doesn't exist for blocked element not being clickable
James Reed:

> There is no entry in https://w3c.github.io/webdriver/webdriver-spec.html#handling-errors for an error code that is specific for an element not being clickable because it is blocked by another element being in front of it. I currently have a pull request open (https://github.com/SeleniumHQ/selenium/pull/1967) and was asked to open this spec bug to support it.

AutomatedTester:
…Fixes #395

If an element at the coordinates for a click, when doing hit testing,
is not the same element that was requested in Element Click return
the error Element Intercepting Click.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/525)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -1263,9 +1263,9 @@ var respecConfig = {
  </tr>
 
  <tr>
-  <td><dfn>element intercepting click</dfn>
+  <td><dfn>element click intercepted</dfn>
   <td>400
-  <td><code>element intercepting click</code>
+  <td><code>element click intercepted</code>
   <td>An element <a>command</a> could not be completed
     because the element receiving the events is not the
     original element requested in <a>Element Click</a>.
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -1262,6 +1263,15 @@
  </tr>
 
  <tr>
+  <td><dfn>element intercepting click</dfn>
+  <td>400
+  <td><code>element intercepting click</code>
```
shs96c: "element click intercepted". The other errors tend to be in the past tense, not the present tense.

shs96c:
LGTM. Should probably go in as one diff.

AutomatedTester:
landed this already

PR #554, opened 2017-01-18
Add a command to minimize the window

AutomatedTester:
It is likely that users are going to want to minimize a window during
their testing. Adding that here.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/554)


AutomatedTester:
@thejohnjansen I have created a PR for this and will start reviewing now.

AutomatedTester:
fixed and landed in 4850eea

PR #559, opened 2017-01-18
`Element Send Keys`: Ensure that uploaded files actually exist

### LINKED ISSUE #383: Send "invalid argument" error if file does not exist
https://www.w3.org/Bugs/Public/show_bug.cgi?id=28382

Andreas Tolfsen:

> On sending keys to an <input type=file> element we should send back an "invalid argument" error with a helpful message if the file you try to set doesn't exist.  
> 
> As there may be guards against setting invalid/non-existing files in UAs, we should avoid doing so as it may cause the driver to have an internal error.

shs96c:
Closes #383

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/559)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -5116,7 +5116,7 @@ named "<code>firstMatch</code>" from <var>capabilities request</var>.
 
      <li><p>Verify that each file given by the user exists. If any do
       not, return <a>error</a> with <a>error code</a> <a>invalid
-      argument</a> detailing which files were not present.
+      argument</a>.
 
      <li><p>Complete implementation specific steps equivalent to setting the
        <a>selected files</a> on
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -5114,6 +5114,10 @@
       length of <var>files</var> is not equal to 1 return
       an <a>error</a> with <a>error code</a> <a>invalid argument</a>.
 
+     <li><p>Verify that each file given by the user exists. If any do
+      not, return <a>error</a> with <a>error code</a> <a>invalid
+      argument</a> detailing which files were not present.
```
andreastt: Remove “detailing which files were not present”. The `message` field is voluntary.

Issue #410, opened 2016-10-04
Click does not mention if it should/shouldnt wait for page loads if its noticed


andreastt:
David Burns :automatedtester:

> Currently the spec does not mention if we should use and page load strategies when interact with links. Currently the OSS project does a number of different things in these situations

andreastt:
Simon Stewart:
> Clicks are assumed to be asynchronous. Therefore the browser can return whenever it wants to, perhaps before any click handlers have fired or JS being processed. Tests are required to determine when it's safe to proceed.
>
> This is largely because there's no way of telling beforehand whether a click will cause a page load or not.

andreastt:
David Burns :automatedtester:
> *** [Bug 24960](https://www.w3.org/Bugs/Public/show_bug.cgi?id=24960) has been marked as a duplicate of this bug. ***

andreastt:
samuong:
> PageLoadingTest.testShouldTimeoutIfAPageTakesTooLongToLoadAfterClick() requires that the driver waits (and eventually times out) after clicking on a link:
>
> https://github.com/SeleniumHQ/selenium/blob/01399fffecd5a20af6f31aed6cb0043c9c5cde65/java/client/test/org/openqa/selenium/PageLoadingTest.java#L419
>
> Section 6 of the current spec says "For commands that cause a new document to load...return after the load event fires on the new page" (for a normal page loading strategy), but I'm not sure if this applies only to the navigation commands in that section or any command that causes a new document to load:
>
> https://w3c.github.io/webdriver/webdriver-spec.html#navigation
>
> But Simon's also right in that a click is asynchronous and it's not easy to tell if a click handler will navigate to a new page or not. In the worst case, a click handler might do an animation (or something else that takes time) before navigating.
>
> It'd be good if there was a decision on exactly what should happen here.

andreastt:
seva:
> I am for simplicity and against spec requirements that I don't believe that vendors will thoroughly implement. So I +1 Comment 1.
>
> Also https://github.com/joshbruning (from whom we will hear more and more) as well as jleyba and other people have different good ideas in this area. Such as WebDriver exposing hooks that app under test (e.g. web ui frameworks) can use to tell that page has finished loading. Ideas that might make it easy to implement on the WebDriver side and easy for web ui frameworks (or just web apps) to integrate with - making test code free of explicit waits.

andreastt:
seva:
> (Those kinds of things could go in Level 2 and extensions of the standard)

[...]

PR #560, opened 2017-01-18
Ensure we wait for navigation to complete.

shs96c:
Note that this also alters the `Processing Model`.

Address #410.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/560)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -766,6 +766,9 @@ var respecConfig = {
 
   <p>Otherwise, let <var>parameters</var> be null.</p>
 
+ <li><p>If a <a>navigation</a> is currently underway <a>wait for
+  navigation to complete</a>.
+
  <li><p>Let <var>response result</var> be the return value
   obtained by running the <a>remote end steps</a> for <var>command</var>
   with <var>url variables</var> as arguments.
@@ -2872,12 +2875,15 @@ named "<code>firstMatch</code>" from <var>capabilities request</var>.
   or for the <a>session page load timeout</a> milliseconds to pass,
   whichever occurs sooner.
 
-  <p class=issue>This doesn’t actually wait for the page to load
-   if traversing the <a>session history</a> kicks off a navigation.
+  <li><p>If traversing the session history causes <a>navigation</a>:
 
- <li><p>If the previous step completed by the <a>session page load timeout</a> being reached,
-   and the browser is not currently displaying an alert,
-   return <a>error</a> with <a>error code</a> <a>timeout</a>.
+  <ol>
+   <li><p>Run the <a>post-navigation checks</a>
+    and return its value if it is an <a>error</a>.
+
+   <li><p><a>Wait for navigation to complete</a>
+    and return its value if it is an <a>error</a>.
+  </ol>
 
  <li><p>Return <a>success</a> with data null.
 </ol>
@@ -2921,12 +2927,20 @@ named "<code>firstMatch</code>" from <var>capabilities request</var>.
   or for the <a>session page load timeout</a> milliseconds to pass,
   whichever occurs sooner.
 
-  <p class=issue>This doesn't actually wait for the page to load
-   if traversing the session history kicks off a navigation.
+  <li><p>If traversing the session history causes <a>navigation</a>:
 
- <li><p>If the previous step completed by the <a>session page load timeout</a> being reached,
-   and the browser is not currently displaying an alert,
-   return <a>error</a> with <a>error code</a> <a>timeout</a>.
+  <ol>
+   <li><p>Run the <a>post-navigation checks</a>
+    and return its value if it is an <a>error</a>.
+
+   <li><p><a>Wait for navigation to complete</a>
+    and return its value if it is an <a>error</a>.
+  </ol>
+
+ <li><p>If the previous step completed by the <a>session page load
+   timeout</a> being reached, and the browser is not currently
+   displaying an alert, return <a>error</a> with <a>error
+   code</a> <a>timeout</a>.
 
  <li><p>Return <a>success</a> with data null.
 </ol>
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -2850,12 +2853,15 @@
   or for the <a>session page load timeout</a> milliseconds to pass,
   whichever occurs sooner.
 
-  <p class=issue>This doesn’t actually wait for the page to load
-   if traversing the session history kicks off a navigation.
+  <li><p>If traversing the session history causes <a>navigation</a>:
```
andreastt: Navigating a browser’s bfcache does not cause navigation.

webdriver-spec.html
```diff
@@ -2899,12 +2905,20 @@
   or for the <a>session page load timeout</a> milliseconds to pass,
   whichever occurs sooner.
 
-  <p class=issue>This doesn't actually wait for the page to load
-   if traversing the session history kicks off a navigation.
+  <li><p>If traversing the session history causes <a>navigation</a>:
```
andreastt: Navigating a browser’s bfcache forward does not cause navigation.

webdriver-spec.html
```diff
@@ -742,6 +742,9 @@
 
   <p>Otherwise, let <var>parameters</var> be null.</p>
 
+ <li><p>If a <a>navigation</a> is currently underway <a>wait for
+  navigation to complete</a>.
```
andreastt: This only applies to some commands.

shs96c: It applies to all commands. See: [nsCommandProcessor.js](https://github.com/SeleniumHQ/selenium/blob/master/javascript/firefox-driver/js/nsCommandProcessor.js#L528) This unconditionally wraps every incoming command to conform with this.

webdriver-spec.html
shs96c: Suggested alternative wording?

webdriver-spec.html
shs96c: We should be waiting for page loads. Suggested alternative wording?

webdriver-spec.html
jgraham: Is "currently underway" a defined concept for navigation? I don't think it is, so I don't know what this section means except in a handwavy way.

andreastt: I think there is language for it, and you can find an example under [traversing the history by a delta](https://html.spec.whatwg.org/#traverse-the-history-by-a-delta), where this paragraph seems relevant:

> If there is an ongoing attempt to navigate specified browsing context that has not yet _matured_ (i.e. it has not passed the point of making its Document the active document), then cancel that attempt to navigate the browsing context.

And [‘matured’](https://html.spec.whatwg.org/#concept-navigate-mature) seems to be a defined concept too.

So if we want to have this for _every_ command (in the WG we only talked having it for navigation and interaction commands) we ought to change this language to something like

> If there is an ongoing attempt to _navigate_ the _current browsing context_ that has not yet _matured_, _wait for navigation to complete_.

webdriver-spec.html
andreastt: Well if this happens post every command, you don’t need to say it specifically for backwards and forwards.

webdriver-spec.html
jgraham: Is that what's intended here? Block until the navigation is complete but before the new document has started parsing?

shs96c: The intent is to match what we do on "`Go`": use the current `page loading strategy` to determine when to process the next command. This is the (admittedly not clearly documented) behaviour of the Selenium WebDrivers.

webdriver-spec.html
shs96c: True. If we can agree on using the `page load strategy` to determine when to process the next command then I'll remove this text, and the matching one below.

shs96c:
Gentle poke

webdriver-spec.html
```diff
@@ -2872,12 +2882,10 @@
...
   whichever occurs sooner.
 
-  <p class=issue>This doesn’t actually wait for the page to load
-   if traversing the <a>session history</a> kicks off a navigation.
-
- <li><p>If the previous step completed by the <a>session page load timeout</a> being reached,
-   and the browser is not currently displaying an alert,
-   return <a>error</a> with <a>error code</a> <a>timeout</a>.
+ <li><p>If the previous step completed by the <a>session page load
+   timeout</a> being reached, and the browser is not currently
```
AutomatedTester: instead of `and the browser is not currently displaying an alert,` we could put something like `and <a data-lt='Handle any user prompts'>user prompts have been handled</a>`

webdriver-spec.html
```diff
@@ -2921,12 +2929,10 @@
...
 
-  <p class=issue>This doesn't actually wait for the page to load
-   if traversing the session history kicks off a navigation.
-
- <li><p>If the previous step completed by the <a>session page load timeout</a> being reached,
-   and the browser is not currently displaying an alert,
-   return <a>error</a> with <a>error code</a> <a>timeout</a>.
+ <li><p>If the previous step completed by the <a>session page load
+   timeout</a> being reached, and the browser is not currently
+   displaying an alert, return <a>error</a> with <a>error
```
AutomatedTester: same as previous comment

AutomatedTester:
also, squash before merging

webdriver-spec.html

webdriver-spec.html
shs96c: Assume this was for the same block in `Forward`. Done

shs96c:
Squashed and pushed manually. Thanks for the reviews @AutomatedTester and @andreastt

webdriver-spec.html
andreastt: @jgraham If you have further comments on this, I would suggest reading the text on master and raising a new issue before it’s too late.

PR #569, opened 2017-01-19
Decorate CSSOM-VIEW dependencies with their characteristic

vkatsikaros:
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/569)


shs96c:
+1 to what @andreastt said. I don't think this improves readability or removes ambiguity.

AutomatedTester:
I have to agree with @andreastt here. Also there are places where you have used attribute. What type of attribute?

AutomatedTester:
since we have 3 agreements I am going to close. Thanks for helping, it is really appreciated the hard work you are putting in.

vkatsikaros:
No problem, there is an unavoidable learning curve for me :) I am glad my nitpickings are useful and thank you all for taking the time to review them!

May I bother you a tiny bit more on this, just to expose my line of thought and avoid similar cases in the future? I had noticed a few places where this occurs and though I should follow the lead:
* attribute: some attributes mention explicitly that they are attributes (example "type attribute")
* element: many elements explicitly mention that they are elements (example "body element")

PR #574, opened 2017-01-19
Ensure the correct browsing context is checked per-command.

### LINKED ISSUE #397: Specify error behaviour when the current browsing context does not exist
https://www.w3.org/Bugs/Public/show_bug.cgi?id=29691

Alan Jeffrey:

> The webdriver spec should say what to do in the case that the current browsing context no longer exists, for example if it is an iframe that is removed from the current document.
> 
> Some IRC chat on #servo: http://logs.glob.uno/?c=mozilla%23servo&s=10+Jun+2016&e=10+Jun+2016#c453675

shs96c:
Most commands were checking the top-level browsing context
rather than the current browser context. Review each command
and ensure that the correct browser context is checked for
whether it's open.

Closes #397

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/574)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -3009,7 +3009,7 @@ named "<code>firstMatch</code>" from <var>capabilities request</var>.
  by its associated <a>window handle</a>.
  A <a>top-level browsing context</a> is selected
  using the <a>Switch To Window</a> command.
- Once this is done, a specific browsing context can be selected
+ Once this is done, a specific <a>browsing context</a> can be selected
  using the <a>Switch to Frame</a> command.
 
 <p class=note>The use of the term “window” to
@@ -3018,8 +3018,8 @@ named "<code>firstMatch</code>" from <var>capabilities request</var>.
  the operating system notion of a “window”
  or the DOM <a><code>Window</code></a> object.
 
-<p>The <a>top-level browsing context</a> is said
- to be <dfn>no longer open</dfn> if it has been <a>discarded</a>.
+<p>A <a>browsing context</a> is said to be <dfn>no longer open</dfn>
+ if it has been <a>discarded</a>.
 
 <p>Each <a>browsing context</a> has an associated
  <dfn data-lt="window handles">window handle</dfn> (string)
@@ -3944,8 +3944,9 @@ named "<code>firstMatch</code>" from <var>capabilities request</var>.
 <p>The <a>remote end steps</a> are:
 
 <ol>
- <li><p>If the <a>current top-level browsing context</a> is <a>no longer open</a>,
-  return <a>error</a> with <a>error code</a> <a>no such window</a>.
+ <li><p>If the <a>current browsing context</a> is <a>no longer
+  open</a>, return <a>error</a> with <a>error code</a> <a>no such
+  window</a>.
 
  <li><p><a>Handle any user prompts</a> and return its value if it is an <a>error</a>.
 
@@ -4188,6 +4189,10 @@ named "<code>firstMatch</code>" from <var>capabilities request</var>.
 <p>The <a>remote end steps</a> are:
 
 <ol>
+ <li><p>If the <a>current browsing context</a> is <a>no longer
+  open</a>, return <a>error</a> with <a>error code</a> <a>no such
+  window</a>.
+
  <li><p>Let <var>start node</var> be
   the <a>current browsing context</a>’s <a>document element</a>.
 
@@ -4232,6 +4237,10 @@ named "<code>firstMatch</code>" from <var>capabilities request</var>.
  that can be used for future <a>commands</a>.
 
 <ol>
+ <li><p>If the <a>current browsing context</a> is <a>no longer
+  open</a>, return <a>error</a> with <a>error code</a> <a>no such
+  window</a>.
+
  <li><p>Let <var>start node</var> be
   the <a>current browsing context</a>’s <a>document element</a>.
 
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -2843,7 +2843,7 @@
 <p>The <a>remote end steps</a> are:
 
 <ol>
- <li><p>If the <a>current top-level browsing context</a> is <a>no longer open</a>,
+ <li><p>If the <a>current browsing context</a> is <a>no longer open</a>,
```
andreastt: For the _Back_ command, we only need to check the top-level browsing context since it cannot be executed if the window is missing.

webdriver-spec.html
```diff
@@ -2891,7 +2891,7 @@
...
```
andreastt: Same here as for _Back_.

webdriver-spec.html
```diff
@@ -3941,8 +3941,9 @@
 <p>The <a>remote end steps</a> are:
 
 <ol>
- <li><p>If the <a>current top-level browsing context</a> is <a>no longer open</a>,
-  return <a>error</a> with <a>error code</a> <a>no such window</a>.
+ <li><p>If the <a>current browsing context</a> is <a>no longer
+  open</a>, return <a>error</a> with <a>error code</a> <a>no such
+  window</a>.
```
andreastt: Yes, in this case it makes sense to check the current browsing context as opposed to the _top-level_ browsing context: when the user has not switched to a frame, the current browsing context is equal to the top-level browsing context implicitly. (-:

webdriver-spec.html
```diff
@@ -5200,7 +5217,7 @@
 <p>The <a>remote end steps</a> are:
 
 <ol>
- <li><p>If the <a>current top-level browsing context</a> is <a>no longer open</a>,
+ <li><p>If the <a>current browsing context</a> is <a>no longer open</a>,
```
andreastt: Hm, yes. Because a frame’s document is separate from that of the owner document if in an `<iframe>`.

webdriver-spec.html
```diff
@@ -5799,7 +5816,7 @@
...
```
andreastt: I guess cookies are tied to domain, and a frame may be on a different domain than the outer document.

webdriver-spec.html
```diff
@@ -7996,7 +8013,7 @@
...
```
andreastt: No, this needs to be the top-level browsing context since alerts are modal and tied to the browser chrome window.

webdriver-spec.html
```diff
@@ -8029,7 +8046,7 @@
...
```
andreastt: Same.

webdriver-spec.html
```diff
@@ -8062,7 +8079,7 @@
...
```
andreastt: Same.

webdriver-spec.html
```diff
@@ -8097,7 +8114,7 @@
...
```
andreastt: Same.

webdriver-spec.html
shs96c: Surely if the window is missing the `current browsing context` will also have been discarded?

webdriver-spec.html
shs96c: As above.

webdriver-spec.html
shs96c: OK. Changed back.

webdriver-spec.html
shs96c: Changed back.

webdriver-spec.html
shs96c: Changed back

webdriver-spec.html
shs96c: Changed back.

webdriver-spec.html
andreastt: Yes, but if the current browsing context is an `<iframe>` and this element has been removed from the DOM, the top-level browsing context still exists. Since the Back command is one of those commands that operate independently of frames, it should still be possible to call it even if the frame one has switched into has gone missing.

shs96c: Ah. Gotcha. Done.

webdriver-spec.html

webdriver-spec.html
```diff
@@ -4270,6 +4279,10 @@
  that can be used for future <a>commands</a>.
 
 <ol>
```
whimboo: This misses the remote steps paragraph in front of the list.

webdriver-spec.html
```diff
@@ -4315,6 +4328,10 @@
...
```
whimboo: Same here.

webdriver-spec.html
```diff
@@ -8305,7 +8322,7 @@
  the <a>element</a> will not be <a>scrolled into view</a>.
 
 <ol>
```
whimboo: And here too.

webdriver-spec.html
```diff
@@ -3015,8 +3015,8 @@
  the operating system notion of a “window”
  or the DOM <a><code>Window</code></a> object.
 
-<p>The <a>top-level browsing context</a> is said
- to be <dfn>no longer open</dfn> if it has been <a>discarded</a>.
+<p>A <a>browsing context</a> is said to be <dfn>no longer open</dfn>
+ if it has been <a>discarded</a>.
```
andreastt: I wonder if we for browsing contexts (not top-level) should just use the HTML definition _discarded_. I added _no longer open_ for top-levels to sort of indicate that the OS window had been closed, but for _browsing context_ checks in commands, we typically mean “is the DOM the user chose still available? ”, or “is the `<iframe>` still there?”

webdriver-spec.html
andreastt: Not relevant for this patch.

andreastt:
> Just 3 things as I noticed from my PR which are missing here.

I think we should address missing “these are the remote steps for X command” in a separate patch. Please feel free to submit it (-:

PR #575, opened 2017-01-19
`Element Send Keys`: Only focus element if necessary

shs96c:
This allows users to do something like:

```
element.sendKeys("Hello Worlt!");
element.sendKeys(Keys.LEFT);
element.sendKeys(Keys.BACKSPACE);
element.sendKeys("d!");
```

and have that be equivalent to:

```
element.sendKeys("Hello Worlt!" + Keys.LEFT + Keys.BACKSPACE + "d!");
```

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/575)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -5092,12 +5092,18 @@ named "<code>firstMatch</code>" from <var>capabilities request</var>.
 
  <li><p><a>Scroll into view</a> the <var>element</var>.
 
- <li><p>Let <var>current text length</var> be the <a>Node Length</a>.
+ <li><p>If <var>element</var> is not the <a>active element</a>:
 
- <li><p>Set the text insertion caret using <a>set selection range</a>
-   with using <var>current text length</var> as the 2 parameters passed in.
+  <ol>
+   <li><p>Run the <a>focusing steps</a> for the <var>element</var>.
 
- <li><p>Run the <a>focusing steps</a> for the <var>element</var>.
+   <li><p>Let <var>current text length</var> be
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -5092,12 +5092,17 @@
...
- <li><p>Let <var>current text length</var> be the <a>Node Length</a>.
+ <li><p>If <var>element</var> is not the currently <a>active
+  element</a>:
 
- <li><p>Set the text insertion caret using <a>set selection range</a>
-   with using <var>current text length</var> as the 2 parameters passed in.
+  <ol>
+   <li><p>Run the <a>focusing steps</a> for the <var>element</var>.
+
+   <li><p>Let <var>current text length</var> be the <a>Node Length</a>.
```
andreastt: We can fix this later too, but this should really say `be the <var><a>element</a></var>’s <a data-lt="node length">length</a>`.

webdriver-spec.html
```diff
@@ -5092,12 +5092,17 @@
 
  <li><p><a>Scroll into view</a> the <var>element</var>.
 
- <li><p>Let <var>current text length</var> be the <a>Node Length</a>.
+ <li><p>If <var>element</var> is not the currently <a>active
```
andreastt: Is _currently active_ superfluous? I think “if element is not the active element” is enough.

shs96c: Sure

webdriver-spec.html
shs96c: I can fix that now.

PR #576, opened 2017-01-19
Refactor description of element container and fix subtle bugs

andreastt:
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/576)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -250,6 +250,7 @@ var respecConfig = {
    <!-- Button --> <li><dfn><a href="https://html.spec.whatwg.org/#button-state-%28type=button%29">Button</a></dfn> state
    <!-- Buttons --> <li><dfn><a href=https://html.spec.whatwg.org/#concept-button>Buttons</a></dfn>
    <!-- Canvas context mode --> <li><dfn><a href=https://html.spec.whatwg.org/#concept-canvas-context-mode>Canvas context mode</a></dfn>
+   <!-- Multiple --> <li><dfn><a href=https://html.spec.whatwg.org/#attr-input-multiple><code>multiple</code> attribute</a></dfn>
    <!-- Checkbox --> <li><dfn><a href="https://html.spec.whatwg.org/#checkbox-state-%28type=checkbox%29">Checkbox</a></dfn> state
    <!-- Checkedness --> <li><dfn><a href=https://html.spec.whatwg.org/#concept-fe-checked>Checkedness</a></dfn>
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -4376,6 +4379,24 @@
  <li><p>Return false.
 </ol>
 
+<p>An <a>element</a>’s <dfn>container</dfn> is:
+
+<dl class=switch>
+ <dt><a><code>option</code> element</a> in a valid <a>element context</a>
```
shs96c: Also `optgroup`

webdriver-spec.html
```diff
@@ -4882,59 +4903,46 @@
...
+ <li><p><a>Scroll into view</a>
+  the <var><a>element</a></var>’s <a>container</a> <a>element</a>.
 
  <li><p>Wait in an implementation-specific way up to the <a>session
-  implicit wait timeout</a> for <var>container</var> to
+  implicit wait timeout</a> for the <a>container</a> to
   become <a>pointer-interactable</a>.
 
- <li><p>If <var>container</var> is not <a>pointer-interactable</a>,
+ <li><p>If <a>element</a>’s <a>container</a> is not <a>pointer-interactable</a>,
```
shs96c: &lt;var>element&lt;/var>

webdriver-spec.html
```diff
@@ -4882,59 +4903,46 @@
...
-  </dl>
-  </li>
+ <li><p>If the <var>element</var> is
+  an <a><code>input</code> element</a> in the <a>file upload state</a>
+  or it does not have a <a>container</a> <a>element</a>,
+  return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
 
- <li><p><a>Scroll into view</a> the <var>container</var>.
+ <li><p><a>Scroll into view</a>
+  the <var><a>element</a></var>’s <a>container</a> <a>element</a>.
```
shs96c: Also "&lt;var>element&lt;/var>"

andreastt: The last instance here isn’t referring to the variable _element_, but to the concept.

webdriver-spec.html
andreastt: Added match.

webdriver-spec.html
andreastt: Second instance refers to the concept element, and `<var>` is used for the first.

webdriver-spec.html

Issue #346, opened 2016-10-04
Define Element Send Keys


andreastt:

shs96c:
@andreastt I think we've now got a reasonable definition of send keys in place. Ok go close?

shs96c:
I think this is now in reasonable shape, with the remaining open questions and enhancements in their own issues. Closing this one.

PR #578, opened 2017-01-19
`Element Send Keys`: base on top of `Actions`

### LINKED ISSUE #553: `Element Send Keys`: allow resetting keyboard state in command
The Selenium WebDriver APIs offers Keys.NULL (`'\uE000'`) which releases all currently held modifier keys. It's therefore possible to use a command like this:

  element.sendKeys(Keys.SHIFT, "h", Keys.NULL, "ello");

to type "Hello". Without this special character, two `sendKeys` calls would need to be made.

shs96c:
This introduces the ability to unset modifiers using the
`null` key (`\eE000`).

We still leave "`processing keys`" as poorly defined as
before this change, since I think we may need a follow
up diff to handle graphemes and properly define what
"processing keys" actually is.

Closes #553. Addresses #346

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/578)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -5064,6 +5064,34 @@ named "<code>firstMatch</code>" from <var>capabilities request</var>.
  In case the <a>element</a> is not <a>keyboard-interactable</a>,
  an <a>element not interactable</a> <a>error</a> is returned.
 
+<p>The <a>key input state</a> used for
+ input by the command may be cleared mid-way through
+ &quot;typing&quot; by sending the <dfn>null key</dfn>, which
+ is &quot;U+E000 (NULL)&quot;.
+
+<p>When required to <dfn>clear the modifier key state</dfn> with an
+ argument of <var>undo actions</var> and <var>keyboard</var>
+ a <a>remote end</a> must run the following steps:
+
+<ol>
+ <li><p>If <var>keyboard</var> is not a <a>key input source</a> return
+  an <a>error</a> with <a>error code</a> <a>invalid argument</a>.
+
+ <li><p><p>For each <var>action</var> corresponding to an indexed
+  property in <var>undo actions</var>:
+
+ <ol>
+  <li><p>If <var>action</var> is not an <a>action object</a>
+   of <code>type</code> <code>"key"</code>
+   and <code>subtype</code> <code>"keyUp"</code> return
+   an <a>error</a> with <a>error code</a> <a>invalid argument</a>.
+
+  <li><p><a>Dispatch a keyUp action</a> with
+   arguments <var>action</var> and <var>keyboard</var>'s <a>key input
+   state</a>.
+ </ol>
+</ol> <!-- /clear modifier key state -->
+
 <p>The <a>remote end steps</a> are:
 
 <ol>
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -5148,35 +5148,65 @@
...
 
-      <li><p>Generate <a data-lt=keyPress-Event>keyPress</a> <a>Keyboard Event</a>
-        with the relevant <a>keyboard event code</a> for <var>key to type</var>
-        against <var>active element</var>.
-
-      <li><p>Generate <a data-lt=keyUp-Event>keyUp</a> <a>Keyboard Event</a>
-        with the relevant <a>keyboard event code</a> for <var>key to type</var>
-        against <var>active element</var>.
+    <ol>
+     <li><p>Let <var>key to type</var> be the result of <a>processing keys</a> with argument <var>char</var>.
```
AutomatedTester: when I wrote it, and I thought I had defined it, it was mostly there to make sure that we translated from, our PUA to a virtual key

webdriver-spec.html
```diff
@@ -5148,35 +5148,65 @@
...
+
+     <li><p>Let <var>action</var> be an <a>action object</a>
+      constructed with
+      arguments <var>keyboard</var>'s <var>id</var>, <code>"key"</code>,
+      and <code>"keyDown"</code>.
```
shs96c: Renamed the action object. Keeping `keyboard`, but redefined it above.

webdriver-spec.html
```diff
@@ -5148,35 +5148,65 @@
...
+
+     <li><p>Set the <code>value</code> property of <var>action</var> to <var>key to type</var>.
+
+     <li><p><a>Dispatch a keyDown action</a> with
+     arguments <var>action</var> and <var>keyboard</var>.
```
shs96c: The second argument to `keyDown` is an `input state`, not a `device state`. Presumably you're in the middle of a series of changes to that terminology? In either case, `keyboard` has been defined as an `input state` so it should be fine to just pass it in.

webdriver-spec.html
```diff
@@ -5148,35 +5148,65 @@
...
+
+     <li><p>Let <var>action</var> be an <a>action object</a>
+      constructed with
+      arguments <var>keyboard</var>'s <var>id</var>, <code>"key"</code>,
+      and <code>"keyUp"</code>.
```

webdriver-spec.html
```diff
@@ -5148,35 +5148,65 @@
...
+
+     <li><p>Set the <code>value</code> property of <var>action</var>
+      to <var>key to type</var>.
+
+     <li><p>Run the substeps in the first matching statement:
+      <dl>
```
shs96c: Nice catch. Done.

webdriver-spec.html
andreastt: Sorry, the second point I made above is wrong. I wrote that before I realised Element Send Keys doesn’t preserve state.

webdriver-spec.html
```diff
@@ -5068,6 +5068,12 @@
  In case the <a>element</a> is not <a>keyboard-interactable</a>,
  an <a>element not interactable</a> <a>error</a> is returned.
 
+<p>Calling this command effectively clears the <a>current
+ session</a>'s <a>input state</a>. The <a>key input state</a> used for
+ input by the command may be cleared mid-way through
+ &quot;typing&quot; by sending the <dfn>null key</dfn>, which
+ is <code>\uE000</code>.
```
andreastt: “U+0000 (NULL)” and drop `<code>`.

webdriver-spec.html
```diff
@@ -5068,6 +5068,12 @@
  In case the <a>element</a> is not <a>keyboard-interactable</a>,
  an <a>element not interactable</a> <a>error</a> is returned.
 
+<p>Calling this command effectively clears the <a>current
+ session</a>'s <a>input state</a>. The <a>key input state</a> used for
```
andreastt: This isn’t what I meant. We shouldn’t clear any input state when running this command.

What I meant is that each invocation of Element Send Keys uses a new temporary, virtual keyboard device and consequently its state is cleared when the command is done.

webdriver-spec.html
```diff
@@ -5148,35 +5154,64 @@
...
-  <li><p>While <var>i</var> &lt; <var>length</var>:
-
-    <ol>
-      <li><p>Let <var>char</var> be the entry in <var>character array</var> at
-        index <var>i</var>.
-
-      <li><p>Let <var>key to type</var> be the result of <a>processing keys</a>
-        with argument <var>key</var>.
+   <li><p>Remove all <a>key input source</a>s from the <a>current
+    session</a>'s <a>input state</a>.
```
andreastt: See comment above.

webdriver-spec.html
```diff
@@ -5148,35 +5154,64 @@
...
 
-      <li><p>If the <var>key to type</var> is a <a>modifier key</a> then set the relevant
-        attribute on <a>Keyboard Event</a> to true for the rest of this loop.
+   <li><p><a>Create a new key input state
+    object</a>. Let <var>keyboard</var> equal this newly
+    created <a>key input state</a>.
```
andreastt: We don’t have to be explicit about the creating a new key input state object since it’s implicitly created when creating a _key input source_: https://w3c.github.io/webdriver/webdriver-spec.html#dfn-key-input-source

webdriver-spec.html
```diff
@@ -5148,35 +5154,64 @@
...
-        with the relevant <a>keyboard event code</a> for <var>key to type</var>
-        against <var>active element</var>.
+   <li><p>For each <var>grapheme cluster</var> in <var>value</var>:
 
-      <li><p>Generate <a data-lt=keyUp-Event>keyUp</a> <a>Keyboard Event</a>
-        with the relevant <a>keyboard event code</a> for <var>key to type</var>
-        against <var>active element</var>.
+    <ol>
+     <li><p>Let <var>key to type</var> be the result of <a>processing
+      keys</a> with argument <var>char</var>.
```
andreastt: s/char/grapheme cluster/

webdriver-spec.html
```diff
@@ -5132,44 +5137,58 @@
...
-    session</a>'s <a>input cancel list</a>.
+   <li><p>Let <var>keyboard</var> be a new <a>key input source</a>.
 
-  <li><p>Let <var>i</var> be 0.
+   <li><p>Let <var>undo actions</var> be an empty list.
 
-  <li><p>Let <var>length</var> be the length of <var>text</var>.
-
-  <li><p>While <var>i</var> &lt; <var>length</var>:
+   <li><p>For each <var>grapheme cluster</var> in <var>text</var>:
```
andreastt: This also needs an `<a>`. We can define it later.

webdriver-spec.html
```diff
@@ -5132,44 +5137,58 @@
...
-        with argument <var>key</var>.
-
-      <li><p>If the <var>key to type</var> is a <a>modifier key</a> then set the relevant
-        attribute on <a>Keyboard Event</a> to true for the rest of this loop.
-
-      <li><p>Generate <a data-lt=keyDown-Event>keyDown</a> <a>Keyboard Event</a>
-        with the relevant <a>keyboard event code</a> for <var>key to type</var>
-        against <var>active element</var>.
+     <li><p>Let <var>key to type</var> be the result of <a>processing
+      keys</a> with argument <var>grapheme cluster</var>.
```
andreastt: I think maybe this needs to change later. If the intention is that Shift + 2 should produce @, we need to establish some internal state here _across_ grapheme clusters.

(This isn’t an issue.)

webdriver-spec.html
```diff
@@ -5132,44 +5137,58 @@
...
+      </dl>
+    </ol>
 
-      <li><p>Generate <a data-lt=keyPress-Event>keyPress</a> <a>Keyboard Event</a>
-        with the relevant <a>keyboard event code</a> for <var>key to type</var>
-        against <var>active element</var>.
+ <li><p>For each <var>action</var> corresponding to an indexed
+  property in <var>undo actions</var> <a>dispatch a keyUp action</a>
+  with arguments <var>action</var>
+  and <var>keyboard</var>'s <a>key input state</a>.
```
andreastt: This is repeated from above. I suggest we bring this out to a definition called, maybe, **clear the modifier keys state**. If you think it’s better to address this later, we can file an issue.

webdriver-spec.html

webdriver-spec.html

webdriver-spec.html

Issue #508, opened 2017-01-12
Add missing top-level browsing context check to some commands


whimboo:
A couple of sections for commands are missing the the pre-condition for the top-level browsing context to be open:

> If the current top-level browsing context is no longer open, return error with error code no such window. 

Here an example for "Find Element" which will fail without this check:

In chapter [11. Element Retrieval](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-find-element#h-element-retrieval), section [11.2 Find Element](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-find-element#h-find-element)::
> Let start node be the current browsing context’s document element.

whimboo:
I will work on this issue and provide a PR within the next days.

andreastt:
Fixed by https://github.com/w3c/webdriver/pull/574.

PR #579, opened 2017-01-19
Add missing checks for valid top-level browsing context

whimboo:
This adds the missing remote step for checking for a valid top-level browsing context. In some cases we also missed the leading paragraph.

It fixes issue #508.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/579)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -4129,6 +4129,9 @@ named "<code>firstMatch</code>" from <var>capabilities request</var>.
 <p>The <a>remote end steps</a> are:
 
 <ol>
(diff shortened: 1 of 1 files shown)


shs96c:
Does this not do the same thing as #574?

whimboo:
No-one actually pointed out that issue #508 is a dupe, and as such you set it to level-3 3 days ago. Would have been nice to call this out earlier so I wouldn't have wasted my time.

Feel free to close this PR, and check if you caught all missing parts. Thanks.

shs96c:
How frustrating. Sorry for the wasted effort, but thank you for the diff.

PR #580, opened 2017-01-19
Address some JSON-related ReSpec warnings

shs96c:
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/580)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -829,7 +829,7 @@ var respecConfig = {
 
  <li><p>If <var>data</var> is not null, let <var>response</var>’s
   <a>body</a> be a JSON <a>Object</a> with a key <code>value</code>
-  set to the <a>JSON serialisation</a> of <var>data</var>.
+  set to the <a>JSON clone</a> of <var>data</var>.
 
  <li><p>Let <var>response bytes</var> be the byte sequence resulting
   from serializing <var>response</var> according to the rules in [[!RFC7230]].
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -829,7 +829,7 @@
 
  <li><p>If <var>data</var> is not null, let <var>response</var>’s
   <a>body</a> be a JSON <a>Object</a> with a key <code>value</code>
-  set to the <a>JSON serialisation</a> of <var>data</var>.
+  set to the <a>JSON clone</a> of <var>data</var>.
```
andreastt: This should stay the same, I just need to get around to define _JSON serialisation_.

webdriver-spec.html
```diff
@@ -3342,21 +3342,47 @@
...
  and <code>window.outerHeight</code> [[!DOM]] properties.
 
-<p>A <a>top-level browsing context</a>’s <dfn>window position</dfn>
- is defined as a dictionary of the <a>screenX</a> and <a>screenY</a> attributes
- on the <a><code>WindowProxy</code></a> it corresponds to.
- Its <a>JSON representation</a> is the following:
+<p>A <a>top-level browsing context</a>’s <dfn>window position</dfn> is
+ defined as a dictionary of the <a>screenX</a> and <a>screenY</a>
+ attributes on the <a><code>WindowProxy</code></a> it corresponds to.
+ The <dfn>JSON serialization of window position</dfn> is a
```
andreastt: We don’t _need_ a separate definition for each JSON serialisation if we think of it as a trait, but I guess this OK to make the link point to the right location.

webdriver-spec.html
andreastt: What are we trying to do here? I think the data that comes out of each command is JSON conforming already?

AutomatedTester:
I have created #585 which clears up some of the issues I saw in this PR. I did it earlier today but only getting chance to submit the PR now

webdriver-spec.html
shs96c: Possibly, but the clone algorithm handles some edge cases we might miss, and ensures things like collections get handled nicely.

andreastt: Sure, but then we can avoid doing it for _Execute Script_ and _Execute Async Script_?

shs96c: This may not be entirely correct. Needs an edit

andreastt: I’m not super-fussed you need to fix that here. r+ on this from my end.

shs96c:
A lot of these are covered by #585, or made irrelevant thanks to 17d09513c3ae399e3bed8cace24e3a61340bed54. Abandoning.

PR #585, opened 2017-01-19
Json work

AutomatedTester:
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/585)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -834,7 +834,7 @@ var respecConfig = {
 
  <li><p>If <var>data</var> is not null, let <var>response</var>’s
   <a>body</a> be a JSON <a>Object</a> with a key <code>value</code>
-  set to the <a>JSON serialisation</a> of <var>data</var>.
+  set to the <a>JSON Serialization</a> of <var>data</var>.
 
  <li><p>Let <var>response bytes</var> be the byte sequence resulting
   from serializing <var>response</var> according to the rules in [[!RFC7230]].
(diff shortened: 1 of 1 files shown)


shs96c:
Can we break this into the Americanisation of spellings and then a follow up with changes to how serialisation is done?

andreastt:
Also, can we please avoid capitalisation in JSON **s**erialization? The style in the rest of the document is to use a capital letter only if the definition starts a new sentence.

shs96c:
You need to rebase. Once you've addressed @andreastt's comment about capitalisation, ship it!

jgraham:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[webdriver-spec.html, line 674 at r1](https://reviewable.io:443/reviews/w3c/webdriver/585#-KavwUhoW5gXSHDUT7Ke:-KavwUhpPW6Fb69rXggI:b-my3xno) ([raw file](https://github.com/w3c/webdriver/blob/cd788aa6209c80d403abce584978761dd5ce02a7/webdriver-spec.html#L674)):*
> ```HTML
> 
> <p>The result of <dfn>JSON serialization</dfn> with <var>object</var> of type JSON <a>Object</a>
>   and calling the <a>[[\Stringify]]</a> method on <var>object</var>.
> ```

I think you mean `the <a>Stringify</a> function with argument <var>object</var>.`

---

*[webdriver-spec.html, line 677 at r1](https://reviewable.io:443/reviews/w3c/webdriver/585#-KavvokLnEMv4Hl7c5np:-KavvokMU0njKOx0jU3d:b-gogud1) ([raw file](https://github.com/w3c/webdriver/blob/cd788aa6209c80d403abce584978761dd5ce02a7/webdriver-spec.html#L677)):*
> ```HTML
> 
> <p>The result of <dfn>JSON deserialization</dfn> with <var>text</var> is defined as
>   the result of calling the <a>[[\Parse]]</a> method.
> ```

Surely this needs an argument?

Also the function isn't correctly referred to as a "method".

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/webdriver/585)*

AutomatedTester:
landed in 

webdriver/master 3e0666e AutomatedTester: Define the algorithm for JSON Serialization
webdriver/master f121c53 AutomatedTester: Correct spelling of serialization to en-US
webdriver/master 835f2cb AutomatedTester: Define JSON Deserialization

PR #598, opened 2017-01-20
Correct headings in corresponding input state type table

andreastt:
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/598)


webdriver-spec.html
```diff
@@ -6292,30 +6292,34 @@
...
+<table class=simple>
+ <thead>
+  <tr>
+   <th><var>Action type</var>
+   <th>Input state
+  </tr>
+ </thead>
+
+ <tr>
+  <td>"<code>none</code>"
```
shs96c: Why move the quotes outside of the &lt;code> tags?

andreastt: Because it is otherwise unclear if the quotes are part of the string, and it also helps set strings apart from property names, for which we also use `<code>`.

PR #603, opened 2017-01-20
Element Send Keys: Make the `text` parameter a string.

shs96c:
This simplifies the (as yet unwritten) `processing keys`
algorithm, allowing us to break on grapheme clusters as
discussed at the Lisbon F2F.

This also rectifies a historical accident caused when
Selenium WebDriver moved from Java 1.4 to 5 and started
using varargs in the `sendKeys` command.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/603)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -5077,15 +5077,15 @@ named "<code>firstMatch</code>" from <var>capabilities request</var>.
     with using <var>current text length</var> as the 2 parameters passed in.
   </ol>
 
- <li><p>Let <var>character array</var> be the result of <a>getting a property</a>
+ <li><p>Let <var>text</var> be the result of <a>getting a property</a>
     <code>text</code> from the <var>parameters</var> argument.
+
+ <li><p>If <var>text</var> is not a string, return an <a>error</a>
+  with <a>error code</a> <a>invalid argument</a>.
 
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -5128,12 +5128,12 @@
...
 
-  <li><p>Let <var>length</var> be the length of <var>character array</var>.
+  <li><p>Let <var>length</var> be the length of <var>text</var>.
 
   <li><p>While <var>i</var> &lt; <var>length</var>:
 
     <ol>
-      <li><p>Let <var>char</var> be the entry in <var>character array</var> at
+      <li><p>Let <var>char</var> be the entry in <var>text</var> at
         index <var>i</var>.
```
andreastt: This would iterate by character and not code point. I think we need to decide if we want to iterate on code points or grapheme clusters, and right now I tend to lean towards ignoring graphemes for the time being.

Can you change this to iterate without an index, like this?

` <li><p>For each <a><var>code point</var></a> in <var>text</var>:`

shs96c: That change in iteration is done in #578

shs96c: There are known algorithms, provided by the unicode org, for breaking into extended grapheme clusters. I'm planning on basing "processing keys" on that.

andreastt: OK.

Issue #565, opened 2017-01-18
Missing definition for "browser chrome"


vkatsikaros:
"browser chrome" is used in a few places in the spec, but it's not defined.

Kind of related is https://github.com/w3c/webdriver/issues/564

andreastt:
It’s not entirely clear to me we that we _can_ come up with a better definition of that.

vkatsikaros:
Thanks, will add a note.

ltetzlaff:
Just in case somebody stumbles upon this seeking a definition https://www.w3.org/TR/wsc-ui/#def-chrome also defines the term.

andreastt:
@ltetzlaff I like that definition, perhaps we can repurpose it.

andreastt:
Submitted https://github.com/w3c/webdriver/pull/845.

PR #604, opened 2017-01-20
Add non-normative description of 'browser chrome'

vkatsikaros:
Fixes https://github.com/w3c/webdriver/issues/565

Attempt to describe (not define) 'browser chrome' and 'browser chrome elements' in chapter [2. Terminology](https://w3c.github.io/webdriver/webdriver-spec.html#terminology#h-terminology)

Once this is settled I'll link text to it.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/604)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -562,6 +562,14 @@ var respecConfig = {
   defined by the platform for that property i.e. the value it would
   have in the absence of any shadowing by content script.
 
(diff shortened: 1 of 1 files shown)


AutomatedTester:
I think this is good enough to land and if need be we can iterate

andreastt:
I agree, it’s a step in the right direction.

Issue #513, opened 2017-01-16
Dispatch algorithm arguments


vkatsikaros:
In chapter [16. Actions](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-dispatch-a-pointerup-action#h-actions), section [16.4 Dispatching Actions](https://w3c.github.io/webdriver/webdriver-spec.html#h-dispatching-actions) algorithms of dispatch actions are defined.

`dispatch a keyDown action`, `dispatch a keyUp action`: mention only the arguments used in the algorithm:
* action object 
* input state

`dispatch a pause action`, `dispatch a pointerDown action`, `dispatch a pointerUp action`, `dispatch a pointerMove action`, `dispatch a pointerCancel action`: mention all of the possible arguments (regardless whether they are used or not in the algorithm):
* source id
* action object
* input state
* tick duration

Should the definition have a unified way of listing the arguments? 

PS If there is a decision, I could provide the patch.

Edit: fix markdown links to point to published webdriver url

shs96c:
We should pass in the `source id` so that we can correctly create the events for the `input cancel list`. `duration` isn't currently required.

andreastt:
Yes, that’s my interpretation too. But, yes, it would be good to have them all accept the same arguments.

PR #605, opened 2017-01-20
Unify dispatch algorithm arguments

vkatsikaros:
Fixes https://github.com/w3c/webdriver/issues/513

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/605)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -7190,7 +7190,8 @@ named "<code>firstMatch</code>" from <var>capabilities request</var>.
         </table>
 
       <p>When required to <dfn>dispatch a keyDown action</dfn> with arguments
-        <var>action object</var> and <var>input state</var> a <a>remote end</a>
+        <var>source id</var>, <var>action object</var>, <var>input state</var>
+        and <var>tick duration</var> a <a>remote end</a>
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -7190,7 +7190,8 @@
         </table>
 
       <p>When required to <dfn>dispatch a keyDown action</dfn> with arguments
-        <var>action object</var> and <var>input state</var> a <a>remote end</a>
+        <var>source id</var>, <var>action object</var>, <var>input state</var>
+        and <var>tick duration</var> a <a>remote end</a>
```
AutomatedTester: `tick duration` doesn't feel right here as it is just another time of action.

andreastt: If the goal is to synchronise the arguments given to the various dispatch functions, this makes sense. @vkatsikaros?

vkatsikaros: Sorry for the delay, this slipped through the cracks.

Yes, @AutomatedTester the patch's intention is to make uniform the argument of the various dispatch functions. And for most of the other dispatch functions there are 4 arguments, `tick duration` included.

PR #606, opened 2017-01-20
Define Unexpected Alert Behaviour capability. Fixes #402

### LINKED ISSUE #402: Describe how to set the "user prompt handler"
https://www.w3.org/Bugs/Public/show_bug.cgi?id=29778

juangj:

> "A user prompt handler is an enumerated attribute defining what action the remote end must take when a user prompt is encountered. The following table lists the keywords and states for the attribute:"
> 
> How is the user prompt handler defined? Via a capability? The name of the capability should be defined.

AutomatedTester:
This is used to set the value of the user prompt handler
for when encountering a modal dialog.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/606)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -1725,6 +1725,13 @@ with a "-moz-" prefix:
   <td>JSON <a>Object</a>
   <td>Describes the <a>timeouts</a> imposed on certain session operations.
   </tr>
+
+  <tr>
+  <td><dfn>Unexpected Alert Behaviour</dfn>
+  <td>"<code>unexpectedAlertBehaviour</code>"
+  <td>string
+  <td>Describes the <a>current session</a>'s <a>user prompt handler</a>.
+  </tr>
 </tbody>
 </table>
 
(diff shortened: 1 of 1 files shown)


AutomatedTester:
Open questions: 
* Selenium simply has this a boolean, do we change the spec to match?
* Do we allow boolean values and `true` === `accept` and `false` === `dismiss` with `null` meaning there isnt one defined for the session?

andreastt:
Is it really called `unexpectedAlertBehaviour` in any drivers? It would be nice if you could configure the user modal handler by string "accept" (accept state), "dismiss" (dismiss), or null (default boolean attribute missing state) instead of magic variables such as true/false.

shs96c:
A string as the value would be nice.

shs96c:
@andreastt It really is called "[unexpectedAlertBehaviour](https://github.com/SeleniumHQ/selenium/blob/selenium-3.0.1/java/client/src/org/openqa/selenium/remote/CapabilityType.java#L42)"

AutomatedTester:
This is ready to go then if we are going to leave it with the current values

shs96c:
Merging this in, but we should consider using the en-US spelling.

PR #610, opened 2017-01-21
`Element Send Keys`: handle i18n input.

### LINKED ISSUE #506: "processing keys" algorithm
In chapter [13. Element Interaction](https://w3c.github.io/webdriver/webdriver-spec.html#h-element-interaction), section [13.3 Element Send Keys](https://w3c.github.io/webdriver/webdriver-spec.html#h-element-send-keys)::
> the result of processing keys with argument key

respec lib warns that the definition of "processing keys" is missing. I can't find something related in the text, so could it be a missing concept?

shs96c:
We do this by sending composition events. This diff also
resolves the lack of a "processing keys" algorithm and 
handles most common cases when typing a normal string.

There is at least one edge condition (sending a right shift
keyboard input) that will cause unexpected text to be
typed.

Closes #506 and substantially addresses #346.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/610)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -492,6 +492,13 @@ var respecConfig = {
  <dd>The following terms are defined in the  standard: [[!Unicode]]
   <ul>
    <!-- Code point --> <li><dfn data-lt="unicode code point"><a href=http://www.unicode.org/versions/Unicode9.0.0/ch03.pdf#G2212>Code Point</a></dfn>
+   <!-- Extended grapheme cluster --> <li><dfn data-lt="grapheme cluster"><a href=http://www.unicode.org/versions/Unicode9.0.0/ch03.pdf#G2213>Extended Grapheme Cluster</a></dfn>
+  </ul>
+
+ <dt>Unicode Standard Annex #29
+ <dd>The following terms are defined in the standard: [[!UAX29]]
+  <ul>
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -5077,22 +5084,193 @@
...
+   </tr>
+ </table>
+</ul>
+
+<p>When required to <dfn>dispatch actions for a string</dfn> with
+arguments <var>text</var> and <var>keyboard</var>, a <a>remote end</a>
+must run the following steps:
+
+<ol>
+ <li><p>Let <var>clusters</var> be array created by
```
AutomatedTester: an array

AutomatedTester: an array

AutomatedTester: an array

AutomatedTester: an array

webdriver-spec.html
```diff
@@ -5170,8 +5348,10 @@
 
    <li><p>Let <var>keyboard</var> be a new <a>key input source</a>.
 
-   <li><p>Let <var>undo actions</var> be an empty list.
+   <li><p><a>Dispatch actions for a string</a> with
+    arguments <var>text</var> and <var>keyboard</var>.
 
+<!--
```
AutomatedTester: why are we commenting this out now?

webdriver-spec.html
shs96c: Fixed

webdriver-spec.html
shs96c: Because I was referring to it as I wrote the new text. Deleted.

PR #612, opened 2017-01-23
Simplify text and link to html element

vkatsikaros:
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/612)


vkatsikaros:
If the `[HTML]` is needed to differentiate this from (a bit below in the algorithm)
> If element is a [DOM] text node, return true. 

I could add it back. Unless I can also remove `[DOM]` from `[DOM] text node`.

PR #616, opened 2017-01-23
Fix element click

andreastt:
This PR addresses some fundamental misconceptions about pointer-interactability. My changes from earlier today in https://github.com/w3c/webdriver/pull/613 were misleading because I also mis-remembered how it worked.

Pointer-interactability and whether an element is obscured by another is the same basic idea: pointer-interactability assumes the element is inside the viewport, looks at the paint order and checks if the topmost non-transparent at the element’s centrepoint is itself. The obscured element test that was added recently to return a distinctive error message _is exactly the same_ as this check.

This PR changes the behaviour of the _element not interactable_ error to mean that it is outside the viewport. If the element is not pointer-interactable (that is, it is not ‘obscured’), an _element click intercepted_ error is returned instead, preserving existing behaviour.

I’ve implemented this algorithm in Marionette and it seems to work fine.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/616)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -4963,12 +4963,8 @@ named "<code>firstMatch</code>" from <var>capabilities request</var>.
  <li><p>If the <a>current browsing context</a> is <a>no longer open</a>,
   return <a>error</a> with <a>error code</a> <a>no such window</a>.
 
- <li><p>Let <var>element result</var> be the result of <a>getting a known element</a>
-  by <a>UUID</a> reference <var>element id</var>.
-
- <lI><p>Let <var>element</var> be <var>element result</var>’s data
-  if it is a <a>success</a>.
-  Otherwise return <var>element result</var>.
+ <li><p>Let <var>element</var> be the result of <a>trying</a>
+  to <a>get a known element</a> by <a>UUID</a> reference <var>element id</var>.
 
  <li><p>If the <var>element</var> is
   an <a><code>input</code> element</a> in the <a>file upload state</a>
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -469,7 +469,7 @@ var respecConfig = {
   the CSSOM View Module: [[!CSSOM-VIEW]]:
   <ul>
    <!-- elementFromPoint --> <li><dfn>Element from point</dfn> as <a href="https://drafts.csswg.org/cssom-view/#dom-document-elementfrompoint">elementFromPoint()</a>
-   <!-- elementsFromPoint --> <li><dfn>Elements from point</dfn> as <a href=https://drafts.csswg.org/cssom-view/#dom-document-elementsfrompoint>elementsFromPoint()</a>
+   <!-- elementsFromPoint --> <li><dfn data-lt="paint order">Elements from point</dfn> as <a href=https://drafts.csswg.org/cssom-view/#dom-document-elementsfrompoint>elementsFromPoint()</a>
    <!-- getClientRects --> <li><dfn data-lt="DOM client rectangle"><a href=http://www.w3.org/TR/cssom-view/#dom-range-getclientrects>getClientRects</a></dfn>
    <!-- innerHeight --> <li><dfn><a href=https://drafts.csswg.org/cssom-view/#dom-window-innerheight>innerHeight</a></dfn>
    <!-- innerWidth --> <li><dfn><a href=https://drafts.csswg.org/cssom-view/#dom-window-innerwidth>innerWidth</a></dfn>
@@ -1385,7 +1385,7 @@ var respecConfig = {
   <td><code>element click intercepted</code>
   <td>The <a>Element Click</a> <a>command</a> could not be completed
    because the <a>element</a> receiving the events
-   is not the same element that was requested clicked.
+   is <a>obscuring</a> the element that was requested clicked.
  </tr>
 
  <tr>
@@ -3889,7 +3889,7 @@ named "<code>firstMatch</code>" from <var>capabilities request</var>.
  keyboard interactions must be targeted to
  its <a><code>ownerDocument</code></a>.
 
-<p>An <a>element</a>’s <dfn>in-view centerpoint</dfn>
+<p>An <a>element</a>’s <dfn data-lt=centerpoint>in-view centerpoint</dfn>
  is the origin position of the rectangle
  that is the intersection between
  the element’s first <a>DOM client rectangle</a>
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -3814,7 +3814,7 @@
 
 <p>A <dfn data-lt="pointer-interactable">pointer-interactable element</dfn>
  is defined to be the first non-transparent <a>element</a>,
- defined by the paint order found at the centre point
+ defined by the paint order found at the centerpoint
```
shs96c: Link to definition of `paint order`

shs96c: There's a space in `centre point` --- I think this fix must have gone in while you were editing.

webdriver-spec.html
```diff
@@ -3825,13 +3825,13 @@
...
  its <a><code>ownerDocument</code></a>.
 
-<p>An <a>element</a>’s <dfn>in-view centre point</dfn>
- is the centre point of the area
- of the first <a data-lt="getClientRects">DOM client rectangle</a>
- that is inside the <a>viewport</a>.
- It can be calculated this way,
- given an element, here called <var>rectangle</var>,
- off a <a><code>DOMRect</code></a> sequence:
+<p>An <a>element</a>’s <dfn data-lt=centerpoint>in-view centerpoint</dfn>
```
shs96c: `center point` is two words

webdriver-spec.html
```diff
@@ -3863,7 +3863,15 @@
...
 </ol>
 
-<p>In order to produce a <dfn>pointer-interactable elements tree</dfn>
+<p>An <a>element</a> is <dfn>in view</dfn>
+ if it is a member of its own <a>pointer-interactable paint tree</a>.
+
+<p>An <a>element</a> is <dfn data-lt=obscuring>obscured</dfn>
+ if it is not <a>pointer-interactable</a>
+ and the <a>first pointer-interactable element</a>
+ at its <a>centerpoint</a> is not itself.
```
shs96c: `center point`: df38ce9803b668bd8b30870ae41ff13783a4e130

webdriver-spec.html
```diff
@@ -3863,7 +3863,15 @@
  <li><p>Return <var>x</var> and <var>y</var> as a pair.
 </ol>
 
-<p>In order to produce a <dfn>pointer-interactable elements tree</dfn>
+<p>An <a>element</a> is <dfn>in view</dfn>
+ if it is a member of its own <a>pointer-interactable paint tree</a>.
+
+<p>An <a>element</a> is <dfn data-lt=obscuring>obscured</dfn>
+ if it is not <a>pointer-interactable</a>
```
shs96c: Surely if it _is_ pointer-interactable?

webdriver-spec.html
```diff
@@ -3882,16 +3890,16 @@
  <li><p>If <var>rectangles</var> has the length of 0,
   return an empty sequence.
 
- <li><p>Let <var>centre point</var> be
-  the <a>in-view centre point</a> of <var>element</var>
+ <li><p>Let <var>center point</var> be
+  the <a>in-view centerpoint</a> of <var>element</var>
```
shs96c: `center point`

webdriver-spec.html
```diff
@@ -4880,8 +4888,12 @@
 
 <p>The <dfn>Element Click</dfn> <a>command</a>
  <a>scrolls into view</a> the <a>element</a>
- and clicks the <a>in-view centre point</a>.
- If the <a>element</a> is not <a>pointer-interactable</a>,
+ if it is not already <a>pointer-interactable</a>,
+ and clicks its <a>in-view centerpoint</a>.
+ If the element’s <a>centerpoint</a>
```
shs96c: `center point`

webdriver-spec.html
```diff
@@ -4880,8 +4888,12 @@
 
 <p>The <dfn>Element Click</dfn> <a>command</a>
  <a>scrolls into view</a> the <a>element</a>
- and clicks the <a>in-view centre point</a>.
- If the <a>element</a> is not <a>pointer-interactable</a>,
+ if it is not already <a>pointer-interactable</a>,
+ and clicks its <a>in-view centerpoint</a>.
```
shs96c: `center point`

webdriver-spec.html
```diff
@@ -4880,8 +4888,12 @@
...
+ If the element’s <a>centerpoint</a>
+ is <a>obscured</a> by another element,
+ an <a>element click intercepted</a> <a>error</a> is returned.
+ If the element is outside the <a>viewport</a>,
```
shs96c: Or if it's not `pointer-interactable`?

webdriver-spec.html
```diff
@@ -4890,36 +4902,23 @@
...
- <li><p>If <a>element from point</a> for
-  the <a>in-view centre point</a> of <var>element</var>
-  is not equal to the <var>element</var>’s <a>container</a>,
-  return <a>error</a> with <a>error code</a> <a>element click intercepted</a>.
+ <li><p><a>Scroll into view</a> <var>element</var>’s <a>container</a>
+  if it is not <a>pointer-interactable</a>.
 
- <li><p>If <var>element</var>’s <a>container</a>
-  is not <a>pointer-interactable</a>,
+ <li><p>If <var>element</var>’s <a>container</a> is still not <a>in view</a>,
```
shs96c: Add the wait until the session implicit wait timeout here. It appears to have gone AWOL with this change.

webdriver-spec.html
```diff
@@ -7647,7 +7646,7 @@
 
                   <li><p>Let <var>x element</var> and <var>y
                   element</var> be the result of calculating the
-                    <a>in-view centre point</a>
+                    <a>in-view centerpoint</a>
```
shs96c: `center point`

andreastt: Fixed

webdriver-spec.html
andreastt: So I intentionally left this out because I didn’t understand what it did. Is the intention to make the command synchronous by waiting for it to finish scrolling before returning?

If so, I’m not sure we can wait for the full length of the duration since if the element is clearly still obscured after scrolling, we will just sit idle doing nothing.

Anyway, I can add it back in and raise a separate issue on this and we can discuss it there.

webdriver-spec.html
andreastt: Fixed

webdriver-spec.html
andreastt: Fixed

webdriver-spec.html
andreastt: Fixed

webdriver-spec.html
andreastt: Fixed

webdriver-spec.html
andreastt: Fixed

webdriver-spec.html
andreastt: Well “centerpoint” is actually fairly established, but I’m very happy to continue using two words.

webdriver-spec.html
andreastt: An element that is below another is not hittable and thus _not_ pointer-interactable. It is the other element _obscuring it_ that is pointer-interactable. This provides a definition for the first, underlying element.

andreastt:
All issues addressed.

PR #623, opened 2017-01-24
Avoid capitalisation of definitions in Nodes chapter

andreastt:
It’s worth pointing out that ReSpec takes care of this and lower-cases them when viewing the HTML document in a browser. However, it’s worth having the same consistency before us when we edit the document.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/623)


shs96c:
Looks like this needs to be rebased. Fine to land.

PR #630, opened 2017-01-24
Wording and typography fixes

andreastt:
Please don’t squash, since these changes are meant to be atomic.

Also, to be clear, there are no functional changes in this PR.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/630)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -884,8 +884,6 @@ var respecConfig = {
   containing a stack trace report of the active stack frames
   at the time when the error occurred.
 
- <li><p class=issue>Really need a better way to construct JSON literals.
-
   <p>Let <var>data</var> be a new JSON <a>Object</a>
    initialised with the following properties:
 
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -1332,9 +1339,9 @@
 <section>
 <h3>Handling Errors</h3>
 
-<p>Errors are represented in the WebDriver protocol
- with a <a>HTTP response</a> with a <a>HTTP status</a> in the 4xx or 5xx range,
- and a JSON body containing details of the error.
+<p><a>Errors</a> are represented in the WebDriver protocol
+ by an <a>HTTP response</a> with an <a>HTTP status</a> in the 4xx- or 5xx range,
```
shs96c: Nit: Not sure how that `-` creeped in. We can fix in a second pass.

andreastt: Oops.

PR #667, opened 2017-01-25
Specify argument to "getting a known element"

jugglinmike:
I think we could do a bit more cleanup here while we're at it. It seems as though the words "reference" and "parameter" that follow most current usages of "getting a known element by UUID" should be removed. Is that right?

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/667)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -4333,7 +4333,7 @@ with a "<code>moz:</code>" prefix:
   window</a>.
 
  <li><p>Let <var>start node</var> be the result of
-  <a>getting a known element</a> by <a>UUID</a> <var>reference</var>.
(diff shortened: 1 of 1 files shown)


andreastt:
This change in itself looks OK, but generally “parameters” are URL template fields passed as part of the request.

jugglinmike:
That word seems a bit overloaded at the moment. What do you think about defining a term like "URL parameter" for these specific cases (in separate patch, of course)?

andreastt:
Let’s use terminology from [RFC 6570](https://tools.ietf.org/html/rfc6570) which defines URL templates. Filed https://github.com/w3c/webdriver/issues/673. Please feel free to take it.

jugglinmike:
Great idea!

Issue #665, opened 2017-01-25
Drop introduction under User Prompt chapter


andreastt:
In chapter [18. User Prompts](https://w3c.github.io/webdriver/webdriver-spec.html#user-prompts#h-user-prompts), section [18.1 Introduction](https://w3c.github.io/webdriver/webdriver-spec.html#user-prompts#h-introduction):
> 18.1 Introduction

andreastt:
The chapter is empty.

PR #670, opened 2017-01-25
Drop empty introduction chapter under User Prompt

andreastt:
Fixes: https://github.com/w3c/webdriver/issues/665

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/670)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -8024,8 +8024,6 @@ must run the following steps:
 <section>
 <h2>User Prompts</h2>
 
-<section><!-- Introduction -->
-  <h2>Introduction</h2>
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -8071,8 +8071,6 @@
 <section>
 <h2>User Prompts</h2>
 
-<section><!-- Introduction -->
-  <h2>Introduction</h2>
 <p><i>This section is non-normative</i>
```
shs96c: This section contains normative language.

andreastt: I think this paragraph can just be dropped, actually. We don’t have it for all the other sections.

PR #674, opened 2017-01-25
Use trying macro throughout

andreastt:
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/674)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -5347,12 +5347,9 @@ must run the following steps:
 
  <li><p><a>Handle any user prompts</a>, and return its value if it is an <a>error</a>.
 
- <li><p>Let <var>element result</var> be the result of <a>getting a known element</a>
+ <li><p>Let <var>element</var> be the result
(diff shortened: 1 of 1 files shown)


shs96c:
@andreastt, this needs to be rebased on the current master.

webdriver-spec.html
```diff
@@ -3342,14 +3344,9 @@
...
-      <a>getting a known element</a> by
-      <a>UUID</a> reference <var>id</var>.
-
-     <li><p>If <var>element result</var> is a <a>success</a>,
-      let <var>element</var> be <var>element result</var>’s data.
-
-      <p>Otherwise, return <var>element result</var>.
+     <li><p>Let <var>element</var> be the result
+      of <a>trying</a> to <a>get a known element</a>
+      by <a>UUID</a> reference <var>id</var>.
```
shs96c: There is no requirement that a remote end use a UUID as the reference to an element. Raised #681

PR #677, opened 2017-01-26
Correct socksProxyPort to socksProxy

vkatsikaros:
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/677)


shs96c:
The language in this table is confusing.

andreastt:
The table is also wrong. We shouldn’t be using normative language in a reference table.

PR #682, opened 2017-01-27
An element reference should be a UUID

### LINKED ISSUE #681: Element references need not be UUID
It's probably best if an element reference is a UUID, but the only real constraint is that the reference is unique within a session, and if the same element is returned, the same element reference is also returned. 

For example, an increasing count for the lifetime of a session would also be appropriate.

shs96c:
But it may not be. Implementations need only ensure that the
reference is unique within the current session in order for
the protocol to work.

This allows implementors more latitude in how to create the
reference. The original IEDriver used a serialisation of the
COM object's address, and that would meet the constraints
required here too.

Closes #681

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/682)


AutomatedTester:
This needs a manual rebase

Issue #627, opened 2017-01-24
Indistinguishable from a remote end


vkatsikaros:
In chapter [5. Nodes](https://w3c.github.io/webdriver/webdriver-spec.html#terminology#h-nodes):

>  Intermediary nodes must be black-box indistinguishable from a remote end from the point of view of local end

>  An endpoint node must be, like intermediary nodes, indistinguishable from a remote end. 

Does the wording in the different way the two node types are `indistinguishable from a remote end` mean they are somehow different?

andreastt:
I think probably the last quote is a bit confusing.

shs96c:
The spec should make clear that the following is true:

```
local-end -> intermediary -> intermediary -> end-point
             \_______________________________________/
                      all kinds of remote-ends
```

There need not be any intermediary nodes, and so to a local end, both intermediary nodes and end point nodes appear identical (in the same way that a browser can't necessarily tell if it's talking to a proxy or the actual HTTP server)

shs96c:
I think this is okay now. Closing, but please reopen if you disagree.

PR #685, opened 2017-01-27
Attempt to address #627

vkatsikaros:
#627

This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/685)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -676,18 +676,18 @@ var respecConfig = {
  <dt><dfn data-lt="intermediary nodes">Intermediary node</dfn>
  <dd>Intermediary nodes are those that act as proxies,
   implementing both the client and server sides of the <a href=#protocol>protocol</a>.
-  Intermediary nodes must be black-box indistinguishable from a <a>remote end</a>
-  from the point of view of <a>local end</a>,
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -676,19 +676,21 @@
...
  <dd>An endpoint node is the final <a>remote end</a>
   in a chain of nodes that is not an <a>intermediary node</a>.
   The endpoint node is implemented by a user agent or a similar program.
-  An endpoint node must be, like <a>intermediary nodes</a>,
-  indistinguishable from a <a>remote end</a>.
 </dl>
 
+<p> Both <a data-lt="intermediary nodes">intermediary</a> and
+  <a data-lt="endpoint node">endpoint</a> nodes must be black-box
+  indistinguishable from a <a>remote end</a>,
```
shs96c: Because both `intermediary` and `endpoint` nodes are types of `node types`, it's probably enough to say:

```all remote end node types must be black-box indistinguishable from the point of view of local ends, and so...```

This formulation allows to add additional node types if we ever go crazy and do that.

vkatsikaros: Done.

Squashed and force pushed

Issue #691, opened 2017-01-27
Term "input state"


vkatsikaros:
In chapter [17. Actions](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-input-state#h-actions), section [17.2 Input State](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-input-state#h-input-state):
> Each session has an associated input state table. This is a map between input id and the device state

I apologize if some of these points sound unclear to me, due to the lack of understanding some of the concepts involved:
* it is referenced as `object` in:
> As with real devices, virtual devices are stateful; this state is recorded in an input device state object associated with each device. 

Some related terms are also refered to as objects "null input state", "key input state".

* I am not sure I understand the definition: 
> This is a map between "input id" and the "device state" for that input source... 

as "device state" refers back to the "input state" table ie the map.

shs96c:
Each of `key`, `pointer`, and `null` input sources have a related state. `device state` is the general term to refer to these in the general sense (in OO, I'd say that `key input state`, `pointer input state` and `null input state` "*is a*" `device state`

I think we're just seeing some of the evolution of this part of the spec poking through. We could do with refining the language a little more.

It seems like everyone is happy with `key input source`, `pointer input source` and `null input source`, and these can be referred to in the general sense as types of `input source`. Each `input source` has an associated `id`, and some persistent state (`device state` in this text, but probably we'd be okay with `source state` too)

The `input state table` is a mapping between `input source` IDs and `input source` instances. It's used to provide an efficient way of finding the right `input source`, but you could imagine a different data structure being used to hold the information --- all we need to be able to do is find the right `device state` given a particular `input source` ID.

Does that help clear things up?

vkatsikaros:
Thank you for the clarifications!

If I understood you comment correctly, `input state table` and `device state` seem intertwined but shouldn't, and the reasons are:
* `key input state`, `pointer input state` and `null input state` are types of  `device state`, but this isn't mentioned somewhere
* at the same time `device state` links to `input state table` which is incorrect.
* and the definition of term `input state table` is leaving "table" out of the definition.

shs96c:
I think this is now done. Closing.

PR #692, opened 2017-01-30
Disentangle "input state" and "device state"

vkatsikaros:
Fixes https://github.com/w3c/webdriver/issues/691

* "key input state", "pointer input state" and "null input state" are
  types of "device state", but this isn't mentioned somewhere
* at the same time "device state" links to "input state table" which is incorrect.
* the definition of term "input state table" is leaving "table" out
  of the definition.

If something is wrong with the above concepts we could continue the discussion in https://github.com/w3c/webdriver/issues/691

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/692)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -2311,7 +2311,7 @@ with a "<code>moz:</code>" prefix:
 
 <p>A <a>session</a> has an associated <a>user prompt handler</a>.
 
-<p>A <a>session</a> has an associated <a>input state</a>.
+<p>A <a>session</a> has an associated <a>input state table</a>.
 
 <p>A <a>session</a> has an associated <a>input cancel list</a>.
 
@@ -5380,7 +5380,7 @@ must run the following steps:
     arguments <var>text</var> and <var>keyboard</var>.
 
  <li><p>Remove <var>keyboard</var> from the <a>current
-  session</a>'s <a>input state</a>
+  session</a>'s <a>input state table</a>
 
  <li><p>Return <a>success</a> with data null.
 </ol>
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -6327,7 +6327,7 @@
  and a <dfn>source type</dfn> which determines
  the kind of input the device can provide.
  As with real devices, virtual devices are stateful;
- this state is recorded in an <a data-lt="input state">input device state</a> object
+ this state is recorded in an <a>input state table</a>
```
shs96c: The state of the virtual devices is held by the input source, and these are what is held in the input state table. 

This might be the best place to define what "device state" is: a general term for the persistent state held by each input source.

shs96c: Or just link here to the later definition.

vkatsikaros: Done

shs96c:
The more I review this, the more I think we've gone down the wrong path. The table is really a dictionary and is used for efficient look up. As such, it's an implementation detail. Each session really has a set of input sources, and each input source has an ID, source type, and "state", which can currently be one of the three types.

I don't think it's fair to keep asking for tweaks to this PR when what we need to do is tidy up the thinking and language quite a lot. Thank you for pushing on this, and helping to clarify what needs to be done. It's very much appreciated.

I'm going to close this PR and fashion a new one with our better understanding in it. Would love your input on it.

vkatsikaros:
@shs96c happy to help in any way! Mention me on the new PR and I'll take a look.

PR #694, opened 2017-01-30
`Find` algorithm: ensure all elements are actually elements.

### LINKED ISSUE #693: `Find` algorithm must check elements are actually elements
And not text nodes, processing instructions or comments. As expressed now, the xpath location strategy can return anything the xpath returns.

shs96c:
There was no check to ensure this. Also fixes a problem
where the link `JSON Serialization` exactly matched a
definition, and so linked to the wrong place.

Closes #693

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/694)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -4073,13 +4073,9 @@ with a "<code>moz:</code>" prefix:
 
  <li><p>Let <var>result</var> be an empty JSON <a>List</a>.
 
- <li><p>For each <var>element</var> in <var>elements returned</var>:
-
-  <ul>
-   <li><p>Append to <var>result</var>
-    the <a data-lt="JSON Serialization of an element">JSON Serialization</a>
-    of the <var>element</var>.
-  </ul>
+ <li><p>For each <var>element</var> in <var>elements returned</var>,
+  append the <a data-lt="JSON serialization of an element">serialization</a>
+  of <var>element</var> to <var>result</var>.
 
  <li><p>Return <var>result</var>.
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -4064,13 +4064,12 @@
...
  <li><p>Let <var>result</var> be an empty JSON <a>List</a>.
 
- <li><p>For each <var>element</var> in <var>elements returned</var>:
-
-  <ul>
-   <li><p>Append to <var>result</var>
-    the <a data-lt="JSON Serialization of an element">JSON Serialization</a>
-    of the <var>element</var>.
-  </ul>
+ <li><p>For each <var>element</var> in <var>elements returned</var>,
```
AutomatedTester: By this point `elements returned` will be a `NodeList` so we don't need to any other error checks. we just need to append. What situations would not return a `NodeList`?

shs96c: If you run an xpath that returns attributes or text nodes, then the nodes won't be elements

AutomatedTester: we should then make sure that [`XPath`](https://w3c.github.io/webdriver/webdriver-spec.html#xpath) returns a `NodeList` to find. We have already done that for `link text`

shs96c: We could, or do the check in the `Find` algorithm, remove the text from `link text` and make it so that if we add something new (there's an open issue for finding by plain text) we don't need to remember to add this check.

PR #695, opened 2017-01-31
Add dependency for ECMAScript internal method and links to it

vkatsikaros:
TBH I wasn't aware what an ECMAScript "internal method" method is. This patch makes more explicit that it's something well defined somewhere.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/695)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -194,7 +194,8 @@ var respecConfig = {
   </ul>
 
  <dd>This specification also presumes that you are able to call
-  some of the internal methods from the ECMAScript Language Specification:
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -757,11 +758,11 @@
 <p>The result of <dfn data-lt="getting properties">getting a property</dfn>
  with <var>name</var> from an <a>Object</a> is defined as
  the result of calling the <a>[[\GetOwnProperty]]</a>
- internal method of with property name <var>name</var>.
+ <a>internal method</a> of with property name <var>name</var>.
```
shs96c: Unrelated nit: missing word in this sentence. Probably `that Object`

PR #697, opened 2017-01-31
Make web element references normative

andreastt:
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/697)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -2308,8 +2308,8 @@ with a "<code>moz:</code>" prefix:
  but must be exactly one for a <a>remote end</a> that is an <a>endpoint node</a>.
 
 <p>A <a>session</a> has an associated <dfn>session ID</dfn>
- (a <a>UUID</a>) used to uniquely identify this session.  Unless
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -3753,19 +3753,14 @@
  with the string "<code>element-6066-11e4-a52e-4f735466cecf</code>".
 
 <p>Each <a>element</a> has an associated <dfn>web element
- reference</dfn> that uniquely identifies the <a>element</a> across
- all <a>browsing contexts</a>.  The <a>web element reference</a> for
- every <a>element</a> representing the same <a>element</a> is the
- same. The <a>web element reference</a> should be a <a>UUID</a>.
+ reference</dfn> (a <a>UUID</a>)
```
shs96c: There's no requirement for the reference to be a UUID. See 81bd8355e6c509c8349f2e7302960148b5a95703. Use `should` to make this clear

andreastt: OK, fixed.

PR #702, opened 2017-02-01
Update unexpectedAlertBehaviour to unhandledPrompt handler

AutomatedTester:
This updates the name of the capability and adds details to
processing capabilities on how to use it.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/702)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -1804,8 +1804,8 @@ with a "<code>moz:</code>" prefix:
  </tr>
 
  <tr>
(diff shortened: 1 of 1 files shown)


shs96c:
Guess the naming makes more sense given that it's handling `user prompts`.

andreastt:
If we are renaming it away from the capability that FirefoxDriver is already using, we should ensure to use the US spelling of behaviour.

PR #704, opened 2017-02-02
Update behaviour to en-US

AutomatedTester:
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/704)


webdriver-spec.html
```diff
@@ -1804,8 +1804,8 @@
  </tr>
 
  <tr>
-  <td><dfn>Unhandled Prompt Behaviour</dfn>
-  <td>"<code>unhandledPromptBehaviour</code>"
+  <td><dfn>Unhandled Prompt behavior</dfn>
```
shs96c: Nit: change in capitalisation

PR #709, opened 2017-02-02
Define `input source state` and `input state table` and use them

shs96c:
This clears up the current respec warnings in the spec,
but it's a little clumsy --- the `input state table` is not
really necessary.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/709)


AutomatedTester:
this needs a rebase and then can land

PR #714, opened 2017-02-03
`Get Element Text`: use the ECMAScript atom

### LINKED ISSUE #413: Proposal for Get Element Text
https://www.w3.org/Bugs/Public/show_bug.cgi?id=29530

clmartin:

> "12.5 getElementText()"
> 
> John Jansen and I were reading the specification for getElementText() and he remembers discussing it at a face to face but can't find the minutes. Below is what we think was agreed upon:
> 
> Get Element Text
> 
> GET /session/{session id}/element/{element id}/text
> 
> The Get Element Text command retrieves the textContent value of the given web element.
> 1. If the current top-level browsing context is no longer open, return error with error code no such window.
> 2. Handle any user prompts and return its value if it is an error.
> 3. Let element result be the result of getting a known element by UUID parameter element id.
> 4. If element result is a success, let element be element result's data. Otherwise, return element result.
> 5. If element is stale, return error with error code stale element reference.
> 6. Let element text be the result of calling <a href="https://www.w3.org/TR/2004/REC-DOM-Level-3-Core-20040407/core.html#Node3-textContent">textContent</a> on the specified element.
> 7. Let body be a JSON Object with the "value" memeber set to the element text.
> 8. Return success with data body.
> 
> What do you think?

shs96c:
The atom here is aware of the ShadowDOM and so should produce
better results if that's used. It's functionally equivalent
to the version used in Selenium 3.1

Closes #413

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/714)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -437,6 +437,7 @@ var respecConfig = {
   revision <code>1721e627e3b5ab90a06e82df1b088a33a8d11c20</code>.
   <ul>
    <!-- bot.dom.getVisibleText --> <li><dfn><a href="https://github.com/SeleniumHQ/selenium/blob/1721e627e3b5ab90a06e82df1b088a33a8d11c20/javascript/atoms/dom.js#L979"><code>bot.dom.getVisibleText</code></a></dfn>
+   <!-- bot.dom.isShown --> <li><dfn><a href="https://github.com/SeleniumHQ/selenium/blob/1721e627e3b5ab90a06e82df1b088a33a8d11c20/javascript/atoms/dom.js#L548"><code>bot.dom.isShown</code></a></dfn>
   </ul>
 
  <dt>Styling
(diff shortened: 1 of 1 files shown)


shs96c:
There are two distinct changes here, but they are both atoms from the selenium project.

webdriver-spec.html
```diff
@@ -4715,43 +4715,30 @@
...
 
- <li><p>If <var>element</var> <a>is stale</a>,
-  return <a>error</a> with <a>error code</a> <a>stale element reference</a>.
-
- <li><p>Let <var>rendered text</var> be
-  the result of <a>getting <code>innerText</code></a>
-  of <var>element</var>.
+ <li><p>Invoke the <a>[[\Call]]</a> <a>internal method</a> of
+  the <a>get text function</a>, providing <code>null</code> as the
+  this value and <var>element</var> as the argument values. If doing
```
andreastt: `<code>this</code>`

webdriver-spec.html
```diff
@@ -4715,43 +4715,30 @@
...
+  so does not produce an exception, let <var>rendered text</var> be
```
andreastt: s/exception/error/

webdriver-spec.html
```diff
@@ -8399,6 +8386,80 @@
...
+ Element Text</a> command should work. As such, the function presented
+ here is known to be flawed, but provides the best compatibility with
+ existing users.
+
+<p>The function presented here is
+ the <a href=""><code>bot.dom.getVisibleText</code></a> function
+ compiled with
+ Google's <a href="https://developers.google.com/closure/compiler/">Closure
+ Compiler</a> with advanced optimizations enabled. The code may be
+ recreated by cloning the selenium project, checking out
```
andreastt: s/selenium/Selenium/

webdriver-spec.html
```diff
@@ -8399,6 +8386,80 @@
...
+ revision <a href="https://github.com/SeleniumHQ/selenium/commit/c1d39c66115053f71571b9f46d21028bd57ec83f">c1d39c6611</a>
+ and running <code>go //javascript/atoms/fragments:get-text</code>.
+
+<p>All code is available under the Apache 2.0 license, and features
+ code copyright Google and the Software Freedom Conservancy.
```
andreastt: This is an interesting point. Are we actually allowed to include licensed code as normative in a spec?

webdriver-spec.html
```diff
@@ -8474,196 +8535,69 @@
...
-      is less than (<var>parent dimension</var>'s <code>right</code>
-      − the <a>normalized style pixel float value</a>
-      of <var>parent style</var>'s <code>borderRightWidth</code> casted as a float).
-
-     <li><p><var>element dimension</var>'s <code>right</code>
-      is less than (<var>parent dimension</var>'s <code>left</code>
-      − the <a>normalized style pixel float value</a>
-      of <var>parent style</var>'s <code>borderLeftWidth</code> casted as a float).
-   </ul>
+<div class=note>
```
andreastt: This seems to be repeated earlier.

webdriver-spec.html

webdriver-spec.html

webdriver-spec.html

webdriver-spec.html
shs96c: Started an email thread to discuss this with @plehegar and Mike[tm] Smith

webdriver-spec.html
shs96c: Same constraints and instructions on how to get the atom, but the actual function has changed. Applied comments above to here too.

shs96c:
Assigning to @plehegar as we wait for an answer to the licensing question. I can move this forward after that.

webdriver-spec.html
thejohnjansen: My thinking is that HTML5.1 does this with a bunch of normative links; the high-order bit should be whether that link is stable or if the content is going to be changing. I feel like it's pretty concrete, and if so it seems OK to me (of course, verifying with Mike[tm] Smith makes a ton of sense as well).

andreastt: @shs96c Was this resolved?

andreastt:
Alas, this now also needs a rebase.

webdriver-spec.html
```diff
@@ -8402,6 +8390,80 @@
 </section> <!-- /Take Element Screenshot -->
 </section> <!-- /Screen Capture -->
 
+<section>
+<h2>Get Text Function</h2>
+
+<div class=note>
+<p>One of the major inputs to this specification was the Open
+ Source <a href="http://seleniumhq.org">Selenium</a> WebDriver
+ project. This was in wide-spread use before this specification
```
andreastt: I think we should s/WebDriver// here to avoid confusion.

webdriver-spec.html
```diff
@@ -8402,6 +8390,80 @@
...
+ written, and as such had set user expectations of how the <a>Get
+ Element Text</a> command should work. As such, the function presented
+ here is known to be flawed, but provides the best compatibility with
+ existing users.
+
+<p>The function presented here is
+ the <a href=""><code>bot.dom.getVisibleText</code></a> function
```
andreastt: Empty link.

webdriver-spec.html
```diff
@@ -8402,6 +8390,80 @@
...
+ compiled with
+ Google's <a href="https://developers.google.com/closure/compiler/">Closure
+ Compiler</a> with advanced optimizations enabled. The code may be
+ recreated by cloning the Selenium project, checking out
+ revision <a href="https://github.com/SeleniumHQ/selenium/commit/c1d39c66115053f71571b9f46d21028bd57ec83f">c1d39c6611</a>
+ and running <code>go //javascript/atoms/fragments:get-text</code>.
```
andreastt: Missing `./`

webdriver-spec.html
```diff
@@ -8402,6 +8390,80 @@
...
+function W(a,c,b,d,e){if(3==a.nodeType&&b)eb(a,c,d,e);else if(R(a))if(R(a,"CONTENT"))C(a.getDistributedNodes(),function(a){W(a,c,b,d,e)});else if(R(a,"SHADOW")){for(;a.parentNode;)a=a.parentNode;if(a instanceof ShadowRoot&&a)for(a=a.olderShadowRoot;a;)C(a.childNodes,function(a){W(a,c,b,d,e)}),a=a.olderShadowRoot}else gb(a,c)}
+function gb(a,c){a.shadowRoot&&C(a.shadowRoot.childNodes,function(a){W(a,c,!0,null,null)});bb(a,c,function(a,c,e,f,g){var r=null;1==a.nodeType?r=a:3==a.nodeType&&(r=a);r&&r.getDestinationInsertionPoints&&0&lt;r.getDestinationInsertionPoints().length||W(a,c,e,f,g)})};function hb(a){var c=[];Sa?gb(a,c):db(a,c);var b=c;a=b.length;for(var c=Array(a),b=k(b)?b.split(""):b,d=0;d&lt;a;d++)d in b&&(c[d]=ab.call(void 0,b[d]));return ab(c.join("\n")).replace(/\xa0/g," ")}var X=["_"],Y=h;X[0]in Y||!Y.execScript||Y.execScript("var "+X[0]);for(var Z;X.length&&(Z=X.shift());)X.length||void 0===hb?Y[Z]?Y=Y[Z]:Y=Y[Z]={}:Y[Z]=hb;; return this._.apply(null,arguments);}.apply({navigator:typeof window!='undefined'?window.navigator:null,document:typeof window!='undefined'?window.document:null}, arguments);}
+
```
andreastt: Drop empty line and move `</pre>` tag up behind `}`. Since this is in a `<pre>`, the whitespace will be visible.

webdriver-spec.html
andreastt: But I thought this entire section could be deleted now that we have pointer-interactability defined?

webdriver-spec.html
```diff
@@ -8477,196 +8539,69 @@
...
+function Sa(a){if(Ma){if("relative"==U(a,"position"))return 1;a=U(a,"filter");return(a=a.match(/^alpha\(opacity=(\d*)\)/)||a.match(/^progid:DXImageTransform.Microsoft.Alpha\(Opacity=(\d*)\)/))?Number(a[1])/100:1}return Va(a)}function Va(a){var c=1,b=U(a,"opacity");b&&(c=Number(b));(a=T(a))&&(c*=Va(a));return c};function Wa(a,c){var b;b=Na?function(c){if("none"==U(c,"display"))return!1;var d;do{d=c.parentNode;if(c.getDestinationInsertionPoints){var f=c.getDestinationInsertionPoints();0&lt;f.length&&(d=f[f.length-1])}if(d instanceof ShadowRoot){if(d.host.shadowRoot!=d)return!1;d=d.host}else if(9==d.nodeType||11==d.nodeType)d=null}while(a&&1!=a.nodeType);return!d||b(d)}:function(a){if("none"==U(a,"display"))return!1;a=T(a);return!a||b(a)};return Pa(a,!!c,b)}var X=["_"],Y=h;
+X[0]in Y||!Y.execScript||Y.execScript("var "+X[0]);for(var Z;X.length&&(Z=X.shift());)X.length||void 0===Wa?Y[Z]?Y=Y[Z]:Y=Y[Z]={}:Y[Z]=Wa;; return this._.apply(null,arguments);}.apply({navigator:typeof window!='undefined'?window.navigator:null,document:typeof window!='undefined'?window.document:null}, arguments);}
+
+</pre>
```
andreastt: Same regarding whitespace and `</pre>` here.

webdriver-spec.html
shs96c: Still ongoing. So far the SFC have agreed, and they're following up with Google.

webdriver-spec.html

webdriver-spec.html
shs96c: Updated with content.

webdriver-spec.html
shs96c: On Windows, `go` on its own is fine. On OS X and Linux, `./` is only needed if the `cwd` isn't on `$PATH`. Adding for clarity.

webdriver-spec.html

webdriver-spec.html
shs96c: I think there's still the case of `isDisplayed` being implemented, unless we define that in terms of pointer-intractability too….

webdriver-spec.html

webdriver-spec.html
andreastt: I can’t find any references to **element displayed** in the document. Everything else is defined in terms of pointer-interactability.

andreastt:
@shs96c When you remove `isDisplayed` and the _element displayed_ definition, I suggest we go ahead and merge this so that the document is up to date with the current reality. We can file an issue about the licensing issue and take further action from there.

shs96c:
We're still chasing to figure out licensing. The SFC have agreed to republishing the code, and they're following up with Google. I'm not keen on landing something into the spec where we're still figuring out whether or not it's okay to include the text.

webdriver-spec.html
```diff
@@ -8739,6 +8741,77 @@
...
+<div class=note>
+<p>One of the major inputs to this specification was the Open
+ Source <a href="http://seleniumhq.org">Selenium</a> project. This was
+ in wide-spread use before this specification written, and as such had
+ set user expectations of how the <a>Get Element Text</a> command
+ should work. As such, the function presented here is known to be
+ flawed, but provides the best compatibility with existing users.
+
+<p>The function presented here is
+ the <a href="https://github.com/SeleniumHQ/selenium/blob/w3c-atoms/javascript/atoms/dom.js#L979">
```
vkatsikaros: Should the link point to a commit instead of a remote branch name that could be deleted?

Example: 
https://github.com/vkatsikaros/webdriver/blob/issue_678/webdriver-spec.html#L6 (branch that used to exist on remote and was deleted now 404s)
https://github.com/vkatsikaros/webdriver/blob/master/webdriver-spec.html#L6 (branch that still exists 200s and shows content)

webdriver-spec.html
```diff
@@ -8814,196 +8887,68 @@
...
-    <li><p>Let <var>element dimensions</var> be the <a>rectangle</a>
-     that is the first element of the <code>DOMRect</code>
-     <a>sequence</a> returned by calling <a>getClientRects</a> on <var>element</var>.
-
-    <li><p>Let <var>parent style</var> be the computed style of <var>parent</var>.
-
-    <li><p>Return false if any the following conditions evaluate to false:
+<div class=note>
+<p>The function presented here is
+ the <a href="https://github.com/SeleniumHQ/selenium/blob/w3c-atoms/javascript/atoms/dom.js#L548">
```
vkatsikaros: Should the link point to a commit instead of a remote branch name that could be deleted? (same as above)

shs96c:
Alright. Further conversations with the SFC suggest that this approach of relicensing the code isn't going to work. I'm going to amend this and figure out an alternative.

webdriver-spec.html
```diff
@@ -432,6 +432,13 @@
      <li><dfn><a href="https://www.w3.org/TR/page-visibility/#dom-document-hidden">Document <code>hidden</code> attribute</a></dfn>
    </ul>
 
+ <dt>Selenium
+ <dd>The following functions are defined within
+  the <a href="http://seleniumhq.org">Selenium</a> project.
+  <ul>
+   <!-- bot.dom.getVisibleText --> <li><dfn><a href="https://github.com/SeleniumHQ/selenium/blob/1721e627e3b5ab90a06e82df1b088a33a8d11c20/javascript/atoms/dom.js#L979"><code>bot.dom.getVisibleText</code></a></dfn>
+   <!-- bot.dom.isShown --> <li><dfn><a href="https://github.com/SeleniumHQ/selenium/blob/1721e627e3b5ab90a06e82df1b088a33a8d11c20/javascript/atoms/dom.js#L548"><code>bot.dom.isShown</code></a></dfn>
+  </ul>
```
andreastt: Possibly include normative reference to SHAs in the text too? They’re only in the links now.

PR #727, opened 2017-02-06
`Actions`: Insist on an ID for each input source

### LINKED ISSUE #723: Generated action sequence id cannot be reused
In chapter [17. Actions](https://w3c.github.io/webdriver/webdriver-spec.html#actions#h-actions), section [17.3 Processing Actions Requests](https://w3c.github.io/webdriver/webdriver-spec.html#actions#h-processing-actions-requests):
> If id is undefined let id be a new UUID. Otherwise if id is not a String return error with error code invalid argument. 

When the action sequence id is generated by the remote end, that id is not returned to the local end, so the local end can't reuse it in later commands.

Example: let's say we send a command A for `(keyDown, "m")` without specifying an id. The remote end creates a new input state with id "1" and `inputStateTable["1"].pressed = ["m"]`. Then if we send another command B for `(keyDown, "f")`, again a new id is generated so the remote end has another input state for that, say "2", and we have `inputStateTable["2"].pressed = ["f"]`.  Commands A and B resulted in two independent input states although the actions might all be dispatched for the same keyboard because there's only one such device.

That seems odd. A client can unintentionally send "anonymous" action sequences in a way that makes a single input device _appear_ stateless.

If the local end wants to send many actions commands for one device, it has to create an id so that it can reuse that id in subsequent commands. 

So maybe the local end should be able to GET input source ids and types to determine what's what?

shs96c:
Closes #723

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/727)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -6635,10 +6635,8 @@ must run the following steps:
 
  <!-- TODO: consider picking the current input of the right
   type if one exists -->
- <li><p>If <var>id</var> is <a>undefined</a> let <var>id</var> be a
-  new <a>UUID</a>. Otherwise if <var>id</var> is not a <a>String</a>
-  return <a>error</a> with <a>error code</a> <a>invalid
-  argument</a>.
+ <li><p>If <var>id</var> is <a>undefined</a> or is not a <a>String</a>
+  return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
 
  <li><p>Let <var>action items</var> be the result of <a>getting a
   property</a> named <code>actions</code> from


webdriver-spec.html
```diff
@@ -6622,10 +6622,10 @@
 
  <!-- TODO: consider picking the current input of the right
   type if one exists -->
- <li><p>If <var>id</var> is <a>undefined</a> let <var>id</var> be a
-  new <a>UUID</a>. Otherwise if <var>id</var> is not a <a>String</a>
-  return <a>error</a> with <a>error code</a> <a>invalid
-  argument</a>.
+ <li><p>If <var>id</var> is <a>undefined</a> return an <a>error</a>
+  with <a>error code</a> <a>invalid argument</a>. Otherwise
```
AutomatedTester: I think this can be simplified to ` is <a>undefined</a> or is not a <a>String</a> return <a>error</a> with <a>error code</a> <a>invalid argument</a>.`

PR #730, opened 2017-02-06
`New Session`: check types of strings for named capabilities.

### LINKED ISSUE #716: Is the behaviour for wrong-type arguments intended?
In chapter [7. Capabilities](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-process-capabilities#h-capabilities), section [7.3 Processing Capabilities](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-process-capabilities#h-processing-capabilities):
> Otherwise, if capability value is not a string equal to the "browserName" entry in matched capabilities, return null. 

In this case if `browserName` has the value `2` or `["foo"]`, we simply ignore the entry. Is that intended? Should we throw an error instead?

shs96c:
Closes #716.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/730)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -2134,18 +2134,30 @@ with a "<code>moz:</code>" prefix:
 
     <dl class=switch>
      <dt>"<code>browserName</code>"
-     <dd><p>If <var>capability value</var> is
-      <a>null</a>, <a>undefined</a>, or equal to the string "<code>*</code>",
-      it is considered a match for the entry in <var>matched capabilities</var>.
+     <dd>
+      <p>If <var>capability value</var> is not a string
+       or <code>null</code>, return an <a>error</a> with <a>error
+       code</a> <a>invalid argument</a>.
+
+      <p>If <var>capability value</var> is
+       <a>null</a>, <a>undefined</a>, or equal to the string
+       "<code>*</code>", it is considered a match for the entry
+       in <var>matched capabilities</var>.
 
       <p>Otherwise, if <var>capability value</var> is not a string equal
        to the "<code>browserName</code>" entry in <var>matched capabilities</var>,
        return null.
 
      <dt>"<code>browserVersion</code>"
-     <dd><p>If <var>capability value</var> is
-      <a>null</a>, <a>undefined</a>, or equal to the string "<code>*</code>",
-      it is to be considered a match for the entry in <var>matched capabilities</var>.
+     <dd>
+      <p>If <var>capability value</var> is not a string
+       or <code>null</code>, return an <a>error</a> with <a>error
+       code</a> <a>invalid argument</a>.
+
+      <p>If <var>capability value</var> is
+       <a>null</a>, <a>undefined</a>, or equal to the string
+       "<code>*</code>", it is to be considered a match for the entry
+       in <var>matched capabilities</var>.
 
       <p>Otherwise, compare <var>capability value</var>
        to the "<code>browserVersion</code>" entry in <var>matched capabilities</var>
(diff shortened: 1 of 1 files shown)


jgraham:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[webdriver-spec.html, line 2111 at r1](https://reviewable.io:443/reviews/w3c/webdriver/730#-KcImR0Cai1FmD33rcCi:-KcImR0Cai1FmD33rcCj:b-axekri) ([raw file](https://github.com/w3c/webdriver/blob/52c6e8bd82856c0937ca4c5cda7a67df348828aa/webdriver-spec.html#L2111)):*
> ```HTML
>    <dd>Lowercase name of the user agent.
> 
>    <dt>"<code>browserVersion</code>"
> ```

Whilst I think about it, this doesn't work in real life (although it sort of makes sense as a spec fiction, it doesn't actually guide implementors in the right way). You need to get the browser properties for each entry given the composed set of capabilities (because a realistic implementation might allow passing in e.g. a path to the browser, which might affect which version it is, whether it allows insecure certs, etc.).

---

*[webdriver-spec.html, line 2140 at r1](https://reviewable.io:443/reviews/w3c/webdriver/730#-KcIm16A-yNiFUwE32cu:-KcIm16BxLnol1Dq5SaM:b-dtvio5) ([raw file](https://github.com/w3c/webdriver/blob/52c6e8bd82856c0937ca4c5cda7a67df348828aa/webdriver-spec.html#L2140)):*
> ```HTML
>       <p>If <var>capability value</var> is not a string
>        or <code>null</code>, return an <a>error</a> with <a>error
>        code</a> <a>invalid argument</a>.
> ```

This will only typecheck up to the first matching entry. So you could pass

    […]
    "firstMatch": [{"browserName": "firefox"}, {"browserName": 2}]
    […]

and it would work in firefox and throw in other browsers.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/webdriver/730)*

shs96c:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[webdriver-spec.html, line 2111 at r1](https://reviewable.io:443/reviews/w3c/webdriver/730#-KcImR0Cai1FmD33rcCi:-KcJ9ThNnkMcTuIaTPuI:bhsqt9o) ([raw file](https://github.com/w3c/webdriver/blob/52c6e8bd82856c0937ca4c5cda7a67df348828aa/webdriver-spec.html#L2111)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

Whilst I think about it, this doesn't work in real life (although it sort of makes sense as a spec fiction, it doesn't actually guide implementors in the right way). You need to get the browser properties for each entry given the composed set of capabilities (because a realistic implementation might allow passing in e.g. a path to the browser, which might affect which version it is, whether it allows insecure certs, etc.).
</blockquote></details>

I'm not sure what you mean. 

---

*[webdriver-spec.html, line 2140 at r1](https://reviewable.io:443/reviews/w3c/webdriver/730#-KcIm16A-yNiFUwE32cu:-KcJ7b5N3qSqSeR5YmhY:b-o9cg9y) ([raw file](https://github.com/w3c/webdriver/blob/52c6e8bd82856c0937ca4c5cda7a67df348828aa/webdriver-spec.html#L2140)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

This will only typecheck up to the first matching entry. So you could pass

    […]
    "firstMatch": [{"browserName": "firefox"}, {"browserName": 2}]
    […]

and it would work in firefox and throw in other browsers.
</blockquote></details>

So you're suggesting we eagerly validate the results of merged capabilities before attempting any merging? I don't see how that would offer much value in the case of any capability not in the spec, such as "firefox options"….

That seems orthogonal to this PR.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/webdriver/730)*

jgraham:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[webdriver-spec.html, line 2111 at r1](https://reviewable.io:443/reviews/w3c/webdriver/730#-KcImR0Cai1FmD33rcCi:-KcJAS-JIMcSmnZ4nWBB:b-dq87vz) ([raw file](https://github.com/w3c/webdriver/blob/52c6e8bd82856c0937ca4c5cda7a67df348828aa/webdriver-spec.html#L2111)):*
<details><summary><i>Previously, shs96c (Simon Stewart) wrote…</i></summary><blockquote>

I'm not sure what you mean. 
</blockquote></details>

In a real implementation if a user passes in `"chromeOptions:{"path": "/path/to/chome"}` or whatever, you need to check the version, and other capabilities, of chrome at that path, not some default version of Chrome. That means you can't decide what the browser capabilities are once upfront, you have to do it per possible set.

---

*[webdriver-spec.html, line 2140 at r1](https://reviewable.io:443/reviews/w3c/webdriver/730#-KcIm16A-yNiFUwE32cu:-KcJB-flJVCf_DKMtogn:bykos0i) ([raw file](https://github.com/w3c/webdriver/blob/52c6e8bd82856c0937ca4c5cda7a67df348828aa/webdriver-spec.html#L2140)):*
<details><summary><i>Previously, shs96c (Simon Stewart) wrote…</i></summary><blockquote>

So you're suggesting we eagerly validate the results of merged capabilities before attempting any merging? I don't see how that would offer much value in the case of any capability not in the spec, such as "firefox options"….

That seems orthogonal to this PR.
</blockquote></details>

It seems related to this PR in that the original issue isn't really fixed unless this is also fixed. I think I would actually prefer never erroring (and just ignoring those capabilities) to only erroring sometimes for the same input.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/webdriver/730)*

shs96c:
This will be handled in #755. Closing.

PR #731, opened 2017-02-06
`New Session`: list commonly used platform names

shs96c:
I've used SHOULD instead of MUST to give implementations some
latitude in OS naming.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/731)


shs96c:
I realise that this has come up several times in the F2F meetings, and I think we all agreed against making this a hard requirement, but without some guidance, I fear we may end up with unnecessarily incompatible versions of this spec.

PR #733, opened 2017-02-06
Check for valid Page Load Strategies and then error instead of just a…

### LINKED ISSUE #634: Deserialisation of pageLoadStrategy not defined
In chapter [7. Capabilities](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-user-prompt-handler#h-capabilities), section [7.3 Processing Capabilities](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-user-prompt-handler#h-processing-capabilities):
> Otherwise, set the "pageLoadStrategy" entry in matched capabilities to capability value. 

We need to define a deserialisation of the `pageLoadStrategy`. As it is currently, if the user passes in a string value for the capability it is unconditionally set.

I imagine we need a deserialisation definition of the definition **page load strategy** to looks in the **table of page load strategies** to see if an entry with the _keyword_ of the string exists.

AutomatedTester:
… string. Fixes #634

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/733)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -2191,7 +2191,7 @@ with a "<code>moz:</code>" prefix:
        in <var>matched capabilities</var> to <var>capability value</var>.
 
      <dt>"<code>pageLoadStrategy</code>"
-     <dd><p>If <var>capability value</var> is not a string,
+     <dd><p>If <var>capability value</var> is not a keyword in the <a>table of page load strategies</a>,
       return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
 
       <p>Otherwise, set the "<code>pageLoadStrategy</code>" entry


webdriver-spec.html
```diff
@@ -2191,7 +2191,7 @@
        in <var>matched capabilities</var> to <var>capability value</var>.
 
      <dt>"<code>pageLoadStrategy</code>"
-     <dd><p>If <var>capability value</var> is not a string,
+     <dd><p>If <var>capability value</var> is not in the <a>page load strategies table</a>,
```
shs96c: Why not just `in the <a>table of page load strategies</a>`. That would be equally good English and avoid the need for the alternative definition below.

andreastt: This also needs to highlight which column to look at in said table. Building on what Simon said above, I would suggest the following wording:

> If capability value is not a keyword in the table of page load strategies,

AutomatedTester:
fixed as requested

shs96c:
I see you've responded to @andreastt too. Merging

Issue #721, opened 2017-02-03
Current session and session id on some commands


vkatsikaros:
[Processing Model](https://w3c.github.io/webdriver/webdriver-spec.html#processing-model) sets the "current session" to the session with ID "session id"
from the url variables. However, I don't see how some commands, with `{session id}`
in their Path Template, manage to find the session on which they will run.

I apologize, if I have missed something in the definitions that clarifies this.

Examples:

> The Get Timeouts command gets timeout durations associated with the current session. 

The remote steps are running against a known session, the [current session](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-current-session). Cool.

> The Go command is used to cause the user agent to navigate the current top-level browsing context a new location. 

However, the [current top-level browsing context](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-current-top-level-browsing-context) belongs to _a_ session, not
necessarily the [current session](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-current-session). 

Other similar commands: Get Current URL, Get Window Handle, Close Window, ...

> The Find Elements command is used to find elements in the current browsing context that can be used for future commands. 

Again, the [current browsing context](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-current-browsing-context) belongs to _a_ session, not
necessarily the [current session](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-current-session). 

Other similar commands: Get Active Element, Is Element Selected, Element Click, ...

shs96c:
I'm not sure that this is a problem. Each end-node can only have one session at any given time, and an intermediary node can have any number of sessions: https://w3c.github.io/webdriver/webdriver-spec.html#dfn-maximum-active-sessions

Since this is the case, and the [processing model](https://w3c.github.io/webdriver/webdriver-spec.html#processing-model) (step 5) says that we'll have stopped processing if the session isn't active, we can't execute any command on a session that isn't active.

We may need to tighten up the language to say that end nodes are expected to perform these steps, but then from a local-end view, all kinds of [remote end](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-remote-end) are equivalent.

vkatsikaros:
Ah, I get it now. I apologize as I failed to notice it when going through the definitions.

I looked at the text to see if there was something that assisted my own oversight & stupidity. I feel that the beginning of the chapter introduction has some entangled spots:
* mixing definitions and non-normative language
[...]

PR #739, opened 2017-02-07
Improve the text flow in the Session section introduction

vkatsikaros:
* mixing definitions and non-normative language
* The "session is created by a New session request" is repeated twice
* The "remote end that is not an intermediary node has at most one
  active session at a given time." clarification is mentioned twice,
  but never alongside the definition of "active sessions".

Suggestion for https://github.com/w3c/webdriver/issues/721 I apologize if the suggestion reflects personal taste, rather than an actual text flow improvement.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/739)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -2281,7 +2281,7 @@ with a "<code>moz:</code>" prefix:
  It is an <a>error</a> to send any commands before starting a session,
  or to continue to send commands after
  the <a>session</a> has been closed.
- Maintaining session continuity between requests to
+ Maintaining session continuity between <a>commands</a> to
  the <a>remote end</a> requires passing a <a>session ID</a>.
 
 <p>A <a>session</a> is torn down at some later point;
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -2281,24 +2284,20 @@
...
- has at most one <a>active session</a> at a given time.
-
-<p>The <a>session</a> is set up at the invocation of a <a>New Session</a>,
- and torn down at some later point;
+<p>A <a>session</a> is torn down at some later point;
  either explicitly by invoking <a>Delete Session</a>,
  or implicitly when <a>Close Window</a> is called
  at the last remaining <a>top-level browsing context</a>.
 
+<p>All requests, except <a>New Session</a> and <a>Status</a> requests,
```
shs96c: s/request/command/

vkatsikaros: Done!

webdriver-spec.html
```diff
@@ -2281,17 +2281,17 @@
  It is an <a>error</a> to send any commands before starting a session,
  or to continue to send commands after
  the <a>session</a> has been closed.
- Maintaining session continuity between requests to
+ Maintaining session continuity between <a>commands</a> to
```
vkatsikaros: Same change applied here

Issue #680, opened 2017-01-27
POST request parameters


vkatsikaros:
In chapter [6. Protocol](https://w3c.github.io/webdriver/webdriver-spec.html#h-protocol), section [6.3 Processing Model](https://w3c.github.io/webdriver/webdriver-spec.html#h-processing-model):
> If request’s method is POST:
> 
> Let parse result be the result of parsing as JSON with request’s body as the argument.
> 
> If parse result is an error or if it is a success but its associated data is not an Object, send an error with error code invalid argument and jump back to step 1 in this overall algorithm.
> 
> Otherwise, let parameters be parse result’s data. 
> 
> Otherwise, let parameters be null.
> 
> ...
> 
> Let response result be the return value obtained by running the remote end steps for command with url variables as arguments. 

I see that `parameters` (JSON in the request’s body) is not used anywhere in the algorithm. Should it be one of the arguments passed to `running the remote end steps for command`?

Many of the commands that consume the parameters, do mention it as a `<var>` argument, but it's not defined somewhere in their respective algorithm. Example:

section [10.7 Resizing and Positioning Windows](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-set-window-rect#h-resizing-and-positioning-windows): 
> 4. If the properties width or height or x or y from the parameters are not integers, or less than 0 or greater than 264 – 1, return error with error code invalid argument.

andreastt:
I think probably we need to have a clearer distinction between the matches from the URI template parsing, which is sometimes referred to as ‘parameters’, and the request body/data.

vkatsikaros:
I can start working on a patch based on your comments, unless there is some ongoing work for this.

andreastt:
@vkatsikaros I haven’t made any stabs at this. Please feel free to pick it up.

shs96c:
Can I ask why we don't just say that for each command, variables in the URL template are to be treated as named variables in the algorithms for that command? That handles the case where we refer to them in the middle of the algorithm without a lot of additional boiler-plate text in the spec.

It's entirely possible I'm missing something here….

andreastt:
I think it’s the intention of this bug to fix that. As @vkatsikaros points out, we currently mix up body parameters and URI template parameters.

vkatsikaros:
Yep, my goal is to:

1. Clearly mention somewhere that `url variables` refer to the the URL template and `parameters` refer to the POST request body, and that they are the arguments to the algorithm described by the command's remote end steps.
[...]

Issue #673, opened 2017-01-25
Use terminology from RFC 6570 on URI templates


andreastt:
From the discussion in https://github.com/w3c/webdriver/pull/667 it became clear that we are not reusing terminology from the [URI Templates standard](https://tools.ietf.org/html/rfc6570) when we talk about ‘parameters’ for command endpoints.

shs96c:
I think that this is clear. The definition of `url parameters` in the `Processing Model` are:

> Otherwise, let command and url variables be request match’s data.

This makes it clear that the url parameters are those returned from `match a request`. This makes it clear that the variables being referred to are from the URI Template.

`parameters` in the processing model are derived from the POST data:

> If request’s method is POST:
>
>Let parse result be the result of parsing as JSON with request’s body as the argument.
>
>If parse result is an error or if it is a success but its associated data is not an Object, send an error with error code invalid argument and jump back to step 1 in this overall algorithm.
>
>Otherwise, let parameters be parse result’s data.
>
>Otherwise, let parameters be null.

I believe that all this is clear. Closing this issue, but please reopen if you disagree.

PR #754, opened 2017-02-09
Draft for "url variables" and "parameters" in remote end steps

vkatsikaros:
This patch is a _draft_ for #680 :
* Mention the remote end steps arguments url variables and parameters.
* Updated a couple of commands to explicitly mention that they accept url variables and parameters
* If needed, the remote end steps were reworded slightly
* (Optional) add a definition that can be linked, to clarify that `all remote end steps accept the arguments url variables and parameters` 

If you agree with this direction, I'll continue the editing.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/754)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -791,6 +791,11 @@ var respecConfig = {
  has an associated list of <dfn>remote end steps</dfn>.
  This provides the sequence of actions that a <a>remote end</a> takes
  when it receives a particular <a>command</a>.
+ All <a>remote end steps</a> accept the arguments
+ <var>url variables</var> and <var>parameters</var>, obtained from
+ the <a>command</a> as described in the
+ <a href=#processing-model>processing model</a>.
+
 </section>
 
 <section>
@@ -868,7 +873,7 @@ var respecConfig = {
 
  <li><p>Let <var>response result</var> be the return value
   obtained by running the <a>remote end steps</a> for <var>command</var>
-  with <var>url variables</var> as arguments.
+  with <var>url variables</var> and <var>parameters</var> as arguments.
 
  <li><p>If <var>response result</var> is an <a>error</a>,
   <a>send an error</a> with <a>error code</a>
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -2415,7 +2415,8 @@ with a "<code>moz:</code>" prefix:
 }</pre>
 </aside>
 
-<p>The <a>remote end steps</a> are:
+<p>The <a>remote end steps</a> with arguments <var>url variables</var>
+ and <var>parameters</var> are:
 
 <ol>
  <li><p>If the <a>remote end</a> is an <a>intermediary node</a>,
@@ -2535,7 +2536,8 @@ with a "<code>moz:</code>" prefix:
  terminates the <a>connection</a>,
  and finally <a data-lt="close the session">closes</a> the <a>current session</a>.
 
-<p>The <a>remote end steps</a> are:
+<p>The <a>remote end steps</a> with arguments <var>url variables</var>
+ and <var>parameters</var> are:
 
 <ol>
  <li><p><a>Close the session</a>.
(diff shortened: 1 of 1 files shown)


shs96c:
@vkatsikaros, would you like a review of this, or is it still a draft?

vkatsikaros:
@shs96c sorry for not clarifying earlier. I would like an opinion on the direction; I have changed a representative sample of the commands. If the direction is fine, I'll editing the rest of the commands and then it will be ready for a final review and merge.

webdriver-spec.html
```diff
@@ -791,6 +791,11 @@
  has an associated list of <dfn>remote end steps</dfn>.
  This provides the sequence of actions that a <a>remote end</a> takes
  when it receives a particular <a>command</a>.
+ All <dfn data-lt="RES-arguments"><a>remote end steps</a>
```
andreastt: What is this `RES-arguments` thing? Do we really need a definition for this?

webdriver-spec.html
```diff
@@ -791,6 +791,11 @@
...
+ accept the arguments</dfn> <var>url variables</var> and <var>parameters</var>,
```
andreastt: s/url/URL/g

andreastt: Provide reference for the typing of these two arguments. I’m guessing that the _URL variables_ should be a _URI Templates variable-list_, and that _parameters_ should be a _JSON Object_.

webdriver-spec.html
```diff
@@ -791,6 +791,11 @@
...
+ obtained from the <a>command</a> as described in the
+ <a href=#processing-model>processing model</a>.
```
andreastt: I think we should explain this when the individual commands are invoked. It’s very hard for us as implementors to parse this out-of-band normative requirement.

If you instead modify step 8 of the processing model to say that the remote end steps takes these variables, it will be much more obvious.

webdriver-spec.html
```diff
@@ -2410,7 +2415,8 @@
 }</pre>
 </aside>
 
-<p>The <a>remote end steps</a> are:
+<p>The <a>remote end steps</a> <a data-lt="RES-arguments">with arguments</a> <var>url variables</var>
+ and <var>parameters</var> are:
```
andreastt: I think we can afford to be terse and not modify these lines. If the processing model says that the remote end steps accept _URL variables_ and _parameters_, that should be sufficient. Thoughts?

webdriver-spec.html
vkatsikaros: It is used so that when a command says "The remote end steps with arguments url variables and parameters are:" there is a link from "with arguments" to this. However, if we don't want to change this for every command https://github.com/w3c/webdriver/pull/754#discussion_r102746154 this is definitely not needed. Will remove along with the change for each command

andreastt: Actually I’m ambivalent towards https://github.com/w3c/webdriver/pull/754#discussion_r102746154. Maybe it is better for clarity.

webdriver-spec.html
vkatsikaros: If the fact that remote end steps algorithm input is only mentioned in the processing model, one has to go through the  `remote end steps` link in the processing model to find out that this algorithm takes some input.

This is terse but makes it clearer for people like me who landed on the spec and are reading parts of it, without having an overall deep understanding.

andreastt: I don’t have strong feelings about whether we do or don’t change this. Do what you think is right.

vkatsikaros: Thanks for the input. I'll do some more and see how the text looks and I'll continue with the rest of the comments/changes.

AutomatedTester:
Looking at this PR, I am not really sure we need it. It feels like concentrating on #673 would solve this better.

vkatsikaros:
@AutomatedTester makes sense. I'll pause this and if there is something I'll see if there is something to submit after #673

vkatsikaros:
Closing as other work could solve this in a better way

PR #755, opened 2017-02-09
Capabilities strategies

### LINKED ISSUE #636: JSON deserialisation for timeouts object
In chapter [7. Capabilities](https://w3c.github.io/webdriver/webdriver-spec.html#h-capabilities), section [7.3 Processing Capabilities](https://w3c.github.io/webdriver/webdriver-spec.html#h-processing-capabilities):
> If capability value is not a session timeouts configuration object, return error with error code invalid argument. 

We don’t explain how to deserialise the timeouts configuration object. Which fields are optional, which are not?

Similarly to JSON serialisation for various objects, we could provide a JSON deserialisation definition for this.

shs96c:
Closes #636 amongst others.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/755)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -768,7 +768,7 @@ var respecConfig = {
 
 <p>When calling a fallible algorithm,
  the construct “Let <var>result</var> be the result
- of <dfn>trying</dfn> to call <var>algorithm</var>”
+ of <dfn data-lt="try">trying</dfn> to call <var>algorithm</var>”
  is equivalent to
 
 <ol>
@@ -2094,8 +2094,9 @@ with a "<code>moz:</code>" prefix:
    <li><p>If <var>required capabilities</var> is <a>undefined</a>,
     set the value to an empty JSON <a>Object</a>.
 
-   <li><p>If <var>required capabilities</var> is not a JSON <a>Object</a>,
-    return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
+   <li><p>Let <var>required capabilities</var> be the result
+    of <a>trying</a> to <a>validate capabilities</a> with
+    argument <var>required capabilities</var>.
   </ol>
 
  <li><p>Let <var>all first match capabilities</var> be the result of <a>getting a property</a>
@@ -2109,13 +2110,13 @@ with a "<code>moz:</code>" prefix:
    return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
   </ol>
 
+ <!-- Validate all entries first so we fail as quickly as possible -->
  <li><p>For each <var>first match capabilities</var> corresponding
   to an indexed property in <var>all first match capabilities</var>:
-
   <ol>
-   <li><p>If <var>first match capabilities</var> is not a
-    JSON <a>Object</a>, return <a>error</a> with <a>error
-    code</a> <a>invalid argument</a>.
+   <li><p>Let <var>first match capabilities</var> be the result
+    of <a>trying</a> to <a>validate capabilities</a> with argument
+    <var>first match capabilities</var>.
 
    <li><p>Let <var>merged capabilities</var> be the result of
     <a>trying</a> to <a data-lt="merging capabilities">merge
@@ -2135,6 +2136,79 @@ with a "<code>moz:</code>" prefix:
  <li><p>Return <a>success</a> with data <a><code>null</code></a>.
 </ol>
 
+<p>When required to <dfn>validate capabilities</dfn> with
+ argument <var>capability</var>:
+
+<ol>
+ <li><p>If <var>capability</var> is not a JSON <a>Object</a> return
+  an <a>error</a> with <a>error code</a> <a>invalid argument</a>.
+
+ <li><p>Let <var>result</var> be an empty JSON <a>Object</a>.
+
+ <li><p>For each enumerable <a>own property</a>
+  in <var>capability</var>, run the following substeps:
+  <ol>
+   <li><p>Let <var>name</var> be the name of the property.
+
+   <li><p>Let <var>value</var> be the result of <a>getting a
+    property</a> named <var>name</var> from <var>capability</var>.
+
+   <li><p>Run the substeps of the first matching condition:
+    <dl class=switch>
+     <dt><var>value</var> is <a><code>null</code></a>
+     <dd><p>Let <var>deserialized</var> be set to
+      <a><code>null</code></a>.
+ 
+     <dt><var>name</var> equals "<code>acceptInsecureCerts</code>"
+     <dd><p>If <var>value</var> is not a <a>boolean</a> return
+      an <a>error</a> with <a>error code</a> <a>invalid
+      argument</a>. Otherwise, let <var>deserialized</var> be
+      set to <var>value</var>
+
+     <dt><var>name</var> equals "<code>browserName</code>"
+     <dt><var>name</var> equals "<code>browserVersion</code>"
+     <dt><var>name</var> equals "<code>platformName</code>"
+     <dd><p>If <var>value</var> is not a <a>string</a> return
+      an <a>error</a> witha <a>error code</a> <a>invalid
+      argument</a>. Otherwise, let <var>deserialized</var> be
+      set to <var>value</var>.
+
+     <dt><var>name</var> equals "<code>pageLoadStrategy</code>"
+     <dd><p>Let <var>deserialized</var> be the result of <a>trying</a> to 
+      <a>deserialize as a page load strategy</a> with argument
+      <var>value</var>.
+
+     <dt><var>name</var> equals "<code>proxy</code>"
+     <dd><p>Let <var>deserialized</var> be the result of <a>trying</a>
+      to <a>deserialize as a proxy</a> with argument <var>value</var>.
+
+     <dt><var>name</var> equals "<code>timeouts</code>"
+     <dd><p>Let <var>deserialized</var> be the result of <a>trying</a>
+      to <a>deserialize as a timeout</a> with argument <var>value</var>.
+
+     <dt><var>name</var> equals "<code>unhandledPromptBehavior</code>"
+     <dd><p>Let <var>deserialized</var> be the result of <a>trying</a>
+      to <a>deserialize as an unhandled prompt behavior</a> with argument
+      <var>value</var>.
+
+     <dt><var>name</var> is the key of an <a>extension capability</a>
+     <dd><p>Let <var>deserialized</var> be the result of <a>trying</a>
+      to deserialize <var>value</var> in an implementation-specific way.
+
+     <dt>The <a>remote end</a> is an <a>endpoint node</a>
+     <dd><p>Return an <a>error</a> with <a>error code</a>
+      <a>invalid argument</a>.
+    </dl>
+
+   <li><p>If <var>deserialized</var> is
+    not <a><code>null</code></a>, <a>set a property</a>
+    on <var>result</var> with name <var>name</var> and
+    value <var>deserialized</var>.
+  </ol>
+
+ <li>Return <a>success</a> with data <var>result</var>.
+</ol><!-- /validating capabilities -->
+
 <p>When <dfn>merging capabilities</dfn> with JSON <a>Object</a> arguments
  <var>primary</var> and <var>secondary</var>,
  an <a>endpoint node</a> must take the following steps:
@@ -2150,9 +2224,8 @@ with a "<code>moz:</code>" prefix:
    <li><p>Let <var>value</var> be the the result of <a>getting a
     property</a> named <var>name</var> from <var>primary</var>.
 
-   <li><p>If <var>value</var> is not <a><code>null</code></a>
-    <a>set a property</a> on <var>result</var> with name
-    <var>name</var> and value <var>value</var>.
+   <li><a>Set a property</a> on <var>result</var> with
+    name <var>name</var> and value <var>value</var>.
   </ol>
 
  <li><p>If <var>secondary</var> is <a>undefined</a>, return <var>result</var>.
@@ -2169,19 +2242,12 @@ with a "<code>moz:</code>" prefix:
     <a>getting a property</a> named <var>name</var> from
     <var>primary</var>.
 
-   <li><p>Run the substeps of the first matching condition:
-    <dl class=switch>
-     <dt><var>value</var> is <a><code>null</code></a>
-     <dd>Do nothing.
+   <li><p>If <var>primary value</var> is
+    not <a><code>undefined</code></a> return an <a>error</a>
+    with <a>error code</a> <a>invalid argument</a>.
 
-     <dt><var>primary value</var> is <a><code>undefined</code></a>
-     <dd><a>Set a property</a> on <var>result</var> with name
+   <li><a>Set a property</a> on <var>result</var> with name
       <var>name</var> and value <var>value</var>.
-
-     <dt>Otherwise
-     <dd>Return an <a>error</a> with <a>error code</a>
-      <a>invalid argument</a>.
-    </dl>
   </ol>
 
  <li><p>Return <var>result</var>.
@@ -2249,71 +2315,25 @@ with a "<code>moz:</code>" prefix:
  <li><p>For each <var>name</var> and <var>value</var> corresponding
   to <var>capability</var>'s <a>own properties</a>:
   <ol>
-   <li><p>Run the substeps of the first matching condition:
-    <dl class=switch>
-     <dt><var>name</var> equals "<code>acceptInsecureCerts</code>"
-     <dd><p>If <var>value</var> is not a <a>boolean</a> return
-      an <a>error</a> with <a>error code</a> <a>invalid
-      argument</a>. Otherwise, let <var>capability value</var> be set
-      to <var>value</var>.
-
-     <dt><var>name</var> equals "<code>browserName</code>"
-     <dt><var>name</var> equals "<code>browserVersion</code>"
-     <dt><var>name</var> equals "<code>platformName</code>"
-     <dd><p>If <var>value</var> is not a <a>string</a> return
-      an <a>error</a> with an <a>error code</a> <a>invalid
-      argument</a>. Otherwise, let <var>capability value</var> be set
-      to <var>value</var>.  return <var>value</var>.
-
-     <dt><var>name</var> equals "<code>pageLoadStrategy</code>"
-     <dd><p>Let <var>capability value</var> be the result
-      of <a>trying</a> to <a>deserialize as a page load strategy</a>
-      <var>value</var>.
-
-     <dt><var>name</var> equals "<code>proxy</code>"
-     <dd><p>Let <var>capability value</var> be the result
-      of <a>trying</a> to <a>deserialize as a proxy</a>
-      <var>value</var>.
-
-     <dt><var>name</var> equals "<code>timeouts</code>"
-     <dd><p>Let <var>capability value</var> be the result
-      of <a>trying</a> to <a>deserialize as a timeout</a>
-      <var>value</var>.
-
-     <dt><var>name</var> equals "<code>unhandledPromptBehavior</code>"
-     <dd><p>Let <var>capability value</var> be the result
-      of <a>trying</a> to <a>deserialize as an unhandled prompt behavior</a>
-      <var>value</var>.
-
-     <dt><var>name</var> is the name of an <a>extension capability</a>
-     <dd><p>Let <var>capability value</var> be the result
-      of <a>trying</a> to deserialize <var>value</var> in an
-      implmentation-specific way.
-
-     <dt>The <a>remote end</a> is an <a>endpoint node</a>
-     <dd>Let <var>capability value</var> be <a><code>null</code></a>.
-
-     <dt>Otherwise
-     <dd>Let <var>capability value</var> be set to <var>value</var>.
-    </dl>
-
-   <li><p>If <var>capability value</var> is <a><code>null</code></a>
-    return <a>success</a> with data <a><code>null</code></a>.
-
    <li><p>Run the substeps of the first matching <var>name</var>:
 
     <dl class=switch>
      <dt>"<code>browserName</code>"
-     <dd><p>If <var>capability value</var> is not a string equal to
+     <dd><p>If <var>value</var> is not a string equal to
        the "<code>browserName</code>" entry in
        <var>matched capabilities</var>, return <a>success</a> with
        data <a><code>null</code></a>.
 
+     <p class=note>There is a chance the <a>remote end</a> will need
+      to start a browser process to correctly determine
+      the <code>browserName</code>. Lightweight checks are preferred
+      before this is done.
+
      <dt>"<code>browserVersion</code>"
-     <dd><p>Compare <var>capability value</var>
+     <dd><p>Compare <var>value</var>
        to the "<code>browserVersion</code>" entry in <var>matched capabilities</var>
        using an implementation-defined comparison algorithm.
-       The comparison is to accept a <var>capability value</var>
+       The comparison is to accept a <var>value</var>
        that places constraints on the version using
        the "<code>&lt;</code>", "<code>&lt;=</code>", "<code>&gt;</code>", 
        and "<code>&gt;=</code>" operators.
@@ -2326,8 +2346,13 @@ with a "<code>moz:</code>" prefix:
        of encoding the user agent version,
        and standardizing these schemes is beyond the scope of this standard.
 
+     <p class=note>There is a chance the <a>remote end</a> will need
+      to start a browser process to correctly determine
+      the <code>browserVersion</code>. Lightweight checks are preferred
+      before this is done.
+
      <dt>"<code>platformName</code>"
-     <dd><p>If <var>capability value</var> is not a string equal
+     <dd><p>If <var>value</var> is not a string equal
        to the "<code>platformName</code>" entry in <var>matched capabilities</var>,
        return <a>success</a> with data <a><code>null</code></a>.
 
@@ -2351,33 +2376,35 @@ with a "<code>moz:</code>" prefix:
       </div>
 
      <dt>"<code>acceptInsecureCerts</code>"
-     <dd><p>If <var>capability value</var> is not a boolean
-      return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
-
-      <p>If <var>capability value</var> is <code>true</code>
+      <dd><p>If <var>value</var> is <code>true</code>
        and the <a>endpoint node</a> does not support <a>insecure TLS certificates</a>,
-       return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
+       return <a>success</a> with data <a><code>null</code></a>.
+
+       <p class=note>If the <a>endpoint node</a> does not
+        support <a>insecure TLS certificates</a> and this is the reason
+        no match is ultimately made, it is useful to provide this
+        information to the <a>local end</a>.
 
      <dt>"<code>proxy</code>"
      <dd><p>If the <a>endpoint node</a> does not support the proxy
-       configuration defined in <var>capability value</var>,
+       configuration defined in <var>value</var>,
        return <a>success</a> with data <a><code>null</code></a>.
 
      <dt><strong>Otherwise</strong>
-     <dd><p>If <var>name</var> is not the name of an
+     <dd><p>If <var>name</var> is not the key of an
       <a>extension capability</a> and the <a>remote end</a> is
       an <a>endpoint node</a> return <a>success</a> with data 
       <code>null</code>.
 
-      <p>Let <var>matched value</var> be the result of <a>trying</a>
+      <p>Let <var>value</var> be the result of <a>trying</a>
        implementation-specific steps to match on <var>name</var>
-       with <var>capability value</var>. If the match is not
+       with <var>value</var>. If the match is not
        successful, return <a>success</a> with
        data <a><code>null</code></a>.
    </dl>
 
    <li><p><a>Set a property</a> on <var>matched capability</var>
-    with name <var>name</var> and value <var>capability value</var>.
+    with name <var>name</var> and value <var>value</var>.
   </ol>
 
  <li><p>Return <a>success</a> with data <var>matched capabilities</var>.
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -2098,39 +2098,41 @@ with a "<code>moz:</code>" prefix:
     return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
   </ol>
 
- <li><p>Let <var>desired capabilities</var> be the result of <a>getting a property</a>
+ <li><p>Let <var>all first match capabilities</var> be the result of <a>getting a property</a>
   named "<code>firstMatch</code>" from <var>capabilities request</var>.
 
   <ol>
-   <li><p>If <var>desired capabilities</var> is <a>undefined</a>,
+   <li><p>If <var>all first match capabilities</var> is <a>undefined</a>,
     set the value to a JSON <a>List</a> with a single entry of an empty JSON <a>Object</a>.
 
-   <li><p>If <var>desired capabilities</var> is not a JSON <a>List</a>,
+   <li><p>If <var>all first match capabilities</var> is not a JSON <a>List</a>,
    return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
   </ol>
 
- <li><p>For each <var>current desired capabilities</var> corresponding
-  to an indexed property in <var>desired capabilities</var>:
+ <li><p>For each <var>first match capabilities</var> corresponding
+  to an indexed property in <var>all first match capabilities</var>:
 
   <ol>
-   <li><p>If <var>current desired capabilities</var> is not a
+   <li><p>If <var>first match capabilities</var> is not a
     JSON <a>Object</a>, return <a>error</a> with <a>error
     code</a> <a>invalid argument</a>.
 
    <li><p>Let <var>merged capabilities</var> be the result of
     <a>trying</a> to <a data-lt="merging capabilities">merge
     capabilities</a> with <var>required capabilities</var>
-    and <var>current desired capabilities</var> as arguments.
+    and <var>first match capabilities</var> as arguments.
 
    <li><p>Let <var>matched capabilities</var> be the result of
-    <a>matching capabilities</a> with <var>merged capabilities</var>
-    as an argument.
+    <a>trying</a> to <a data-lt="matching capabilities">match
+    capabilities</a> with <var>merged capabilities</var> as an
+    argument.
 
-   <li><p>Return <var>matched capabilities</var> if it is
-    not <code>null</code>.
+   <li><p>If <var>matched capabilities</var> is
+    not <a><code>null</code></a> return
+    <var>matched capabilities</var>.
   </ol>
 
- <li><p>Return <code>null</code>.
+ <li><p>Return <a>success</a> with data <a><code>null</code></a>.
 </ol>
 
 <p>When <dfn>merging capabilities</dfn> with JSON <a>Object</a> arguments
@@ -2149,8 +2151,8 @@ with a "<code>moz:</code>" prefix:
     property</a> named <var>name</var> from <var>primary</var>.
 
    <li><p>If <var>value</var> is not <a><code>null</code></a>
-    <a>set a property</a> <var>name</var> to <var>value</var>
-    on <var>result</var>.
+    <a>set a property</a> on <var>result</var> with name
+    <var>name</var> and value <var>value</var>.
   </ol>
 
  <li><p>If <var>secondary</var> is <a>undefined</a>, return <var>result</var>.
@@ -2170,12 +2172,11 @@ with a "<code>moz:</code>" prefix:
    <li><p>Run the substeps of the first matching condition:
     <dl class=switch>
      <dt><var>value</var> is <a><code>null</code></a>
-     <dt><var>value</var> and <var>primary value</var> are equal.
      <dd>Do nothing.
 
      <dt><var>primary value</var> is <a><code>undefined</code></a>
-     <dd><a>Set a property</a> named <var>name</var>
-      to <var>value</var> on <var>result</var>.
+     <dd><a>Set a property</a> on <var>result</var> with name
+      <var>name</var> and value <var>value</var>.
 
      <dt>Otherwise
      <dd>Return an <a>error</a> with <a>error code</a>
@@ -2185,6 +2186,18 @@ with a "<code>moz:</code>" prefix:
 
  <li><p>Return <var>result</var>.
 </ol>
+
+<aside class=note>
+ <p>The algorithm outlined in <a>matching capabilities</a> blithely
+  ignores real-world problems that make implemention less than
+  perfectly straightforward, particularly since capbilities can
+  interact in unforeseen ways.
+
+ <p>As an example, an implementation could have a capbility that gives
+  the path to the browser binary to use. This could cause
+  both <code>browserName</code> and <code>browserVersion</code> to be
+  impossible to match against until the browser process is started.
+</aside>
 
 <p>When <dfn>matching capabilities</dfn>
  with JSON <a>Object</a> argument <var>capabilities</var>,
@@ -2196,16 +2209,16 @@ with a "<code>moz:</code>" prefix:
 
   <dl>
    <dt>"<code>browserName</code>"
-   <dd>Lowercase name of the user agent.
+   <dd>Lowercase name of the user agent as a <a>string</a>.
 
    <dt>"<code>browserVersion</code>"
-   <dd>The user agent version, as a string.
+   <dd>The user agent version, as a <a>string</a>.
 
    <dt>"<code>platformName</code>"
-   <dd>Lowercase name of the current platform.
+   <dd>Lowercase name of the current platform as a <a>string</a>.
 
    <dt>"<code>acceptInsecureCerts</code>"
-   <dd>Boolean initially set to false,
+   <dd><a>Boolean</a> initially set to <code>false</code>,
     indicating the session will not implicitly trust untrusted
     or self-signed TLS certificates on <a data-lt=go>navigation</a>.
 
@@ -2214,30 +2227,87 @@ with a "<code>moz:</code>" prefix:
     <a>commands</a> in <a href=#h-resizing-and-positioning-windows>Resizing and Positioning Windows</a>.
   </dl>
 
- <li><p>Add any implementation-specific capabilities as entries
-  to <var>matched capabilities</var>
+ <li><p>Optionally add <a>extension capabilities</a> as entries
+  to <var>matched capabilities</var>. The values of these may be
+  ellided, and there is no requirement that all
+  <a>extension capabilities</a> be added.
 
-  <p class=note>
+  <aside class=note>
+  <p>
+   This allows a <a>remote end</a> to add information that might be
+   useful to a <a>local end</a> without unnecessarily bloating the
+   response sent back to the user with (e.g.) an entire browser
+   profile.
+
+  <p>
    For example, an implementation could choose to indicate that a
    screenshot will be taken when returning an error by setting the
    capability <code>selenium.screenshot-on-error</code>
    to <code>true</code>.
-  </p>
+  </aside>
 
- <li><p>For each <var>capability name</var> corresponding to a
-  property in <var>capabilities</var>:
-
+ <li><p>For each <var>name</var> and <var>value</var> corresponding
+  to <var>capability</var>'s <a>own properties</a>:
   <ol>
-   <li><p>Let <var>capability value</var> be the value of
-    property <var>capability name</var> in <var>capabilities</var>.
+   <li><p>Run the substeps of the first matching condition:
+    <dl class=switch>
+     <dt><var>name</var> equals "<code>acceptInsecureCerts</code>"
+     <dd><p>If <var>value</var> is not a <a>boolean</a> return
+      an <a>error</a> with <a>error code</a> <a>invalid
+      argument</a>. Otherwise, let <var>capability value</var> be set
+      to <var>value</var>.
 
-   <li><p>Run the substeps of the first matching <var>capability name</var>:
+     <dt><var>name</var> equals "<code>browserName</code>"
+     <dt><var>name</var> equals "<code>browserVersion</code>"
+     <dt><var>name</var> equals "<code>platformName</code>"
+     <dd><p>If <var>value</var> is not a <a>string</a> return
+      an <a>error</a> with an <a>error code</a> <a>invalid
+      argument</a>. Otherwise, let <var>capability value</var> be set
+      to <var>value</var>.  return <var>value</var>.
+
+     <dt><var>name</var> equals "<code>pageLoadStrategy</code>"
+     <dd><p>Let <var>capability value</var> be the result
+      of <a>trying</a> to <a>deserialize as a page load strategy</a>
+      <var>value</var>.
+
+     <dt><var>name</var> equals "<code>proxy</code>"
+     <dd><p>Let <var>capability value</var> be the result
+      of <a>trying</a> to <a>deserialize as a proxy</a>
+      <var>value</var>.
+
+     <dt><var>name</var> equals "<code>timeouts</code>"
+     <dd><p>Let <var>capability value</var> be the result
+      of <a>trying</a> to <a>deserialize as a timeout</a>
+      <var>value</var>.
+
+     <dt><var>name</var> equals "<code>unhandledPromptBehavior</code>"
+     <dd><p>Let <var>capability value</var> be the result
+      of <a>trying</a> to <a>deserialize as an unhandled prompt behavior</a>
+      <var>value</var>.
+
+     <dt><var>name</var> is the name of an <a>extension capability</a>
+     <dd><p>Let <var>capability value</var> be the result
+      of <a>trying</a> to deserialize <var>value</var> in an
+      implmentation-specific way.
+
+     <dt>The <a>remote end</a> is an <a>endpoint node</a>
+     <dd>Let <var>capability value</var> be <a><code>null</code></a>.
+
+     <dt>Otherwise
+     <dd>Let <var>capability value</var> be set to <var>value</var>.
+    </dl>
+
+   <li><p>If <var>capability value</var> is <a><code>null</code></a>
+    return <a>success</a> with data <a><code>null</code></a>.
+
+   <li><p>Run the substeps of the first matching <var>name</var>:
 
     <dl class=switch>
      <dt>"<code>browserName</code>"
      <dd><p>If <var>capability value</var> is not a string equal to
        the "<code>browserName</code>" entry in
-       <var>matched capabilities</var>, return <code>null</code>.
+       <var>matched capabilities</var>, return <a>success</a> with
+       data <a><code>null</code></a>.
 
      <dt>"<code>browserVersion</code>"
      <dd><p>Compare <var>capability value</var>
@@ -2245,9 +2315,11 @@ with a "<code>moz:</code>" prefix:
        using an implementation-defined comparison algorithm.
        The comparison is to accept a <var>capability value</var>
        that places constraints on the version using
-       the "<code>&lt;</code>", "<code>&lt;=</code>", "<code>&gt;</code>", and "<code>&gt;=</code>" operators.
+       the "<code>&lt;</code>", "<code>&lt;=</code>", "<code>&gt;</code>", 
+       and "<code>&gt;=</code>" operators.
 
-      <p>If the two values do not match, return <code>null</code>.
+      <p>If the two values do not match, return <a>success</a> with
+       data <a><code>null</code></a>.
 
       <p class=note>Version comparison is left as an implementation detail
        since each user agent will likely have conflicting methods
@@ -2257,7 +2329,7 @@ with a "<code>moz:</code>" prefix:
      <dt>"<code>platformName</code>"
      <dd><p>If <var>capability value</var> is not a string equal
        to the "<code>platformName</code>" entry in <var>matched capabilities</var>,
-       return <code>null</code>.
+       return <a>success</a> with data <a><code>null</code></a>.
 
       <div class=note>
        <p>The following platform names are in common usage with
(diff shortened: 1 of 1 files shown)


shs96c:
There's a further step where we hook this up into the `Processing Capabilities` algorithms, but at least the ground work is there now.

jgraham:
OK, done a first pass, but I don't think I got everything. I'll need to re-check the logic later.

---

Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 23 unresolved discussions.

---

*[webdriver-spec.html, line 1979 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kcrst_0CIWGzGeEokkM:-Kcrst_0CIWGzGeEokkN:b-71bpy7) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L1979)):*
> coressponding

spelling

---

*[webdriver-spec.html, line 1980 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kcrvzqy6iSMfjG4ZD4D:-Kcrvzqy6iSMfjG4ZD4E:b-khghot) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L1980)):*
> ```HTML
> 
>  <li><p>For each <var>key</var> and <var>value</var> coressponding
>   to <var>parameter</var>'s <a>own properties</a>:
> ```

Not necessarily an issue for here, but I think that this construct is confusing. The fact that you have to infer semantics from the variable names is bad. You should probably iterate over the keys corresponding to enumerable own properties and get the value inside the loop.

---

*[webdriver-spec.html, line 2062 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kcrwl2fe86yc5wA21Mi:-Kcrwl2fe86yc5wA21Mj:bi94x2r) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2062)):*
> ```HTML
>     set the value to an empty JSON <a>Object</a>.
> 
>    <li><p>If the result of <a>validate capabilities</a> with
> ```

You could make `validate capabilities` return suceess or error and just `<a>try</a> to validate capabilities.`

---

*[webdriver-spec.html, line 2068 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-KcrxBuxEtZF9L5GPysS:-KcrxBuyWLO5BJGQVkG6:b-l24xtw) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2068)):*
> ```HTML
>   </ol>
> 
>  <li><p>Let <var>desired capabilities</var> be the result of <a>getting a property</a>
> ```

I think we shouldn't call this `desired capabilities` because it's historically confusing. 

---

*[webdriver-spec.html, line 2083 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kcrwze5iczq6t13bVwJ:-Kcrwze5iczq6t13bVwK:bko9rkk) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2083)):*
> ```HTML
> 
>  
>  <li><p>For each <var>current desired capabilities</var> coressponding
> ```

You still can't spell `corresponding`

---

*[webdriver-spec.html, line 2090 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-KcrxirLIteAbd6uLvR2:-KcrxirMjcf0-jt8G42f:b-v5fn7c) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2090)):*
> ```HTML
>     <var>current desired capabilities</var>.
> 
>    <li><p>If <var>mergeable</var> is <code>true</code>, append
> ```

Huh? So we just filter out all the ones that don't validate? That seems wrong. Surely we want to return an error?

---

*[webdriver-spec.html, line 2122 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-KcrxrKAfCImR7cjpGwf:-KcrxrKAfCImR7cjpGwg:bgdtbza) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2122)):*
> ```HTML
>   an <a>error</a> with <a>error code</a> <a>invalid argument</a>.
> 
>  <li>For each <var>key</var> and <var>value</var> coressponding
> ```

`corresponding`.

---

*[webdriver-spec.html, line 2131 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-KcrxxJeHAiiIRhRwqKh:-KcrxxJeHAiiIRhRwqKi:b1m4b26) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2131)):*
> ```HTML
>     return <code>false</code>.
> 
>    <li><p><a>Try</a> to call <var>strategy</var> with
> ```

You can't use `try` here (this is a "function" with return type `bool`, not one that returns success or error. `try` isn't like exceptions because it doesn't unwind the stack more than one level; it's just implicit return-on-error).

---

*[webdriver-spec.html, line 2172 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kcs112L7EH1iw0nONv-:-Kcs112MC--DW6S07ihK:buartg2) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2172)):*
> ```HTML
> 
> <ol>
>  <li><p>Let <var>matched capabilities</var> be a JSON <a>Object</a>
> ```

This still doesn't make sense, because you can't work out what these things are until you have processed the passed in capabilities (at least in a real implementation).

---

*[webdriver-spec.html, line 2191 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kcs1Iz0ArNsLRiEkdqQ:-Kcs1Iz1zAduBgGigvMY:b-8pwfgl) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2191)):*
> ```HTML
>   </dl>
> 
>  <li><p>Add any implementation-specific capabilities as entries
> ```

This is also confusing. Real implementations mostly use parameters passed to new session as configuration options. It shouldn't be required to return every configuration option that  was passed in (e.g. profile because of efficiency concerns).

---

*[webdriver-spec.html, line 2204 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kcs2bGYDw1TlvrrG-fu:-Kcs2bGZ3-gMmdcGOpmx:b-870ead) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2204)):*
> ```HTML
>   to <var>capability</var>'s <a>own properties</a>:
>   <ol>
>    <li><p>Let <var>strategy</var> be the result of <a>finding a
> ```

I feel like this deserialization strategy stuff is mostly useless indirection. We could simply have a table of allowed names and types here. We also need to deal with the possibility of extension properties which should have a name containing a colon.

---

*[webdriver-spec.html, line 2285 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kcs3jjQSdPYQyFfTJyD:-Kcs3jjWixkyvynle5Lf:b-p1w92q) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2285)):*
> ```HTML
> 
>      <dt>"<code>proxy</code>"
>      <dd><p>If the <a>endpoint node</a> does not support the proxy
> ```

I don't think this is implementable. For starters, does not support" isn't well defined. But even if it was it's generally impossible to tell if a proxy configuration is supported without actually trying it (with PAC files it is arguably equivalent to the halting problem), which you can't do until you actually start the session.

---

*[webdriver-spec.html, line 2290 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-KcsMnS1KIp7FSibnViV:-KcsMnS380KU_hZNCCYZ:byz4wqz) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2290)):*
> ```HTML
> 
>      <dt><strong>Otherwise</strong>
>      <dd><p>If <var>name</var> is not recognized
> ```

What does "recognised" mean?

Can't we constrain nonstandard options to have a `:` in the name (c.f. `moz:firefoxOptions`). That produces a clear test for whether a capability is "recognised"; it is either a standard one or starts with a vendor prefix.

---

*[webdriver-spec.html, line 2306 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-KcsNliCFZOyfdimlk9U:-KcsNliDVCKtpJZBoqgX:b-7fv4eo) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2306)):*
> ```HTML
> 
> <section>
> <h3>Deserialization Strategies</h3>
> ```

This seems like a layer of indirection that isn't necessary. Why not just call the appropriate deserialization steps directly rather than having this intermediate table?

---

*[webdriver-spec.html, line 2321 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-KcsO3KgLDMvL-fUZXE2:-KcsO3KgLDMvL-fUZXE3:bh7dz1q) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2321)):*
> ```HTML
>  </tr>
>  <tr><td>"<code>acceptInsecureCerts</code>"</td><td><a>Deserialize as a boolean</a></td></tr>
>  <tr><td>"<code>browserName</code>"</td><td><a>Deserialize as a string</a></td></tr>
> ```

I don't think that this kind of simple "deserializer" is that helpful. Elsewhere we just inlined the logic.

---

*[webdriver-spec.html, line 2334 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-KcsOHfW6dMCey3rteI1:-KcsOHfX3wa00IpGs9Am:b-1qlij9) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2334)):*
> ```HTML
> 
> <ol>
>  <li><p>If <var>value</var> is <code>null</code>
> ```

This is only for `acceptInsecureCerts`, right? I don't think that converting `null`=>`false` makes sense in general, but maybe does in that specific case. But we could always just remove any capabilities that are set to null in a single step, making it explicit that `null` is equivalent to not specifying anything.

---

*[webdriver-spec.html, line 2358 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-KcsOr2NTu0mstWoUGHx:-KcsOr2NTu0mstWoUGHy:b-x92fka) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2358)):*
> ```HTML
> </ol>
> 
> <p>To <dfn>deserialize as identity</dfn> with
> ```

Is this used? It seems pretty useless anyway and hopefully goes away with a refactor.

---

*[webdriver-spec.html, line 2822 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-KcsP2lfXa6S4I8NNSYa:-KcsP2lfXa6S4I8NNSYb:b-lqycox) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2822)):*
> ```HTML
> </table>
> 
> <p>The <a>remote end</a> steps to <dfn>deserialize as a timeout</dfn>
> ```

No need to describe these as `remote end steps` since they aren't the main entry point to the command.

---

*[webdriver-spec.html, line 2842 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-KcsRJlFh8NNKuRHA3g2:-KcsRJlG13fR6656Iv7n:bl5h4dp) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2842)):*
> ```HTML
>     or it is less than 0 or greater than 2<sup>64</sup> – 1,
>     return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
>   </ol>
> ```

This algorithm doesn't return anything; it has to return `success` with some data. Given the way it's used it should return `success` with data `parameters` (which is pretty ugly since it's returning its argument).

---

*[webdriver-spec.html, line 2924 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-KcsRhM8QXKZcHxyBbCi:-KcsRhM9yXUdGyE6wV4q:b-gfw4z9) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2924)):*
> ```HTML
> <ol>
>  <li><p>If <var>value</var> is <code>null</code>, let <var>value</var>
>   be set to "<code>normal</code>".
> ```

Quotes inside `<code>`

---

*[webdriver-spec.html, line 2933 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-KcsRmxxk302VN2EYxvv:-KcsRmxyeN_O2iaSHzwb:b-9h4qvi) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2933)):*
> ```HTML
>   an <a>error</a> with <a>error code</a> <a>invalid argument</a>.
> 
>  <li><p>Return <var>value</var>.
> ```

You have to return `success with data <var>value</var>`.

---

*[webdriver-spec.html, line 8198 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-KcsRzSPJVrdk59n4Zep:-KcsRzSPJVrdk59n4Zeq:bgs4fy5) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L8198)):*
> ```HTML
> </table>
> 
> <p>In order to <dfn>deserialize as an unhandled prompt behavior</dfn> an
> ```

"When required to", or similar.

---

*[webdriver-spec.html, line 8209 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-KcsS5_REJhG_qF8JavP:-KcsS5_REJhG_qF8JavQ:b189cr7) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L8209)):*
> ```HTML
>   code</a> <a>invalid argument</a>.
> 
>  <li><p>Return <var>value</var>.
> ```

`Return success with data <value</var>`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/webdriver/755#-:-KcsSoxoHwTKZXiK1LOa:b-5igmtz)*

shs96c:
Review status: all files reviewed at latest revision, 23 unresolved discussions.

---

*[webdriver-spec.html, line 1980 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kcrvzqy6iSMfjG4ZD4D:-KctLaGhergbgvCul84-:bsssol6) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L1980)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

Not necessarily an issue for here, but I think that this construct is confusing. The fact that you have to infer semantics from the variable names is bad. You should probably iterate over the keys corresponding to enumerable own properties and get the value inside the loop.
</blockquote></details>

I followed the same convention here that we have in other parts of the spec to do with iterating over `own properties`. I'm happy to change here, but we should also do so elsewhere too in a follow-up

---

*[webdriver-spec.html, line 2062 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kcrwl2fe86yc5wA21Mi:-KctLq4bwOOUh2LQouto:b-k4z9z7) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2062)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

You could make `validate capabilities` return suceess or error and just `<a>try</a> to validate capabilities.`
</blockquote></details>

Good point :)

---

*[webdriver-spec.html, line 2068 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-KcrxBuxEtZF9L5GPysS:-KctLrb2DDAcXpSacpwB:b-tbm7g) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2068)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

I think we shouldn't call this `desired capabilities` because it's historically confusing. 
</blockquote></details>

SGTM. Will change.

---

*[webdriver-spec.html, line 2083 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kcrwze5iczq6t13bVwJ:-KctLuV8cKZEL6P3XPPh:b-p7c3t5) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2083)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

You still can't spell `corresponding`
</blockquote></details>

It's weekness….

---

*[webdriver-spec.html, line 2090 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-KcrxirLIteAbd6uLvR2:-KctLxVEBxqlHmX1L5VB:b4btd2r) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2090)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

Huh? So we just filter out all the ones that don't validate? That seems wrong. Surely we want to return an error?
</blockquote></details>

There are some capabilities that are valid but non-applicable (eg. "browserName -> chromedriver" being sent to geckodriver.

---

*[webdriver-spec.html, line 2131 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-KcrxxJeHAiiIRhRwqKh:-KctM9hPfuCjMiquqbwM:b46bcz4) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2131)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

You can't use `try` here (this is a "function" with return type `bool`, not one that returns success or error. `try` isn't like exceptions because it doesn't unwind the stack more than one level; it's just implicit return-on-error).
</blockquote></details>

I didn't realise that. Will reword things.

---

*[webdriver-spec.html, line 2191 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kcs1Iz0ArNsLRiEkdqQ:-KctMMr-sTIzrV595tbv:b-49zuji) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2191)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

This is also confusing. Real implementations mostly use parameters passed to new session as configuration options. It shouldn't be required to return every configuration option that  was passed in (e.g. profile because of efficiency concerns).
</blockquote></details>

I wanted to put a reasonable place for implementations to add their own information. Otherwise, we never return anything other than the capabilities that people ask for. In the extreme case, this may mean people can't even tell the "browserName" of the remote end, which is less than ideal. Perhaps "optionally" would help?

---

*[webdriver-spec.html, line 2204 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kcs2bGYDw1TlvrrG-fu:-KctMieCSTI6ZQeKk2Wk:b-51c761) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2204)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

I feel like this deserialization strategy stuff is mostly useless indirection. We could simply have a table of allowed names and types here. We also need to deal with the possibility of extension properties which should have a name containing a colon.
</blockquote></details>

As a developer, this made sense to me --- basically a lookup in a map with a default return value. I could do the table thing too, but the nice thing here is we avoid duplicating information that we've already made normative in the spec.

---

*[webdriver-spec.html, line 2285 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kcs3jjQSdPYQyFfTJyD:-KctN0J93HnlfajFpZZP:bnj7ty6) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2285)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

I don't think this is implementable. For starters, does not support" isn't well defined. But even if it was it's generally impossible to tell if a proxy configuration is supported without actually trying it (with PAC files it is arguably equivalent to the halting problem), which you can't do until you actually start the session.
</blockquote></details>

This is the language that was there before. I just left it alone. Things like safaridriver can probably tell up front that they're not allowed to change system proxy settings.

---

*[webdriver-spec.html, line 2290 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-KcsMnS1KIp7FSibnViV:-KctNSFk2lvupfFLQQa8:b4s1rtq) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2290)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

What does "recognised" mean?

Can't we constrain nonstandard options to have a `:` in the name (c.f. `moz:firefoxOptions`). That produces a clear test for whether a capability is "recognised"; it is either a standard one or starts with a vendor prefix.
</blockquote></details>

Not unless we tighten the language in `protocol extensions`, which would break most implementations out there (SauceLabs and BrowserStack don't prefix their extensions, for example. Nor does selenium)

---

*[webdriver-spec.html, line 2306 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-KcsNliCFZOyfdimlk9U:-KctNemopwyZWAkQnCVW:bhox6ia) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2306)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

This seems like a layer of indirection that isn't necessary. Why not just call the appropriate deserialization steps directly rather than having this intermediate table?
</blockquote></details>

Because there are multiple steps that follow the same deserialisation strategy (browserName and browserVersion, for example), and I hate copy-and-pasting.

---

*[webdriver-spec.html, line 2321 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-KcsO3KgLDMvL-fUZXE2:-KctNo9f41X7OOMlZktG:bnvexuu) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2321)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

I don't think that this kind of simple "deserializer" is that helpful. Elsewhere we just inlined the logic.
</blockquote></details>

Boolean has the quirk that `null` becomes `false`. I can inline both.

---

*[webdriver-spec.html, line 2334 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-KcsOHfW6dMCey3rteI1:-KctNw6aVtYg94Kik2dC:b3ztzlk) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2334)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

This is only for `acceptInsecureCerts`, right? I don't think that converting `null`=>`false` makes sense in general, but maybe does in that specific case. But we could always just remove any capabilities that are set to null in a single step, making it explicit that `null` is equivalent to not specifying anything.
</blockquote></details>

I like that idea. Will amend.

---

*[webdriver-spec.html, line 2358 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-KcsOr2NTu0mstWoUGHx:-KctNyhmZF63KR_zqNw4:b-yh3y4z) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2358)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

Is this used? It seems pretty useless anyway and hopefully goes away with a refactor.
</blockquote></details>

Identity does get used. It's for intermediary nodes so that they pass capabilities through without changing them.

---

*[webdriver-spec.html, line 2924 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-KcsRhM8QXKZcHxyBbCi:-KctOAfq9MuQ1cIn0ls0:b-9a84vf) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2924)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

Quotes inside `<code>`
</blockquote></details>

That's not how @andreastt likes things to be done, and I've adopted his preferred style. I'm happy to go either way.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/webdriver/755)*

shs96c:
I need to make the amendments, but there are some questions and comments that I hope make some bits clearer. Follow up diffs tomorrow afternoon.

---

Review status: all files reviewed at latest revision, 23 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/webdriver/755#-:-KctOZaHQVwccRZc4Z7L:b-ln2jwq)*

shs96c:
Review status: all files reviewed at latest revision, 23 unresolved discussions, some commit checks failed.

---

*[webdriver-spec.html, line 1979 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kcrst_0CIWGzGeEokkM:-KcxAIdP9DYsnrJfZYAC:bt2vr0b) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L1979)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

> coressponding

spelling
</blockquote></details>

Fixed

---

*[webdriver-spec.html, line 2172 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kcs112L7EH1iw0nONv-:-KcxkCl0Vns3brxKIAGo:b-sgqfrh) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2172)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

This still doesn't make sense, because you can't work out what these things are until you have processed the passed in capabilities (at least in a real implementation).
</blockquote></details>

We're building up the set of capabilities that have been matched. We'll either return them or `null`. As we populate this, we're checking that the values match. If they don't, `null` is returned.

---

*[webdriver-spec.html, line 2822 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-KcsP2lfXa6S4I8NNSYa:-Kd0zCDz9x41Gu1jEjQ-:bt2vr1b) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2822)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

No need to describe these as `remote end steps` since they aren't the main entry point to the command.
</blockquote></details>

Fixed.

---

*[webdriver-spec.html, line 2842 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-KcsRJlFh8NNKuRHA3g2:-Kd1-itroyxGULQbM6Bq:bt2vr1b) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2842)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

This algorithm doesn't return anything; it has to return `success` with some data. Given the way it's used it should return `success` with data `parameters` (which is pretty ugly since it's returning its argument).
</blockquote></details>

Fixed.

---

*[webdriver-spec.html, line 2933 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-KcsRmxxk302VN2EYxvv:-Kd10JJwsWYKeJR56h78:b-886gsx) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2933)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

You have to return `success with data <var>value</var>`.
</blockquote></details>

Fixed

---

*[webdriver-spec.html, line 8198 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-KcsRzSPJVrdk59n4Zep:-Kd10WNLG4vgnzvC6YLr:bb6un1j) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L8198)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

"When required to", or similar.
</blockquote></details>

Done

---

*[webdriver-spec.html, line 8209 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-KcsS5_REJhG_qF8JavP:-Kd10eWaWYWMOnVKhI8r:bb6un1j) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L8209)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

`Return success with data <value</var>`
</blockquote></details>

Done

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/webdriver/755)*

shs96c:
Review status: 0 of 1 files reviewed at latest revision, 23 unresolved discussions.

---

*[webdriver-spec.html, line 2122 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-KcrxrKAfCImR7cjpGwf:-Kd1Dr0g3URmJ1javYA6:b-886gsx) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2122)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

`corresponding`.
</blockquote></details>

Fixed

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/webdriver/755)*

jgraham:
Review status: 0 of 1 files reviewed at latest revision, 16 unresolved discussions.

---

*[webdriver-spec.html, line 1980 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kcrvzqy6iSMfjG4ZD4D:-Kd1H-RfkpEV3OeyrT0U:bo30x99) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L1980)):*
<details><summary><i>Previously, shs96c (Simon Stewart) wrote…</i></summary><blockquote>

I followed the same convention here that we have in other parts of the spec to do with iterating over `own properties`. I'm happy to change here, but we should also do so elsewhere too in a follow-up
</blockquote></details>

https://github.com/w3c/webdriver/issues/772

---

*[webdriver-spec.html, line 2285 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kcs3jjQSdPYQyFfTJyD:-Kd1NHk_GHfI_cEoHHfG:bl203c9) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2285)):*
<details><summary><i>Previously, shs96c (Simon Stewart) wrote…</i></summary><blockquote>

This is the language that was there before. I just left it alone. Things like safaridriver can probably tell up front that they're not allowed to change system proxy settings.
</blockquote></details>

I still think this language needs to be weakened. It seems like you're saying there are exceptional circumstances in which an implementation can know things about the allowed proxy configuration upfront. I agree in that case it should be permitted to error here, but there shouldn't be a must-level requirement to determine that.

---

*[webdriver-spec.html, line 2290 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-KcsMnS1KIp7FSibnViV:-Kd1PrUK24S_vbs16FlG:b-bukk65) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2290)):*
<details><summary><i>Previously, shs96c (Simon Stewart) wrote…</i></summary><blockquote>

Not unless we tighten the language in `protocol extensions`, which would break most implementations out there (SauceLabs and BrowserStack don't prefix their extensions, for example. Nor does selenium)
</blockquote></details>

I think we should have this discussion. People making up random stuff in a global namespace limits our ability to standardise going forward (we have to know about all existing extensions to avoid name clashes).

---

*[webdriver-spec.html, line 2924 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-KcsRhM8QXKZcHxyBbCi:-Kd1QhG95rtY60YUnXVc:b-e2e9iu) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2924)):*
<details><summary><i>Previously, shs96c (Simon Stewart) wrote…</i></summary><blockquote>

That's not how @andreastt likes things to be done, and I've adopted his preferred style. I'm happy to go either way.
</blockquote></details>

I… don't see what the argument for the alternative is.  The code here is a string literal. Surely then the markers of it being such form part of the `<code>`.

---

*[webdriver-spec.html, line 2113 at r3](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kd1JaMYKYT3-nLrgNy4:-Kd1JaMYKYT3-nLrgNy5:b-7sqhyf) ([raw file](https://github.com/w3c/webdriver/blob/48e327c0ddc189763820a1c808ecbb5f126be925/webdriver-spec.html#L2113)):*
> ```HTML
>     argument.
> 
>    <li><p>If <var>matched capabilities</var>'s is
> ```

The "s" should be inside the `<var>` and there shouldn't be an apostrophe.

---

*[webdriver-spec.html, line 2154 at r3](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kd1KZemPgW_IsN5D8bF:-Kd1KZemPgW_IsN5D8bG:boc59us) ([raw file](https://github.com/w3c/webdriver/blob/48e327c0ddc189763820a1c808ecbb5f126be925/webdriver-spec.html#L2154)):*
> ```HTML
>      <dt><var>name</var> equals "<code>platformName</code>"
>      <dd><p>If <var>value</var> is not a <a>string</a> return
>       an <a>error</a> witha <a>error code</a> <a>invalid
> ```

"witha"

---

*[webdriver-spec.html, line 2185 at r3](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kd1LQlVoKnhjg8qig_G:-Kd1LQlWCPBYBR7weOKu:bepmhl0) ([raw file](https://github.com/w3c/webdriver/blob/48e327c0ddc189763820a1c808ecbb5f126be925/webdriver-spec.html#L2185)):*
> ```HTML
>     </dl>
> 
>    <li><p>If <var>deserialized</var> is
> ```

`, <a>set a property</a> of <var>result</var> with name <var>name</var> and value <var>deserialized</var>`

---

*[webdriver-spec.html, line 2209 at r3](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kd1M2AQQUFsYCyMvRQP:-Kd1M2ARfyvog9tl6D7s:byqtazc) ([raw file](https://github.com/w3c/webdriver/blob/48e327c0ddc189763820a1c808ecbb5f126be925/webdriver-spec.html#L2209)):*
> ```HTML
>     property</a> named <var>name</var> from <var>primary</var>.
> 
>    <li><a>Set a property</a> <var>name</var> to <var>value</var>
> ```

Set a property on result with name[…]

---

*[webdriver-spec.html, line 2227 at r3](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kd1IAFG9vvfpBaU9SXM:-Kd1IAFG9vvfpBaU9SXN:b-16cipc) ([raw file](https://github.com/w3c/webdriver/blob/48e327c0ddc189763820a1c808ecbb5f126be925/webdriver-spec.html#L2227)):*
> ```HTML
>     <var>primary</var>.
> 
>    <li><p>Run the substeps of the first matching condition:
> ```

Is there a use case for allowing this, rather than just rejecting cases where there's a key in common? Note that "equal" is not obviously well defined here (does the empty string equal null? Do two Objects with equal keys and values equal each other?).

---

*[webdriver-spec.html, line 2254 at r3](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kd1MPm-IEDAiIaU9cSU:-Kd1MPm-IEDAiIaU9cSV:b-ufd3fb) ([raw file](https://github.com/w3c/webdriver/blob/48e327c0ddc189763820a1c808ecbb5f126be925/webdriver-spec.html#L2254)):*
> ```HTML
> 
>   <dl>
>    <dt>"<code>browserName</code>"
> ```

I still think we need at least a note to indicate that in a real implementation these values may depend on other capabilities that are passed in. Designing that is the central problem in actually implementing this stuff so it is helpful if the spec acknowledges that its model is simplified.

---

*[webdriver-spec.html, line 2343 at r3](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kd1N4MOPXzPXU215Q7p:-Kd1N4MOPXzPXU215Q7q:bpuc7gm) ([raw file](https://github.com/w3c/webdriver/blob/48e327c0ddc189763820a1c808ecbb5f126be925/webdriver-spec.html#L2343)):*
> ```HTML
>       <dd><p>If <var>value</var> is <code>true</code>
>        and the <a>endpoint node</a> does not support <a>insecure TLS certificates</a>,
>        return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
> ```

Why are we returning an error in this case rather than null? You might have a second set of capabilities that don't require accepting insecure SSL.

---

*[webdriver-spec.html, line 2361 at r3](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kd1QJQ-kwZNZCG2eAF4:-Kd1QJQ-kwZNZCG2eAF5:bfng7rv) ([raw file](https://github.com/w3c/webdriver/blob/48e327c0ddc189763820a1c808ecbb5f126be925/webdriver-spec.html#L2361)):*
> ```HTML
>    </dl>
> 
>    <li><p>Create a new entry in <var>matched capabilities</var>
> ```

What does "create a new entry" mean?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/webdriver/755)*

jgraham:
Reviewed 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 15 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/webdriver/755)*

shs96c:
Review status: all files reviewed at latest revision, 16 unresolved discussions.

---

*[webdriver-spec.html, line 2285 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kcs3jjQSdPYQyFfTJyD:-Kd1XEX-IRB7gbiseIA_:b8qmuma) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2285)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

I still think this language needs to be weakened. It seems like you're saying there are exceptional circumstances in which an implementation can know things about the allowed proxy configuration upfront. I agree in that case it should be permitted to error here, but there shouldn't be a must-level requirement to determine that.
</blockquote></details>

I think there does need to be a check once somewhere in the end node that ensures that both "acceptInsecureCerts" and the "proxy" capabilities have been set and will be honoured. 

---

*[webdriver-spec.html, line 2290 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-KcsMnS1KIp7FSibnViV:-Kd1ZNHTe7UrcHrCSAkg:b-nl4nvv) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2290)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

I think we should have this discussion. People making up random stuff in a global namespace limits our ability to standardise going forward (we have to know about all existing extensions to avoid name clashes).
</blockquote></details>

Realistically, people are already extending the spec with non-standard capability names that don't contain a colon, and there's nothing to say that there can't be consensus which has yet to be formalised around a capability name (the Appium, Selendroid, WebDriver Agent folks have some agreed upon capability names that are shared across multiple implementations).

I'll use "implementation-specific" as elsewhere for this text.

---

*[webdriver-spec.html, line 2924 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-KcsRhM8QXKZcHxyBbCi:-Kd1_nqXwKWtapZR-xZV:b-sgm1g7) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2924)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

I… don't see what the argument for the alternative is.  The code here is a string literal. Surely then the markers of it being such form part of the `<code>`.
</blockquote></details>

@andreastt's argument is that there might be confusion about whether or not the quote marks should be part of the string or not. In his view, the quote marks indicate that the type is a string, and value within the <code> section is the actual contents of the string.

I don't care either way, but it'd be nice if we all agreed how to handle string values in code tags. For now, I'd prefer consistency with the rest of the spec, and then a bug to tidy up afterwards.

---

*[webdriver-spec.html, line 2113 at r3](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kd1JaMYKYT3-nLrgNy4:-Kd1VCSr8z1oaoSp51CX:b-886gsx) ([raw file](https://github.com/w3c/webdriver/blob/48e327c0ddc189763820a1c808ecbb5f126be925/webdriver-spec.html#L2113)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

The "s" should be inside the `<var>` and there shouldn't be an apostrophe.
</blockquote></details>

Fixed

---

*[webdriver-spec.html, line 2154 at r3](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kd1KZemPgW_IsN5D8bF:-Kd1VU7MVBd8k4izDItE:b-886gsx) ([raw file](https://github.com/w3c/webdriver/blob/48e327c0ddc189763820a1c808ecbb5f126be925/webdriver-spec.html#L2154)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

"witha"
</blockquote></details>

Fixed

---

*[webdriver-spec.html, line 2185 at r3](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kd1LQlVoKnhjg8qig_G:-Kd1b7-obTVRClq-FeD-:b-886gsx) ([raw file](https://github.com/w3c/webdriver/blob/48e327c0ddc189763820a1c808ecbb5f126be925/webdriver-spec.html#L2185)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

`, <a>set a property</a> of <var>result</var> with name <var>name</var> and value <var>deserialized</var>`
</blockquote></details>

Fixed

---

*[webdriver-spec.html, line 2209 at r3](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kd1M2AQQUFsYCyMvRQP:-Kd1W6cR0WEEMLFYs59b:b-886gsx) ([raw file](https://github.com/w3c/webdriver/blob/48e327c0ddc189763820a1c808ecbb5f126be925/webdriver-spec.html#L2209)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

Set a property on result with name[…]
</blockquote></details>

Fixed

---

*[webdriver-spec.html, line 2227 at r3](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kd1IAFG9vvfpBaU9SXM:-Kd1dx_QTMjE5LdOYiwC:b-80662t) ([raw file](https://github.com/w3c/webdriver/blob/48e327c0ddc189763820a1c808ecbb5f126be925/webdriver-spec.html#L2227)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

Is there a use case for allowing this, rather than just rejecting cases where there's a key in common? Note that "equal" is not obviously well defined here (does the empty string equal null? Do two Objects with equal keys and values equal each other?).
</blockquote></details>

I can imagine local ends duplicating information here because of user inputs. We can use `Object.is` for comparison, which I think gets the semantics right.

---

*[webdriver-spec.html, line 2252 at r3](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kd1WyRsb0lQxZIFMKSE:-Kd1WyRsb0lQxZIFMKSF:b-w19rsd) ([raw file](https://github.com/w3c/webdriver/blob/48e327c0ddc189763820a1c808ecbb5f126be925/webdriver-spec.html#L2252)):*
> I still think we need at least a note to indicate that in a real implementation these values may depend on other capabilities that are passed in.

---

*[webdriver-spec.html, line 2254 at r3](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kd1MPm-IEDAiIaU9cSU:-Kd1WJ0xE3C6amar_5FY:b-ty1jd9) ([raw file](https://github.com/w3c/webdriver/blob/48e327c0ddc189763820a1c808ecbb5f126be925/webdriver-spec.html#L2254)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

I still think we need at least a note to indicate that in a real implementation these values may depend on other capabilities that are passed in. Designing that is the central problem in actually implementing this stuff so it is helpful if the spec acknowledges that its model is simplified.
</blockquote></details>

The `browserName` may depend on other capabilities? Does the end node not know its own name?

---

*[webdriver-spec.html, line 2343 at r3](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kd1N4MOPXzPXU215Q7p:-Kd1X0R0RgYi4SvLt56y:b-k4z08q) ([raw file](https://github.com/w3c/webdriver/blob/48e327c0ddc189763820a1c808ecbb5f126be925/webdriver-spec.html#L2343)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

Why are we returning an error in this case rather than null? You might have a second set of capabilities that don't require accepting insecure SSL.
</blockquote></details>

Good point. 

---

*[webdriver-spec.html, line 2361 at r3](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kd1QJQ-kwZNZCG2eAF4:-Kd1_k-nPGV3vs3fwyIk:b-2rlu5y) ([raw file](https://github.com/w3c/webdriver/blob/48e327c0ddc189763820a1c808ecbb5f126be925/webdriver-spec.html#L2361)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

What does "create a new entry" mean?
</blockquote></details>

*ahem* Fixed

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/webdriver/755)*

jgraham:
Review status: 0 of 1 files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[webdriver-spec.html, line 2172 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kcs112L7EH1iw0nONv-:-KdRAWESs6tkOb0Q9iiv:b-sgaxo) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2172)):*
<details><summary><i>Previously, shs96c (Simon Stewart) wrote…</i></summary><blockquote>

We're building up the set of capabilities that have been matched. We'll either return them or `null`. As we populate this, we're checking that the values match. If they don't, `null` is returned.
</blockquote></details>

I think we agreed to add a note here?

---

*[webdriver-spec.html, line 2285 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kcs3jjQSdPYQyFfTJyD:-KdRBqMAx7U3wiviimbI:bqs46h4) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2285)):*
<details><summary><i>Previously, shs96c (Simon Stewart) wrote…</i></summary><blockquote>

I think there does need to be a check once somewhere in the end node that ensures that both "acceptInsecureCerts" and the "proxy" capabilities have been set and will be honoured. 
</blockquote></details>

We can probably tell if that configuration is set. We can't tell if it's supported in the sense of "actually works". The language here is still wrong. I at least want an issue open to fix this.

---

*[webdriver-spec.html, line 2290 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-KcsMnS1KIp7FSibnViV:-KdRCB3Defurpfo5eur-:b-ityl34) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2290)):*
<details><summary><i>Previously, shs96c (Simon Stewart) wrote…</i></summary><blockquote>

Realistically, people are already extending the spec with non-standard capability names that don't contain a colon, and there's nothing to say that there can't be consensus which has yet to be formalised around a capability name (the Appium, Selendroid, WebDriver Agent folks have some agreed upon capability names that are shared across multiple implementations).

I'll use "implementation-specific" as elsewhere for this text.
</blockquote></details>

So we can't ever add a capability without auditing every extant implementation? That seems like a disaster.

---

*[webdriver-spec.html, line 2924 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-KcsRhM8QXKZcHxyBbCi:-KdREQabT664YLZMdCku:bo30xa7) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2924)):*
<details><summary><i>Previously, shs96c (Simon Stewart) wrote…</i></summary><blockquote>

@andreastt's argument is that there might be confusion about whether or not the quote marks should be part of the string or not. In his view, the quote marks indicate that the type is a string, and value within the <code> section is the actual contents of the string.

I don't care either way, but it'd be nice if we all agreed how to handle string values in code tags. For now, I'd prefer consistency with the rest of the spec, and then a bug to tidy up afterwards.
</blockquote></details>

https://github.com/w3c/webdriver/issues/785

---

*[webdriver-spec.html, line 2227 at r3](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kd1IAFG9vvfpBaU9SXM:-KdR9jkxOZXaKQuYqyTp:b-v9r6j2) ([raw file](https://github.com/w3c/webdriver/blob/48e327c0ddc189763820a1c808ecbb5f126be925/webdriver-spec.html#L2227)):*
<details><summary><i>Previously, shs96c (Simon Stewart) wrote…</i></summary><blockquote>

I can imagine local ends duplicating information here because of user inputs. We can use `Object.is` for comparison, which I think gets the semantics right.
</blockquote></details>

I'm still not happy with this. Although our model is nominally js objects, I don't think that anyone is actually implementing this in js, so everyone will have to duplicate the logic from the js spec. I imagine this will create a lot of possibility for poor interop. On the other hand it's easy for a local end to do something sensible if a user adds duplicate keys to alwaysMatch and firstMatch (e.g. throw, remove from firstMatch). I also think that's clearer for users who might otherwise think they can merge complex values or something.

---

*[webdriver-spec.html, line 2254 at r3](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kd1MPm-IEDAiIaU9cSU:-KdRAeT5zDgG8c8JFXe3:bcjb96j) ([raw file](https://github.com/w3c/webdriver/blob/48e327c0ddc189763820a1c808ecbb5f126be925/webdriver-spec.html#L2254)):*
<details><summary><i>Previously, shs96c (Simon Stewart) wrote…</i></summary><blockquote>

The `browserName` may depend on other capabilities? Does the end node not know its own name?
</blockquote></details>

If Microsoft had had "IEDriver" during the Edge transition then it might have occurred that a single endpoint could front a product called Edge or one called InternetExplorer depending on the path. It is easy to imagine similar rebranding  "it's a new product, honest" exercises in the future.

---

*[webdriver-spec.html, line 2210 at r4](https://reviewable.io:443/reviews/w3c/webdriver/755#-KdR9N_1IBXB1eCPP5m5:-KdR9N_2p5YYBLXqkQPZ:b-oz5luv) ([raw file](https://github.com/w3c/webdriver/blob/126794b8cf9cde616dda3fb3d1d3cce5bb0cd60b/webdriver-spec.html#L2210)):*
> ```HTML
>     property</a> named <var>name</var> from <var>primary</var>.
> 
>    <li><a>Set a property</a> of <var>result</var> with
> ```

s/of/on/

---

*[webdriver-spec.html, line 2300 at r4](https://reviewable.io:443/reviews/w3c/webdriver/755#-KdRBT8OhBgmd_pVg-SZ:-KdRBT8P4_Lrzo4J_XoN:b48cjoj) ([raw file](https://github.com/w3c/webdriver/blob/126794b8cf9cde616dda3fb3d1d3cce5bb0cd60b/webdriver-spec.html#L2300)):*
> ```HTML
>        data <a><code>null</code></a>.
> 
>      <p class=note>There is a chance the <a>remote end</a> will need
> ```

I don't think these notes should be here, but higher up as a general note that the structure of this algorithm is based on a premise that isn't true in real implementations.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/webdriver/755)*

jgraham:
Reviewed 1 of 1 files at r5.
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/webdriver/755)*

shs96c:
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[webdriver-spec.html, line 2172 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kcs112L7EH1iw0nONv-:-KdRU6I7_E8FYK5D-ZZM:bscgl21) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2172)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

I think we agreed to add a note here?
</blockquote></details>

We did, and it's further down. :)

---

*[webdriver-spec.html, line 2285 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kcs3jjQSdPYQyFfTJyD:-KdRUopJeVA6LRX0pyAQ:b-fz77dl) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2285)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

We can probably tell if that configuration is set. We can't tell if it's supported in the sense of "actually works". The language here is still wrong. I at least want an issue open to fix this.
</blockquote></details>

Discussed on IRC. I've raised #786 to address this.

---

*[webdriver-spec.html, line 2290 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-KcsMnS1KIp7FSibnViV:-KdRVl4khMqfdiUjcd23:bizw6wh) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2290)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

So we can't ever add a capability without auditing every extant implementation? That seems like a disaster.
</blockquote></details>

We followed the CSS spec for protocol extensions, and they don't require extensions to use the vendor prefix --- https://www.w3.org/TR/CSS21/syndata.html#vendor-keywords If we're doomed, at least it'll be a party with plenty of company.

I can add a note in the protocol extensions section to say that no spec-supplied capability name will ever have a colon in it.

---

*[webdriver-spec.html, line 2227 at r3](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kd1IAFG9vvfpBaU9SXM:-KdRMwo-UZzkd61yPinc:bgvysxo) ([raw file](https://github.com/w3c/webdriver/blob/48e327c0ddc189763820a1c808ecbb5f126be925/webdriver-spec.html#L2227)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

I'm still not happy with this. Although our model is nominally js objects, I don't think that anyone is actually implementing this in js, so everyone will have to duplicate the logic from the js spec. I imagine this will create a lot of possibility for poor interop. On the other hand it's easy for a local end to do something sensible if a user adds duplicate keys to alwaysMatch and firstMatch (e.g. throw, remove from firstMatch). I also think that's clearer for users who might otherwise think they can merge complex values or something.
</blockquote></details>

I can see the argument for a lack of interoperability. I guess if the validation step only checked that deserialisation was possible we'd still have the raw JSON blob at this point. I'd be amazed if most libraries for JSON in any language didn't have some mechanism for verifying that two blobs are equivalent.

I'll remove the equality check, but I think this will cause problems for users further down the line since local end bindings are written by folks not used to reading specs.

---

*[webdriver-spec.html, line 2254 at r3](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kd1MPm-IEDAiIaU9cSU:-KdRUBjsMNYDFDsxes__:b-sqhk6e) ([raw file](https://github.com/w3c/webdriver/blob/48e327c0ddc189763820a1c808ecbb5f126be925/webdriver-spec.html#L2254)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

If Microsoft had had "IEDriver" during the Edge transition then it might have occurred that a single endpoint could front a product called Edge or one called InternetExplorer depending on the path. It is easy to imagine similar rebranding  "it's a new product, honest" exercises in the future.
</blockquote></details>

*sigh* Marketing.

---

*[webdriver-spec.html, line 2210 at r4](https://reviewable.io:443/reviews/w3c/webdriver/755#-KdR9N_1IBXB1eCPP5m5:-KdRMlWXJn1pfRNZPpTl:bb6un1j) ([raw file](https://github.com/w3c/webdriver/blob/126794b8cf9cde616dda3fb3d1d3cce5bb0cd60b/webdriver-spec.html#L2210)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

s/of/on/
</blockquote></details>

Done

---

*[webdriver-spec.html, line 2300 at r4](https://reviewable.io:443/reviews/w3c/webdriver/755#-KdRBT8OhBgmd_pVg-SZ:-KdRUFTEn3joyuprD4Ay:b-mk2y4r) ([raw file](https://github.com/w3c/webdriver/blob/126794b8cf9cde616dda3fb3d1d3cce5bb0cd60b/webdriver-spec.html#L2300)):*
<details><summary><i>Previously, jgraham wrote…</i></summary><blockquote>

I don't think these notes should be here, but higher up as a general note that the structure of this algorithm is based on a premise that isn't true in real implementations.
</blockquote></details>

I'll move this note up.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/webdriver/755)*

shs96c:
Review status: 0 of 1 files reviewed at latest revision, 9 unresolved discussions.

---

*[webdriver-spec.html, line 2290 at r2](https://reviewable.io:443/reviews/w3c/webdriver/755#-KcsMnS1KIp7FSibnViV:-KdRs8H8o20PANLJHEyU:b-93jjw1) ([raw file](https://github.com/w3c/webdriver/blob/f0baef35c9edbb660058789f4f998fa3d99fa588/webdriver-spec.html#L2290)):*
<details><summary><i>Previously, shs96c (Simon Stewart) wrote…</i></summary><blockquote>

We followed the CSS spec for protocol extensions, and they don't require extensions to use the vendor prefix --- https://www.w3.org/TR/CSS21/syndata.html#vendor-keywords If we're doomed, at least it'll be a party with plenty of company.

I can add a note in the protocol extensions section to say that no spec-supplied capability name will ever have a colon in it.
</blockquote></details>

I've now linked all of these to extension capabilities. The definition of that says that capability keys must have a ":". I think that properly addresses your concern.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/webdriver/755)*

jgraham:
Reviewed 1 of 1 files at r6.

---

*[webdriver-spec.html, line 2227 at r3](https://reviewable.io:443/reviews/w3c/webdriver/755#-Kd1IAFG9vvfpBaU9SXM:-KdW0e5jEhV5BCXZJvhV:b-4cboha) ([raw file](https://github.com/w3c/webdriver/blob/48e327c0ddc189763820a1c808ecbb5f126be925/webdriver-spec.html#L2227)):*
<details><summary><i>Previously, shs96c (Simon Stewart) wrote…</i></summary><blockquote>

I can see the argument for a lack of interoperability. I guess if the validation step only checked that deserialisation was possible we'd still have the raw JSON blob at this point. I'd be amazed if most libraries for JSON in any language didn't have some mechanism for verifying that two blobs are equivalent.

I'll remove the equality check, but I think this will cause problems for users further down the line since local end bindings are written by folks not used to reading specs.

</blockquote></details>

Well it seems easier to allow this in the future if there's a problem than the reverse.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/webdriver/755)*

PR #756, opened 2017-02-10
Move Execute Script to a promise based approach. Fixes #381

### LINKED ISSUE #381: Should be possible to return errors with executeAsyncScript
https://www.w3.org/Bugs/Public/show_bug.cgi?id=28060

jleyba:

> The callback provided to the executeAsyncScript function only accepts a single argument that is always treated as a successful completion. It should be possible for users to call this function with an error to indicate their script failed.
> 
> Node.js has popularized the "Error-first" callback approach: errors are passed as the first argument, successful values the second.
> 
> Another option would be to standardize on the Error-type. If the callback is invoked with an instanceof Error, the script is marked as a failure.

AutomatedTester:
This moves Execute Script to be Promise based and removes the
need for execute script async. The patch also updates the end
points to be like the Selenium JSONWP.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/756)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -288,6 +288,7 @@ var respecConfig = {
    <!-- Getting innerText --> <li><dfn><a href=http://rocallahan.github.io/innerText-spec/index.html#setting-innertext>Getting <code>innerText</code></a></dfn>
    <!-- Hidden --> <li><dfn><a href="https://html.spec.whatwg.org/#hidden-state-%28type=hidden%29">Hidden</a></dfn> state
    <!-- Image Button --> <li><dfn><a href="https://html.spec.whatwg.org/#image-button-state-%28type=image%29">Image Button</a></dfn> state
+   <!-- In parallel --> <li><dfn><a href="https://html.spec.whatwg.org/#in-parallel">In parallel</a></dfn>
    <!-- Input event applies --> <li><dfn><a href=https://html.spec.whatwg.org/#concept-input-apply><code>input</code> event applies</a></dfn>
    <!-- Input event --> <li><dfn><a href=https://html.spec.whatwg.org/#event-input><code>input</code></a> event</dfn>
    <!-- Input type file selected --> <li><dfn><a href="https://html.spec.whatwg.org/#concept-input-type-file-selected">Selected Files</a></dfn>
@@ -532,6 +533,15 @@ var respecConfig = {
    <!-- SyntaxError --> <li><dfn><a href="https://heycam.github.io/webidl/#syntaxerror"><code>SyntaxError</code></a></dfn></li>
 
   </ul>
+
+  <dt>Promises Guide <!-- Eventually this will be merged into WebIDL -->
+  <dd><p>The following terms are defined in the Promises Guide. [[!PROMISES-GUIDE]]
+    <ul>
+      <!-- a new promise --> <li><dfn><a href="https://www.w3.org/2001/tag/doc/promises-guide#a-new-promise">A new promise</a></dfn>
+      <!-- Promise calling --> <li><dfn data-lt='promise-call'><a href="https://www.w3.org/2001/tag/doc/promises-guide#promise-calling">Promise-calling</a></dfn>
+      <!-- Reject --> <li><dfn><a href="https://www.w3.org/2001/tag/doc/promises-guide#reject-promise">reject</a></dfn>
+      <!-- Resolve --> <li><dfn><a href="https://www.w3.org/2001/tag/doc/promises-guide#resolve-promise">resolve</a></dfn>
+    </ul>
 
   <dt>XML Namespaces
   <dd><p>The following terms are defined in the Namespaces in XML [[!XML-NAMES]]
@@ -2215,8 +2225,8 @@ with a "<code>moz:</code>" prefix:
        with the <code>null</code> value being set to <code>false</code>.
 
      <dt>"<code>pageLoadStrategy</code>"
-     <dd><p>If <var>capability value</var> is not a keyword in the 
-      <a>table of page load strategies</a>, return <a>error</a> with 
+     <dd><p>If <var>capability value</var> is not a keyword in the
+      <a>table of page load strategies</a>, return <a>error</a> with
       <a>error code</a> <a>invalid argument</a>.
 
       <p>Otherwise, set the "<code>pageLoadStrategy</code>" entry
@@ -5777,21 +5787,30 @@ must run the following steps:
  <li><p>Let <var>body</var> and <var>arguments</var>
   be <var>script arguments</var>’ data.
 
- <li><p>Let <var>result</var> be the result of calling <a>execute a function body</a>,
-  with arguments <var>body</var> and <var>arguments</var>.
+ <li><p>Let <var>promise</var> be <a>a new Promise</a>.
 
- <li><p>Wait for <a>execute a function body</a> to return
-  or for <a>session script timeout</a> milliseconds to expire,
-  whichever happens sooner.
+ <li><p>Run the following substeps <a>in parallel</a>:
+   <ol>
+     <li><p>Let <var>result</var> be the result of <a>promise-calling</a>
+        <a>execute a function body</a>, with arguments
+        <var>body</var> and <var>arguments</var>.
 
- <li><p>If the previous step completed due to the <a>session script timeout</a> being reached,
-  return <a>error</a> with <a>error code</a> <a>script timeout</a>.
+     <li><p>If <var>result</var> is an error,
+       <a>reject</a> <var>promise</var> with <var>result</var>.
+   </ol>
 
- <li><p>If <var>result</var> is an <a>error</a>, return result.
+  <li><p>If <var>promise</var> is still pending and
+    <a>session script timeout</a> milliseconds is reached,
+    <a>reject</a> <var>promise</var> with <a>error</a> with
+    <a>error code</a> <a>script timeout</a>.
 
- <li><p>Otherwise let <var>value</var> be <var>result</var>’s data.
+  <li><p>Upon fulfillment of <var>promise</var> with value <var>v</var>,
+      return <a>success</a> with data <var>v</var>.
 
- <li><p>Return <a>success</a> with data <var>value</var>.
+  <li><p>Upon rejection of <var>promise</var> with reason <var>r</var>,
+      if <var>r</var> is an <a>error</a>, return <var>r</var>.
+      Otherwise, return <a>error</a> with <a>error code</a> <a>javascript error</a>
+      and data <var>r</var>.
 </ol>
 </section> <!-- /Execute Script -->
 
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -5780,118 +5774,19 @@
  <li><p>Let <var>result</var> be the result of calling <a>execute a function body</a>,
   with arguments <var>body</var> and <var>arguments</var>.
 
- <li><p>Wait for <a>execute a function body</a> to return
-  or for <a>session script timeout</a> milliseconds to expire,
-  whichever happens sooner.
+ <li><p>Let <var>promise</var> be a new Promise.
```
shs96c: `Promise` should be defined as a dependency and link to https://tc39.github.io/ecma262/#sec-promise-constructor

webdriver-spec.html
```diff
@@ -5780,118 +5774,19 @@
...
 
- <li><p>If the previous step completed due to the <a>session script timeout</a> being reached,
-  return <a>error</a> with <a>error code</a> <a>script timeout</a>.
-
- <li><p>If <var>result</var> is an <a>error</a>, return result.
+ <li><p>Let <var>result</var> be the result of <a>execute a function body</a>.
+   If <var>result</var> is an <a>error</a> or if for <a>session script timeout</a>
```
shs96c: `if for`?

shs96c: Timeout should probably be specified as waiting in an implementation-specific way for the promise state to no longer be "pending"

webdriver-spec.html
```diff
@@ -1259,17 +1259,11 @@
 
  <tr>
   <td>POST</td>
-  <td>/session/{<var>session id</var>}/execute/sync</td>
+  <td>/session/{<var>session id</var>}/execute</td>
   <td><a>Execute Script</a></td>
  </tr>
 
  <tr>
-  <td>POST</td>
```
shs96c: Async execute should remain. The difference between `execute` and `execute async` is that the script for `execute async` would get access to the resolve function, as per @jleyba's suggestion here: https://github.com/w3c/webdriver/issues/381#issuecomment-276274426

webdriver-spec.html
```diff
@@ -5780,118 +5774,19 @@
  <li><p>Let <var>result</var> be the result of calling <a>execute a function body</a>,
   with arguments <var>body</var> and <var>arguments</var>.
```
andreastt: We appear to execute twice?

webdriver-spec.html
```diff
@@ -5780,118 +5774,19 @@
...
- <li><p>Wait for <a>execute a function body</a> to return
-  or for <a>session script timeout</a> milliseconds to expire,
-  whichever happens sooner.
+ <li><p>Let <var>promise</var> be a new Promise.
 
- <li><p>If the previous step completed due to the <a>session script timeout</a> being reached,
-  return <a>error</a> with <a>error code</a> <a>script timeout</a>.
-
- <li><p>If <var>result</var> is an <a>error</a>, return result.
+ <li><p>Let <var>result</var> be the result of <a>execute a function body</a>.
```
andreastt: I think the intention is to execute the injected code inside the promise, and resolve the promise to return the result?

```js
  new Promise((resolve, reject) => {
    try {
      let fn = new Function(userScript);
      resolve(fn(...userArgs));
    } catch (ex) {
      reject(ex);
    }
  }).then(onSuccess, onFailure);
```

webdriver-spec.html
jleyba: I'm actually in favor of removing `execute async`, it's totally unnecessary with promise support. I showed how it could be implemented in terms of `execute` should others want to keep it.

webdriver-spec.html
```diff
@@ -5780,118 +5774,19 @@
...
+   If <var>result</var> is an <a>error</a> or if for <a>session script timeout</a>
+   milliseconds to expire,
+   reject <var>promise</var> else
+   resolve <var>promise</var> with <var>result</var>.
 
  <li><p>Otherwise let <var>value</var> be <var>result</var>’s data.
```
jleyba: These last two steps depend on the promise's resolved value. If resolved, return success. If rejected, return `javascript error`.

webdriver-spec.html
jleyba: Is it expected that the following would all eventually trigger a timeout error?

1. The window is closed before the promise is resolved
2. A new page is loaded before the promise is resolved
3. A modal dialog (alert, prompt, etc) is opened

webdriver-spec.html
AutomatedTester: speaking to annevk, we shouldnt put anything at the moment as there is a missing dependency in WebIDL. Promises are "assumed knowledge"

AutomatedTester:
I have rewritten this and used the promises guide from the TAG.

webdriver-spec.html
```diff
@@ -5777,19 +5778,25 @@
...
 
- <li><p>If <var>result</var> is an <a>error</a>, return result.
+     <li><p>If <var>result</var> is an <a>error</a>,
+       reject <var>promise</var> with <var>result</var>.
 
- <li><p>Otherwise let <var>value</var> be <var>result</var>’s data.
+     <li><p>Otherwise resolve <var>promise</var> with <var>result</var>.
+   </ol>
+
+ <li><p>Let <var>value</var> be <var>result</var>’s data.
```
jleyba: These last two steps should be written in terms of the resolved promise. Maybe something like:

```html
<li>Upon fulfillment of <var>promise</var> with value <var>v</var>,
    return <a>success</a> with data <var>v</var>.

<li>Upon rejection of <var>promise</var> with reason <var>r</var>,
    if <var>r</var> is an <a>error</a>, return <var>r</var>.
    Otherwise, return <a>error</a> with <a>error code</a> <a>javascript error</a>
    and data <var>r</var>.
```


This follows recommendations in https://www.w3.org/2001/tag/doc/promises-guide#shorthand-reacting something like:

webdriver-spec.html
```diff
@@ -5834,36 +5841,25 @@
...
+      <a>session script timeout</a> milliseconds is reached,
+     reject <var>promise</var> with <a>error</a> with
+     <a>error code</a> <a>script timeout</a>.
 
- <li><p>Otherwise, let <var>result</var> be <var>webdriver callback result</var>’s data.
+    <li><p>If <var>result</var> is an <a>error</a>,
+      reject <var>promise</var> with <var>result</var>.
 
- <li><p>If <var>result</var> is an <a>error</a>, return <var>result</var>.
+    <li><p>Otherwise resolve <var>promise</var> with <var>result</var>.
```
jleyba: Remove this step. `resolve` was provided as the last argument, so the function body is responsible for calling it to resolve the promise

webdriver-spec.html
```diff
@@ -5777,19 +5778,25 @@
...
 
- <li><p>If the previous step completed due to the <a>session script timeout</a> being reached,
-  return <a>error</a> with <a>error code</a> <a>script timeout</a>.
+     <li><p>If <var>promise</var> is still pending and
+       <a>session script timeout</a> milliseconds is reached,
+      reject <var>promise</var> with <a>error</a> with
+      <a>error code</a> <a>script timeout</a>.
 
- <li><p>If <var>result</var> is an <a>error</a>, return result.
+     <li><p>If <var>result</var> is an <a>error</a>,
```
jleyba: This sounds like you're saying:

```js
new Promise((resolve, reject) => {
  let result = fn();
  if (result instanceof Error) {
    reject(result);
  }
});
```

Instead of

```js
new Promise((resolve, reject) => {
  try {
    let result = fn();
  } catch (ex) {
    reject(ex);
  }
});

```

webdriver-spec.html
```diff
@@ -5834,36 +5841,25 @@
...
- <li><p>If the previous step completed due the <a>session script timeout</a> being reached,
-  set <var>webdriver callback result</var> to <code>expired</code>
-  and return <a>error</a> with <a>error code</a> <a>script timeout</a>.
+    <li><p>If <var>promise</var> is still pending and
+      <a>session script timeout</a> milliseconds is reached,
+     reject <var>promise</var> with <a>error</a> with
+     <a>error code</a> <a>script timeout</a>.
 
- <li><p>Otherwise, let <var>result</var> be <var>webdriver callback result</var>’s data.
+    <li><p>If <var>result</var> is an <a>error</a>,
```
jleyba: Same comment as synchronous script: sounds like you're requiring an instanceof check instead of a try-catch

webdriver-spec.html
```diff
@@ -5834,36 +5841,25 @@
...
+      reject <var>promise</var> with <var>result</var>.
 
- <li><p>If <var>result</var> is an <a>error</a>, return <var>result</var>.
+    <li><p>Otherwise resolve <var>promise</var> with <var>result</var>.
+  </ol>
```
jleyba: Same comment as `executeScript`, the final steps for returning a response should be in terms of the resolved promise

webdriver-spec.html
```diff
@@ -5777,19 +5778,25 @@
  <li><p>Let <var>body</var> and <var>arguments</var>
   be <var>script arguments</var>’ data.
 
- <li><p>Let <var>result</var> be the result of calling <a>execute a function body</a>,
-  with arguments <var>body</var> and <var>arguments</var>.
+ <li><p>Let <var>promise</var> be a new Promise.
```
jleyba: Should "a new Promise" link to https://www.w3.org/2001/tag/doc/promises-guide#intro?

webdriver-spec.html
```diff
@@ -5777,19 +5778,25 @@
...
-  or for <a>session script timeout</a> milliseconds to expire,
-  whichever happens sooner.
+ <li><p>Run the following substeps <a>in parallel</a>:
+   <ol>
+     <li><p>Let <var>result</var> be the result of calling <a>execute a function body</a>,
+      with arguments <var>body</var> and <var>arguments</var>.
 
- <li><p>If the previous step completed due to the <a>session script timeout</a> being reached,
-  return <a>error</a> with <a>error code</a> <a>script timeout</a>.
+     <li><p>If <var>promise</var> is still pending and
```
jleyba: This step should run in parallel with function execution, so it should it should be part of the parent list.

1. Let `promise` be a new Promise.
2.  Run the following steps in parallel:
      1. Let `result` be the result of calling execute a function body...
      2. If `result` is an error...
      3. Otherwise resolve `promise` with `result`
3. If `promise` is still pending and `session script timeout` milliseconds is reached...

I might even put the timeout in its own parallel track:


1. Let `promise` be a new Promise.
2.  Run the following steps in parallel:
      1. Let `result` be the result of calling execute a function body...
      2. If `result` is an error...
      3. Otherwise resolve `promise` with `result`
3. Run the following steps in parallel:
      1. Wait `session script timeout` milliseconds
       2. Reject `promise` with an `error` with `error code` `script timeout`

webdriver-spec.html
```diff
@@ -5777,19 +5778,25 @@
...
- <li><p>Let <var>result</var> be the result of calling <a>execute a function body</a>,
-  with arguments <var>body</var> and <var>arguments</var>.
+ <li><p>Let <var>promise</var> be a new Promise.
 
- <li><p>Wait for <a>execute a function body</a> to return
-  or for <a>session script timeout</a> milliseconds to expire,
-  whichever happens sooner.
+ <li><p>Run the following substeps <a>in parallel</a>:
+   <ol>
+     <li><p>Let <var>result</var> be the result of calling <a>execute a function body</a>,
```
jleyba: Either this, or the `execute a function body` section (GitHub won't let me add a comment to an unmodified line...) should be reworded to use `promise-calling`:

https://www.w3.org/2001/tag/doc/promises-guide#promise-calling

webdriver-spec.html
AutomatedTester: According to annevk, we need to wait for WebIDL to get their act together.

webdriver-spec.html
AutomatedTester: updated locally

webdriver-spec.html
AutomatedTester: done locally

webdriver-spec.html
AutomatedTester: done locally.

webdriver-spec.html
AutomatedTester: done locally.

AutomatedTester:
I think that is all the comments addressed so far. Ready for further review.

AutomatedTester:
My original patch removed it but @shs96c wanted it back. Happy to let y'all hash that out.

webdriver-spec.html
```diff
@@ -5777,21 +5787,32 @@
...
+        <a>execute a function body</a>, with arguments
+        <var>body</var> and <var>arguments</var>.
 
- <li><p>If <var>result</var> is an <a>error</a>, return result.
+     <li><p>If <var>result</var> is an error,
+       <a>reject</a> <var>promise</var> with <var>result</var>.
+   </ol>
+   <li><p>Run the following substeps <a>in parallel</a>:
+     <ol>
+       <li><p>If <var>promise</var> is still pending and
```
shs96c: There's only one step here. Doesn't matter whether this is in parallel or in serial. Should this be part of the list above?

AutomatedTester: @jleyba suggested that it be out on its own. see https://github.com/w3c/webdriver/pulls#discussion_r100877514

shs96c: It's fine for it to be on its own, but this construct looks weird. Just let it be a step on its own without the "run in parallel" stuff, as there's nothing its running in parallel with.

PR #759, opened 2017-02-10
`Actions`: Shuffle content around

shs96c:
This places the definition of the generic "input state"
term at the top of the section, then proceeds to list
the specialisations before explaining what the "input
state table" and "input cancel list" are. This feels
like a more sensible ordering for this section of the
spec.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/759)


shs96c:
@vkatsikaros, this is part of the clean up for #691

PR #761, opened 2017-02-10
`Actions`: Track the known input sources

shs96c:
This allows us to properly validate the parameters given
to an action sequence.

This diff also causes webdriver implementations to create
new input sources as necessary. #progress

Address issue #691

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/761)


shs96c:
@vkatsikaros Thought you might be interested in tracking this one too.

PR #768, opened 2017-02-14
Return point for all elements not just HTML Element. Fixes #392

### LINKED ISSUE #392: "calculate the absolute position" uses offset* for non-html Elements
https://www.w3.org/Bugs/Public/show_bug.cgi?id=29500

Ms2ger:

> offsetParent is only defined for HTMLElement

AutomatedTester:
This allows us to be able to get the value from other elements,
like SVG Elements, instead of just working with HTMLElements

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/768)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -473,6 +473,7 @@ var respecConfig = {
  <dd>The following functions are defined in
   the CSSOM View Module: [[!CSSOM-VIEW]]:
   <ul>
+   <!-- getBoundingClientRect() --> <li><dfn><a href="https://drafts.csswg.org/cssom-view/#dom-element-getboundingclientrect">getBoundingClientRect</a></dfn>
    <!-- elementFromPoint --> <li><dfn>Element from point</dfn> as <a href="https://drafts.csswg.org/cssom-view/#dom-document-elementfrompoint">elementFromPoint()</a>
    <!-- elementsFromPoint --> <li><dfn data-lt="paint order">Elements from point</dfn> as <a href=https://drafts.csswg.org/cssom-view/#dom-document-elementsfrompoint>elementsFromPoint()</a>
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -2219,8 +2222,8 @@
        with the <code>null</code> value being set to <code>false</code>.
 
      <dt>"<code>pageLoadStrategy</code>"
-     <dd><p>If <var>capability value</var> is not a keyword in the 
-      <a>table of page load strategies</a>, return <a>error</a> with 
+     <dd><p>If <var>capability value</var> is not a keyword in the
```
shs96c: This should be in a different diff.

webdriver-spec.html
```diff
@@ -6445,7 +6452,7 @@
 source</a> is added to the list of <a>active input sources</a>, a
 corresponding entry is made in the <a>input state table</a> where the
 key is the <a>input source</a>'s <code>input id</code> and the value is
-the <a>input source</a>'s <a>input source state</a>. When an 
+the <a>input source</a>'s <a>input source state</a>. When an
```
shs96c: Unrelated to this diff.

AutomatedTester:
Could you elaborate how this is different behaviour to what currently doing.

getBoundingClientRect returns the `x` and `y` of the viewport. We then add the `scrollX` and `scrollY`. Since `scroll{X,Y}` `Returns the number of pixels that the document has already been scrolled {vertically, horizontally}` it will return the value from the top left of the document.

shs96c:
We used to return the size of the element on the DOM. If we're larger than the current viewport, the old version would return a different value to this new one. As written, I don't think this even attempts to scroll the element into view, so it'll give surprising results quite often.

PR #769, opened 2017-02-14
`Processing Capabilities`: don't include null values when merging capabilities

shs96c:
This simplifies some of the later logic and means
that all the fields of the returned result have
defined values.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/769)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -2091,27 +2091,51 @@ with a "<code>moz:</code>" prefix:
  an <a>endpoint node</a> must take the following steps:
 
 <ol>
- <li><p>Let <var>result</var> be a JSON <a>Object</a>
-  with the same entries as <var>primary</var>.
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -2081,27 +2081,53 @@
...
+    property</a> named <var>name</var> from <var>secondary</var>.
+
+   <li><p>Let <var>primary value</var> be the result of
+    <a>getting a property</a> named <var>name</var> from
+    <var>primary</var>.
+
+   <li><p>Run the substeps of the first matching condition:
+    <dl class=switch>
+     <dt><var>value</var> is <a><code>null</code></a>
+     <dd>Do nothing.
```
AutomatedTester: we don't need this line.

We can have
```html
<dt>foo
<dt>bar
<dd>I foo bar like cheese
```

PR #773, opened 2017-02-15
Deserializations

shs96c:
These diffs shouldn't be squashed into a single one.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/773)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -1995,6 +1995,37 @@ with a "<code>moz:</code>" prefix:
  </tr>
 </table>
 
+<p>The <a>remote end</a> steps to <dfn>deserialize as a proxy</dfn>
+ argument <var>parameter</var> are:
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -1983,6 +1983,32 @@
...
+ argument <var>parameter</var> are:
+
+<ol>
+ <li><p>If <var>parameter</var> is not a JSON <a>Object</a> return
+  an <a>error</a> with <a>error code</a> <a>invalid argument</a>.
+
+ <li><p>Let <var>proxy</var> be a new, empty
+  <a>proxy configuration object</a>.
+
+ <li><p>For each <var>key</var> and <var>value</var> coressponding
```
AutomatedTester: this needs to be described as `enumerable`

Issue #765, opened 2017-02-13
Resize/reposition capability


jugglinmike:
Chapter [10. Command Contexts](https://w3c.github.io/webdriver/webdriver-spec.html#resizing-and-positioning-windows#h-command-contexts), section [10.7 Resizing and Positioning Windows](https://w3c.github.io/webdriver/webdriver-spec.html#resizing-and-positioning-windows#h-resizing-and-positioning-windows) reads:

> WebDriver provides commands for interacting with the operating system window
> containing the current top-level browsing context. Because different
> operating system’s window managers provide different abilities, not all of
> the commands in this section can be supported by all remote ends. Where a
> command is not supported, an unsupported operation error is returned.

It may be that an application developer wishes to test some behavior based on window resizing/repositioning. Those tests are irrelevant in some contexts, but allowing failure on a test-by-test basis (by conditionally accepting the "unsupported operation" error as it is returned) can be cumbersome and tends to clutter test bodies.

If the developer could instead formally request this as a capability when starting the session, then the relevant branching could happen in advance of the tests themselves.

So I'm wondering: would it make sense to extend the Capabilities API with properties describing support for these operations?

shs96c:
It probably makes sense, and this follows in the same footsteps as `acceptInsecureCerts`.

andreastt:
I don’t understand how this is similar to `acceptInsecureCerts`. Configuring a session to

shs96c:
Exactly. And a test that relies on that capability shouldn't have to wait until an `unsupported operation` error comes back to find out. Because it's a capability that might not be supported, it should be possible to require it explicitly when setting up the session. As you point out, some browsers that normally support resizing may not depending on the system they're running on (there are tile-based WMs that prohibit resizing to arbitrary sizes, for example)

That's exactly like the case with `acceptInsecureCerts` where a user requests this capability, and a remote end may decide, for whatever reason, to support it or not.

andreastt:
OK, it sounds like we’re on the same page!

> As you point out, some browsers that normally support resizing may not depending on the system they're running on (there are tile-based WMs that prohibit resizing to arbitrary sizes, for example)

FWIW I don’t think that’s something we can test for on a browser level. The only realistic check would be if you needed to `alwaysMatch` on a browser that was resizable and the remote end was a mobile browser that already knew the characteristics of the system it was running on.

shs96c:
[...]

PR #775, opened 2017-02-15
Introduce capability: `setWindowRect`

jugglinmike:
I initially attempted to define this capability in terms of, "querying/modifying the top-level browsing context's window rect." That language is not quite accurate, though, since it doesn't necessarily describe "Minimize Window." So that's why I'm suggesting referencing "the commands in Resizing and Positioning Windows" instead. It's less precise, and I couldn't find any precedent for it, so I'm wondering if there's a better way.

I believe this exposes the following behavior, given the listed "capabilities" objects:

- `{ "setWindowRect": true }`: user agent MUST support *every* command in
  "Resizing and Positioning Windows"
- `{ "setWindowRect": false }`: user agent MUST NOT support *one or more*
  commands in "Resizing and Positioning Windows"
- `{}`: user agent MAY support all commands in "Resizing and Positioning
  Windows"

...but I would appreciate confirmation.

This is intended to resolve gh-765.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/775)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -1812,6 +1812,14 @@ with a "<code>moz:</code>" prefix:
  </tr>
 
  <tr>
+  <td><dfn data-lt="window dimensioning/positioning">Window dimensioning/positioning</dfn>
+  <td>"<code>setWindowRect</code>"
+  <td>boolean
+   <td>Indicates whether the remote end supports all of the <a>commands</a> in
+   <a href=#h-resizing-and-positioning-windows>Resizing and Positioning Windows</a>.
+ </tr>
+
+ <tr>
   <td><dfn>Session timeouts configuration</dfn>
   <td>"<code>timeouts</code>"
   <td>JSON <a>Object</a>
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -2261,6 +2273,14 @@
       <p>Otherwise, set the "<code>proxy</code>" entry
        in <var>matched capabilities</var> to <var>capability value</var>.
 
+     <dt>"<code>setWindowRect</code>"
+     <dd><p>If <var>capability value</var> is not a boolean
```
AutomatedTester: Since we can't really influence if it can be done that way or not we should just ignore what the user puts here

jugglinmike: I think I'm misunderstanding how this algorithm is used, then. I thought the
effect of this would be an "invalid argument" error if the "New Session" command
was executed with a body like:

```json
{
  "setWindowRect": "I made a mistake and specified a string instead of a boolean"
}
```

...which seems to me like a good way to avoid confusion in ambiguous
circumstances. But based on your feedback, I'm thinking this is *not* the
actual effect of this language. Could you confirm?

AutomatedTester: No you are understanding it correctly.

But if I send through
```
{
  "setWindowRect": "I made a mistake and specified a string instead of a boolean"
}
```
or
```
{
  "setWindowRect": False
}
```

The driver is just going to overwrite it and not going to care about what is in it. If some sends through 
```
{
  "setWindowRect": false
}
```

We can't guarantee that we can't manipulate the page. If it's set to false but some clever clogs might have buttons to go fullscreen and we can't block that.

jugglinmike:
Thanks, David!

PR #781, opened 2017-02-18
`Actions`: Remove duplicated input state check

shs96c:
By the time we're processing pointer actions we've
already been through the steps of processing an action
sequence. Within _those_ steps we've already checked
that the input state within the `input state table`
correctly matches expected type for the pointer action.

Address issue #691

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/781)


shs96c:
@vkatsikaros, this is one of the final steps in the clean up of #691. Thought you might be interested in tracking it.

PR #787, opened 2017-02-20
use key of 'pageLoad' instead of 'page load' in timeout capability

titusfortner:
Almost every other key in the spec uses camel case instead of spaces. This makes it easy in Ruby to implement with immutable symbols. Having a space in a key requires it to be a string. I can obviously make a special translation method if this needs to have a space, but it would be great if that could be avoided.

The only other keys I see that have spaces are the locator strategies, but those are easily handled in the implementation abstraction rather than being part of a user-facing object (like `Capabilities`).

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/787)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -2596,9 +2596,9 @@ with a "<code>moz:</code>" prefix:
       to the value of property "<code>implicit</code>".
       Otherwise, set the <a>session implicit wait timeout</a> to 0 (zero) milliseconds.
 
-     <li><p>If <var>timeouts</var> has a numeric property with key "<code>page load</code>",
+     <li><p>If <var>timeouts</var> has a numeric property with key "<code>pageLoad</code>",
       set the <a>current session</a>’s <a>session page load timeout</a>
-      to the value of property "<code>page load</code>".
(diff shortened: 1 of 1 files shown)


shs96c:
Adding @andreastt and @jgraham since they have an implementation in geckodriver that's impacted by this change. If they approve, I'll land.

jgraham:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/webdriver/787)*

shs96c:
You can define a symbol with spaces in: `:"hello world".class` returns `Symbol`

shs96c:
I kind of wish we'd gone for spaces to make things like capability names more readable, but consistency is a Good Thing.

andreastt:
This is a backwards-incompatible change to geckodriver, and will require a few turnarounds with compatibility fixes in our various tools.

But I feel that if we want to make this change, the _session timeouts configuration_ object should also change since that still uses "`page load`".

shs96c:
I spoke with @jgraham before approving and merging this change to make sure that this didn't impact geckodriver too badly. Will update the `timeouts configuration object`

PR #792, opened 2017-02-21
Scroll into view: do so instantly

shs96c:
Avoids a potential issue with low wait timeouts and
smooth scrolling.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/792)


andreastt:
As far as I can tell, this isn’t in the CSSOM View specification for [ScrollIntoViewOptions](https://drafts.csswg.org/cssom-view/#dictdef-scrollintoviewoptions). I’m guessing this is an omission from the spec?

shs96c:
`ScrollIntoViewOptions` extend [ScrollOptions](https://drafts.csswg.org/cssom-view/#dictdef-scrolloptions), and these contain `behavior`

andreastt:
It’s also in line with what Marionette does: http://searchfox.org/mozilla-central/source/testing/marionette/element.js#975

PR #796, opened 2017-02-22
`Timeouts`: Standardise `pageLoad` rather than `page load`

shs96c:
A previous change (#787) changed `page load` to `pageLoad`
but we missed the similar change in the `timeouts configuration
object`.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/796)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -2065,7 +2065,7 @@ with a "<code>moz:</code>" prefix:
  </tr>
 
  <tr>
-  <td>"<code>page load</code>"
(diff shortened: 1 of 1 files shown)


andreastt:
It makes sense to use the JSON deserialisation of session timeouts configuration also for Set Timeouts so we don’t have two definitions of how to deserialise it.

andreastt:
I accepted this and filed https://github.com/w3c/webdriver/issues/806 about making Set Timeouts use the session timeouts configuration’s deserialisation algorithm.

PR #799, opened 2017-02-22
`Send Keys`: Allow HTML5 values to be sent

### LINKED ISSUE #535: Allow `Element Send Keys` to send HTML 5 form data
It should be possible to use the [form plain-text encoded](https://html.spec.whatwg.org/#plain-text-form-data) value of an input element to set the value of that input element using `Element Send Keys`.

shs96c:
It should be possible to use `Element Send Keys` to
set the value of INPUT elements that are rendered
as something other than a text control. We do this
by allowing users to send form-serialized values to
the element and setting the value directly.

Closes #535

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/799)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -270,7 +270,9 @@ var respecConfig = {
    <!-- Button --> <li><dfn><a href="https://html.spec.whatwg.org/#button-state-%28type=button%29">Button</a></dfn> state
    <!-- Buttons --> <li><dfn><a href=https://html.spec.whatwg.org/#concept-button>Buttons</a></dfn>
    <!-- Canvas context mode --> <li><dfn><a href=https://html.spec.whatwg.org/#concept-canvas-context-mode>Canvas context mode</a></dfn>
+   <!-- Color state --> <li><dfn><a href="https://html.spec.whatwg.org/#color-state-(type=color)"><code>color</code> state</a></dfn>
    <!-- Multiple --> <li><dfn><a href=https://html.spec.whatwg.org/#attr-input-multiple><code>multiple</code> attribute</a></dfn>
+   <!-- Mutable --> <li><dfn><a href=https://html.spec.whatwg.org/#concept-fe-mutable>Mutable</a></dfn>
    <!-- Change event --> <li><dfn><a href="https://html.spec.whatwg.org/multipage/indices.html#event-change"><code>change</code> event</a></dfn>
    <!-- Checkbox --> <li><dfn><a href="https://html.spec.whatwg.org/#checkbox-state-%28type=checkbox%29">Checkbox</a></dfn> state
    <!-- Checkedness --> <li><dfn><a href=https://html.spec.whatwg.org/#concept-fe-checked>Checkedness</a></dfn>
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -5602,33 +5605,54 @@
   of <a>trying</a> to <a>get a known element</a>
   with argument <var>element id</var>.
 
+ <li><p><a>Scroll into view</a> the <var>element</var>.
```
andreastt: I would like this to behave the way _Element Click_ does: We only need to scroll it into view if it is not already in view. You can steal the necessary steps from _Element Click_.

webdriver-spec.html
```diff
@@ -5602,33 +5605,54 @@
...
+  with <a>error code</a> <a>invalid argument</a>.
 
- <li><p>If <var>element</var> is not the <a>active element</a>:
+ <li><p>If <var>element</var> is not the <a>active element</a> run
+  the <a>focusing steps</a> for the <var>element</var>.
+
+ <li><p>If the user agent renders <var>element</var> as
+  something other than a text control (for example
+  an <a><code>input</code></a> element in the <a><code>color</code>
+  state</a> being presented as a colorwheel):
```
andreastt: I think we should enumerate what controls the implementation should check.

Also we need a normative reference to what a “text control” is. I think `<input>` elements that are in the _text_, _search_, _tel_, _url_, _email_, and _password_ states should cover it.

webdriver-spec.html
shs96c: I may just change the definition of `scroll into view` to handle the case where the element is already in view.

webdriver-spec.html
shs96c: The problem is that the list is unconstrained --- there might be new HTML5 input types at any point. Worse, inputs that are normally text-based (eg. `tel`) might not always be so (eg. on mobile, perhaps `tel` is selected by opening up the contacts on the phone and picking one of them)

Given that's the case, the best we can do is this somewhat wooly language. Neither @jgraham or I could come up with some spec-ese that covered this case properly, and this seemed like a reasonable compromise.

webdriver-spec.html
```diff
@@ -5602,33 +5605,54 @@
...
 
   <ol>
-   <li><p>Run the <a>focusing steps</a> for the <var>element</var>.
+   <li><p>If <var>element</var> does not have an <a>own property</a>
+    named <code>value</code> return an <a>error</a> with <a>error
+    code</a> <a>element not interactable</a>
+
+   <li><p>If <var>element</var> is not <a>mutable</a> return
+    an <a>error</a> with <a>error code</a> <a>element not
+    interactable</a>.
```
shs96c: We don't have that as an error code in this specification, and the intent seems the same.

webdriver-spec.html
```diff
@@ -5602,33 +5605,54 @@
...
- <li><p>Let <var>text</var> be the result of <a>getting a property</a>
-    <code>text</code> from the <var>parameters</var> argument.
+ <li><p>Let <var>current text length</var> be
+  the <var><a>element</a></var>’s <a data-lt="node
+  length">length</a>.
 
- <li><p>If <var>text</var> is not a string, return an <a>error</a>
-  with <a>error code</a> <a>invalid argument</a>.
+ <li><p>Set the text insertion caret using <a>set selection range</a>
+  with using <var>current text length</var> as the 2 parameters passed in.
```
shs96c: Will fix.

webdriver-spec.html
andreastt: That is a better idea, thanks!

webdriver-spec.html
andreastt: OK.

webdriver-spec.html
andreastt: > The problem is that the list is unconstrained --- there might be new HTML5 input types at any point.

Yes, at which point our spec also needs to change.

I think a better long-term solution would be to somehow persuade HTML to put in a definition for HTML5/advanced input controls that might be rendered as textual if it is not implemented by the user agent.

> Worse, inputs that are normally text-based (eg. tel) might not always be

That is a good point: From a spec point of view, we don’t know upfront what sort of IME might be triggered in the UA. When you press `<input type=tel>` on a mobile phone, you might be presented with a dial screen or a phone book widget.

Given these realisations, I wonder if you need to be even _more_ woolly in how you phrase this. I don’t think “something other than a text control” covers the last example.

webdriver-spec.html
andreastt: I stand corrected.

webdriver-spec.html
shs96c: #809

PR #800, opened 2017-02-22
`Navigation`: use `trying` when waiting for navigation.

### LINKED ISSUE #797: Waiting for navigation to complete
There are a few problems with the algorithm for [waiting for navigation](file:///Users/shs/src/selenium/spec/webdriver-spec.html#dfn-wait-for-navigation-to-complete) to complete:

 * The timer for the page loading timeout should be started in step 1, and not after the request has matured.
 * Does not return `success` with data `null`, and so cannot be used with `trying`
 * Calls to `wait for navigation to complete` should use `trying` to be consistent with the rest of the spec.

shs96c:
Closes #797

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/800)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -2996,29 +2996,37 @@ with a "<code>moz:</code>" prefix:
  run the following steps:
 
 <ol>
- <li><p>If the <a>current session</a>'s <a>page loading strategy</a>
-  is not <code>"none"</code> and if there is an ongoing attempt to
-  <a>navigate</a> the <a>current browsing context</a> that has not
-  yet <a>matured</a>, wait for navigation to complete.
+ <li><p>If the <a>current session</a> has a <a>page loading
+  strategy</a> of <a>none</a> return <a>success</a> with
+  data <a><code>null</code></a>.
 
- <li><p>If the <a>current session</a> has a <a>page loading strategy</a>:
+ <li><p>Start a <var>timer</var>. If this algorithm has not
+  completed before <var>timer</var> reaches
+  the <a>session</a>'s <a>session page load timeout</a> in
+  milliseconds, return an <a>error</a> with <a>error
+  code</a> <a>timeout</a>.
 
-  <ol>
-   <li><p>Let <var>readiness target</var> be the <a>document readiness</a> state
-    tied to the <a>current session</a>’s <a>page loading strategy</a>,
-    which can be found in the <a>table of page load strategies</a>.
+ <li><p>If there is an ongoing attempt to <a>navigate</a>
+  the <a>current browsing context</a> that has not
+  yet <a>matured</a>, wait for navigation to <a>mature</a>.
 
-   <li><p>Wait for the the <a>current browsing context</a>’s
-    <a>document readiness</a> state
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -2995,12 +2995,19 @@
  run the following steps:
 
 <ol>
+ <li><p>Start a <var>timer</var>. If this algorithm has not
+  completed before <var>timer</var> reaches
+  the <a>session</a>'s <a>session page load timeout</a> in
+  milliseconds, return an <a>error</a> with <a>error
+  code</a> <a>timeout</a>.
```
andreastt: If the script timeout is set to null, it will evaluate the injected script forever or until it resolves/returns. Should we allow the same for the page load duration?

webdriver-spec.html
```diff
@@ -2995,12 +2995,19 @@
...
+
  <li><p>If the <a>current session</a>'s <a>page loading strategy</a>
-  is not <code>"none"</code> and if there is an ongoing attempt to
+  is not "<code>none</code>" and if there is an ongoing attempt to
```
andreastt: This should not be inside `<code>` and quotes. Replace it with `<a>none</a>` since it’s a definition in the spec.

webdriver-spec.html
```diff
@@ -2995,12 +2995,19 @@
...
   <a>navigate</a> the <a>current browsing context</a> that has not
-  yet <a>matured</a>, wait for navigation to complete.
+  yet <a>matured</a>, wait for navigation to <a>mature</a>.
```
andreastt: The mature before… what? I’d say that if there is a previous navigation attempt ongoing, we should return an error or cancel it. This is how navigating to another document using the address bar works, or how clicking the back button whilst a document is loading works.

webdriver-spec.html
```diff
@@ -2995,12 +2995,19 @@
...
 
- <li><p>If the <a>current session</a> has a <a>page loading strategy</a>:
+ <li><p>If the <a>current session</a> has a <a>page loading
+  strategy</a> that is not "<code>none</code>":
```
andreastt: Same here, regarding `<a>none</a>`.

webdriver-spec.html
```diff
@@ -2995,12 +2995,19 @@
...
 
   <ol>
    <li><p>Let <var>readiness target</var> be the <a>document readiness</a> state
```
andreastt: Mind replacing “tied to” with “associated with” here. It would save me from opening a PR.

webdriver-spec.html
shs96c: Good point.

webdriver-spec.html

webdriver-spec.html
shs96c: Cancelling it seems unwise, since the navigation could have been started by a click issued by webdriver --- we use this algorithm in the `Processing Model`, `Element Click` as well as `Go` and `Refresh`.

webdriver-spec.html

webdriver-spec.html

PR #809, opened 2017-02-24
Make scroll into a view a no-op if the element is in view

shs96c:
This simplifies some of the language in the rest of
the spec by reducing boilerplate.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/809)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -4108,11 +4108,16 @@ with a "<code>moz:</code>" prefix:
  <li><p>Otherwise return false.
 </ol>
 
-<p>An <a>element</a> is
- <dfn data-lt="scroll into view|scrolls into view">scrolled into view</dfn>
- by calling <a>scrollIntoView</a>
- with the <a><code>ScrollIntoViewOptions</code></a>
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -4108,11 +4108,16 @@
...
 </ol>
 
-<p>An <a>element</a> is
- <dfn data-lt="scroll into view|scrolls into view">scrolled into view</dfn>
- by calling <a>scrollIntoView</a>
- with the <a><code>ScrollIntoViewOptions</code></a>
- <code>{behavior: "instant", block: "end", inline: "nearest"}</code>.
+<p>When required to
+<dfn data-lt="scrolled into view|scrolls into view">scroll into view</dfn>
+<a>element</a> <var>element</var>,the <a>endpoint node</a> steps are:
```
andreastt: You could just do `<var><a>element</a></var>` here so the language flows better. You can also drop “the endpoint node steps are”: it is always called from a driver implementation.

I also liked the language from before better. To me, “an element is scrolled into view” sounds better than “when required to scroll into view element element”.

shs96c: Done, and reworded.

PR #811, opened 2017-02-24
Look up proxyType case insensitively

### LINKED ISSUE #803: Should proxyType be case-insensitive?
In chapter [7. Capabilities](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-process-capabilities#h-capabilities), section [7.1 Proxy](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-process-capabilities#h-proxy):
> "pac", "noproxy", "autodetect", "system", or "manual"

These are the legal values for the "proxyType" capability. This value is currently case-sensitive. For example, Geckodriver/Marionette won't set the proxy if the capabilities specify {"proxyType": "MANUAL"}.

The Java Selenium bindings currently send proxyType in all-caps -- the implementation detail is that there is an enum, with values named in all-caps, that is converted to a JSON string.

Should the spec call for case-insensitive matching, or should local ends be fixed to send only lowercase strings?

shs96c:
When deserializing, be kind to folks who are using enums
in the local end and have sent the names of the proxyType
in upper case.

Closes #803

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/811)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -419,6 +419,7 @@ var respecConfig = {
   in the Infra standard: [[!INFRA]]
    <ul>
     <!-- ASCII lower alpha --> <li><dfn><a href=https://infra.spec.whatwg.org/#ascii-lower-alpha>ASCII lower alpha</a></dfn>
+    <!-- ASCII lowercase --> <li><dfn><a href=https://infra.spec.whatwg.org/#ascii-lowercase>ASCII lowercase</a></dfn>
    </ul>
 
  <dt>Page Visibility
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -2022,6 +2022,9 @@
    <li><p>Let <var>value</var> be the result of <a>getting a
     property</a> named <var>name</var> from <var>parameter</var>.
 
+   <li><p>Let <var>key</var> be the result of
+    lower-casing <var>key</var>.
```
andreastt: `<a>ASCII lower alpha</a>`, which is from the Infra spec: https://infra.spec.whatwg.org/#ascii-lower-alpha

shs96c: Perfect. Thank you.

PR #812, opened 2017-02-24
Attempt to improve the langauge around setting a proxy

### LINKED ISSUE #786: Improve language around setting proxies
From the reviews of #755 it seems clear that a better definition of "being able to set a proxy" is required.

As defined in the spec, a proxy is a complex object and browsers might support only some of the keys and some of the values provided. When constructing a new session, under what conditions do we consider a proxy a match?

 * If some of the keys are supported, but the values may be ignored?
 * If all keys are all supported, but the values may or may not be ignored?
 * If all keys are all supported, and their values pass the browser's own validity checks?
 * The exact keys and values provided will lead to a working configuration?

From a local end point-of-view, only the last one has the most value --- what's the point of setting a proxy if one isn't used? From an end-node implementation's point-of-view, there may be no way of determining this until the first HTTP request is made through the browser.

shs96c:
Closes #786

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/812)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -2404,9 +2404,16 @@ with a "<code>moz:</code>" prefix:
         information to the <a>local end</a>.
 
      <dt>"<code>proxy</code>"
-     <dd><p>If the <a>endpoint node</a> does not support the proxy
-       configuration defined in <var>value</var>,
-       return <a>success</a> with data <a><code>null</code></a>.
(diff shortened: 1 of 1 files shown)


jgraham:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/webdriver/812)*

PR #820, opened 2017-02-28
Intermediary Nodes: it's useful to define upstream and downstream

shs96c:
What's in the spec for intermediary nodes isn't sufficient to
write one that correctly handles the webdriver protocol. We're
going to need some better definitions, and these will be
simpler to write if we've already defined what "upstream" and
"downstream" mean.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/820)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -707,9 +707,15 @@ var respecConfig = {
 
 <dl>
  <dt><dfn data-lt="intermediary nodes">Intermediary node</dfn>
- <dd>Intermediary nodes are those that act as proxies,
-  implementing both the client and server sides of the <a href=#protocol>protocol</a>.
-  However they are not expected to implement <a>remote end steps</a> directly.
+ <dd>Intermediary nodes are those that act as proxies, implementing
+  both the <a>local end</a> and <a>remote end</a> of
+  the <a href=#protocol>protocol</a>.  However they are not expected
+  to implement <a>remote end steps</a> directly. All nodes between a
+  specific <a>intermediary node</a> and a <a>local end</a> are said to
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -707,9 +707,14 @@
 
 <dl>
  <dt><dfn data-lt="intermediary nodes">Intermediary node</dfn>
- <dd>Intermediary nodes are those that act as proxies,
-  implementing both the client and server sides of the <a href=#protocol>protocol</a>.
-  However they are not expected to implement <a>remote end steps</a> directly.
+ <dd>Intermediary nodes are those that act as proxies, implementing
+  both the <a>local end</a> and <a>remote ends</a> of
```
andreastt: s/remote ends/remote end/

webdriver-spec.html
```diff
@@ -707,9 +707,14 @@
...
+  the <a href=#protocol>protocol</a>.  However they are not expected
+  to implement <a>remote end steps</a> directly. All nodes between a
+  specific <a>intermediary node</a> and a <a>local end</a> are said to
+  be <dfn>downstream</dfn> of that node. Conversely, any nodes between
+  a specific <a>intermediary node</a> and an <a>endpoint node</a> are
+  said to be <dfn>upstream</dfn>.
```
andreastt: Can you add `<dfn data-lt="upstream node">` and `<dfn data-lt="downstream node">` if people want to link to our spec, please?

webdriver-spec.html
shs96c: Good catch. Fixed.

webdriver-spec.html

PR #846, opened 2017-03-06
Action example diagram should use source instead of action

vkatsikaros:
In the rest of the diagram examples the actions are horizontally
grouped by source, not action. Also, it's referenced like that in the
diagram explanation.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/846)


shs96c:
Great catch.

Issue #865, opened 2017-03-20
Command Release Actions is missing check for open current top-level browsing context


vkatsikaros:
In chapter [17. Actions](https://w3c.github.io/webdriver/webdriver-spec.html#release-actions#h-actions), section [17.6 Release Actions](https://w3c.github.io/webdriver/webdriver-spec.html#release-actions#h-release-actions) the remote end steps are missing the check 

> If the current top-level browsing context is no longer open

According to [8. Sessions](https://w3c.github.io/webdriver/webdriver-spec.html#release-actions#h-sessions):
> It is an error to send any commands before starting a session, or to continue to send commands after the session has been closed.

and 

> All commands, except New Session and Status, have an associated current session

PR #866, opened 2017-03-20
Add current top-level browsing context check to Release Actions

vkatsikaros:
A fix for https://github.com/w3c/webdriver/issues/865 (provided this is indeed missing)

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/866)


shs96c:
Good catch

Issue #867, opened 2017-03-20
Matching capabilities algorithm rejects some valid capabilities


jgraham:
In chapter [7. Capabilities](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-matching-capabilities#h-capabilities), section [7.2 Processing Capabilities](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-matching-capabilities#h-processing-capabilities):
> Otherwise
> 
>     If name is not the key of an extension capability and the remote end is an endpoint node return success with data null. 

By this point we have already validated capabilities, so it shouldn't be neccessary to do any more validation. More importantly, if you pass in e.g. `timeouts` this will cause it to be rejected since that doesn't appear anywhere on the list of capabilities above. I'm not sure if the intent is that those things don't get returned even if they were set? In any case I think the current wording is broken.

shs96c:
Fixed by #868. Thanks @jgraham

PR #868, opened 2017-03-21
Fix incorrect handling of configuration properties in new session

jgraham:
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/868)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -2397,14 +2397,10 @@ with a "<code>moz:</code>" prefix:
        will not be established.
 
      <dt><strong>Otherwise</strong>
-     <dd><p>If <var>name</var> is not the key of an
-      <a>extension capability</a> and the <a>remote end</a> is
-      an <a>endpoint node</a> return <a>success</a> with data
-      <code>null</code>.
-
-      <p>Let <var>value</var> be the result of <a>trying</a>
-       implementation-specific steps to match on <var>name</var>
-       with <var>value</var>. If the match is not
+       <dd><p>If <var>name</var> is the key of an
+      <a>extension capability</a>, <p>let <var>value</var> be the
+       result of <a>trying</a> implementation-specific steps to match
+       on <var>name</var> with <var>value</var>. If the match is not
        successful, return <a>success</a> with
        data <a><code>null</code></a>.
    </dl>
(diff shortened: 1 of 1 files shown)


jgraham:
Fixes #867

webdriver-spec.html
```diff
@@ -2670,18 +2666,27 @@
...
+    substeps of the first matching statement:
+
+       <dl class=switch>
+         <dt><var>proxy</var> is a <a>proxy configuration</a> object</dt>
+         <dd><p>Take implementation-defined steps to set the user agent proxy
+             using the extracted <var>proxy</var> configuration.  If the
+             defined proxy cannot be configured return <a>error</a> with
+             <a>error code</a> <a>session not created</a>.</dd>
+
+         <dt>Otherwise</dt>
```
shs96c: We'll have deserialised as a proxy by this point --- we should never see the "otherwise" branch, and doing so is an error.

webdriver-spec.html
```diff
@@ -2397,14 +2397,10 @@
...
-      an <a>endpoint node</a> return <a>success</a> with data
-      <code>null</code>.
-
-      <p>Let <var>value</var> be the result of <a>trying</a>
-       implementation-specific steps to match on <var>name</var>
-       with <var>value</var>. If the match is not
+       <dd><p>If <var>name</var> is the key of an
+      <a>extension capability</a>, <p>let <var>value</var> be the
+       result of <a>trying</a> implementation-specific steps to match
+       on <var>name</var> with <var>value</var>. If the match is not
```
shs96c: I'm not sure about this. An intermediary node should just be passing on the values without attempting a match, assuming that they're not obviously illegal (`null` or `undefined`, or if the key is not an extension capability key.

jgraham: Fortuantely intermediary nodes never run this step since they all run the "return wahtever you  like as long as it has the right output" step.

webdriver-spec.html
jgraham: We hit this if `proxy` is undefined i.e. if no proxy setting was passed in.

webdriver-spec.html
shs96c: Cool

webdriver-spec.html
shs96c: Ah. I see. You're right. We do hit this branch :) Not sure whether it's better to leave this out entirely if a proxy isn't configured or return this empty object. Your call.

Issue #835, opened 2017-03-02
We seem to lowercase the proxy object keys for no reason


jgraham:
In chapter [7. Capabilities](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-proxy-configuration-object#h-capabilities), section [7.1 Proxy](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-proxy-configuration-object#h-proxy):
> Let key be the result of lowercasing key. 

This seems to be a mistake as the table actually has the keys in camel case, so if you do this they will never match (in general I would much prefer we don't try to correct case here since we don't elsewhere).

juangj:
I think this was a mistake introduced in the process of addressing https://github.com/w3c/webdriver/issues/803; you are correct that it does not make sense to lowercase the key.

andreastt:
As I understand it https://github.com/w3c/webdriver/commit/a17c66a2a78bf649fafc29d2d7fb7e102c08508c should be reverted, and local ends fixed to send `{proxyType: "manual"}`.

PR #870, opened 2017-03-28
Make proxyType value comparison case-insensitive.

juangj:
Some existing clients (Java, notably) send the value in all-caps.

Addresses #835.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/870)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -2037,8 +2037,8 @@ with a "<code>moz:</code>" prefix:
    <li><p>Let <var>value</var> be the result of <a>getting a
     property</a> named <var>name</var> from <var>parameter</var>.
 
-   <li><p>Let <var>key</var> be the result of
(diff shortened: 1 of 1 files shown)


jgraham:
Wait, what? We're making exactly one field in the spec case insensitive? This seems like a terrible idea given that exisiting clients have to change anyway.

andreastt:
Why can’t language bindings be fixed?

More discussion in https://github.com/w3c/webdriver/issues/835.

juangj:
I'm ambivalent. It would be nice to avoid having just one field of the spec be case-insensitive, but it would also be nice to keep the marginal bit of backwards compatibility.

(But we can at least agree that the prior text where it said to lowercase the key didn't make sense.)

shs96c:
I can make the language bindings cope with this neatly (at least for java):

https://github.com/SeleniumHQ/selenium/commit/d0c298d5a6120f6fb1f9ed3a9cb682eb6410ed81

PR #878, opened 2017-04-01
Change normalized key value to Enter instead of Return

mjzffr:
While working on Marionette [1], I ran into an issue where the browser tells me "Return" is an invalid value for the key attribute of a KeyboardEvent. The UI Events spec [2] seems to agree: it should be "Enter", and that fixes the error encountered by Marionette.

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=1332279
[2] https://www.w3.org/TR/uievents-key/#keys-whitespace

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/878)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -7529,7 +7529,7 @@ is also removed.
  <tr><td><code>\uE003</code></td><td><code>"Backspace"</code></td></tr>
  <tr><td><code>\uE004</code></td><td><code>"Tab"</code></td></tr>
  <tr><td><code>\uE005</code></td><td><code>"Clear"</code></td></tr>
- <tr><td><code>\uE006</code></td><td><code>"Return"</code></td></tr>
+ <tr><td><code>\uE006</code></td><td><code>"Enter"</code></td></tr>
  <tr><td><code>\uE007</code></td><td><code>"Enter"</code></td></tr>
  <tr><td><code>\uE008</code></td><td><code>"Shift"</code></td></tr>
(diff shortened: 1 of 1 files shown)


shs96c:
Hmmm... "enter" and "return" have different keycodes, and potentially do different things. It's weird that the UI Events spec combines the two. Perhaps it's a bug in that spec?

mjzffr:
Hm, the two are lumped together in https://www.w3.org/TR/uievents-code/ as well.

I don't know the full context but, right now, both are handled as VK_RETURN in Firefox (which gets `key` "Enter"), and VK_ENTER was removed a few years ago because "it was never used": https://www.fxsitecompat.com/en-CA/docs/2014/keyboardevent-dom-vk-enter-has-been-removed/

shs96c:
Intriguing. We had to add the separate constants to the OSS Selenium project because people _were_ testing that they did different things. Would be interesting to find out how chrome, edge and safari handle this.

shs96c:
After chatting with some of the other people in the WG, and observing the plans for other products that use this specification, it seems best to keep the two separate constants in the spec, even if browsers treat them identically.

The major evidence is the use of the webdriver protocol for use in automating desktop applications, where these two keys (though often treated interchangeably) are different. Perhaps we should add a note that implementations "may" choose to treat VK_ENTER and VK_RETURN identically.

Please feel free to reopen this PR if you disagree with the reasoning or rationale, and many thanks for taking the time to highlight the issue.

Issue #881, opened 2017-04-05
What does "constant with the string" mean?


foolip:
In chapter [11. Elements
    ](https://www.w3.org/TR/webdriver/#dfn-web-element-identifier#h-elements):
> The web element identifier is a constant with the string "element-6066-11e4-a52e-4f735466cecf".

This should probably say "is the string constant" like in these cases:
https://www.w3.org/TR/webdriver/#dfn-web-window-identifier
https://www.w3.org/TR/webdriver/#dfn-web-frame-identifier

PR #882, opened 2017-04-05
Fix string constant language

andreastt:
Thanks to @foolip.

Fixes: https://github.com/w3c/webdriver/issues/881

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/882)


foolip:
Thanks!

PR #883, opened 2017-04-05
Simplify definition of "readiness state"

jugglinmike:
The definition of "maximum active sessions" states:

> [...] must be exactly one for a remote end that is an endpoint node.

This restriction is further supported by the definition of "active
sessions," which states:

> A remote end that is not an intermediary node has at most one active
> session at a given time.

Therefore, the "readiness state" for endpoint nodes can be described in
terms of a precise number ("one") of active sessions. The additional
state concerning the current user prompt is subsumed by this condition.

Simplify the definition of "readiness state" to reflect the limitations
of endpoint nodes.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/883)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -754,20 +754,18 @@ in the spec, as demonstrated in a (yet to be developed)
   and so are bound by the requirements on a <a>remote end</a> in terms
   of the wire protocol.
 
-<p>A <a>remote end</a>’s <dfn>readiness state</dfn> is
- for each <a>node type</a> said to be:
+<p>A <a>remote end</a>’s <dfn>readiness state</dfn> is the value corresponding
+ to the first matching statement:
 
 <dl class=switch>
- <dt><a>Intermediary node</a>
- <dd><p>False if it is known to be in a state
-  in which attempting to create a <a>new session</a> would fail.
-  Otherwise true.
+ <dt>the maximum <a>active sessions</a> is equal to the length of the list of
+  <a>active sessions</a>
+ <dt>the node is an <a>intermediary node</a> and is known to be in a state in
+   which attempting to create a <a>new session</a> would fail
+ <dd><p>False
 
- <dt><a>Endpoint node</a>
- <dd><p>False if the <a>maximum active sessions</a>
-  is equal to the length of the list of <a>active sessions</a>
-  or there is a <a>current user prompt</a> present.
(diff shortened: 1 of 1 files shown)


shs96c:
Your PR highlighted an issue with the way that `readiness state` is calculated for intermediary nodes --- they never took into account the number of `active sessions` (raised as #891) This means that we should probably clarify how we define this condition, to possibly be a logical AND on

# False if the maximum active sessions is equal to the length of the list of active sessions. Otherwise true.
# False if the node is an intermediary node and is known to be in a state in which attempting to create a new session would fail. Otherwise true.
# False is the node is an endpoint node and there is a current user prompt present. Otherwise true.

We can either modify this PR or create a new one.

jugglinmike:
I'm happy to follow up on this, but I believe that collapsing the conditions
across node types makes the endpoint-node-specific condition superfluous.
Basically, if "the node is an endpoint node and there is a current user prompt
present," then this implies that "the maximum active sessions is equal to the
length of the list of active sessions." This is because "the maximum active
sessions [...] must be exactly one for a remote end that is an endpoint node."

In other words, I believe this could be simplified to a logical AND on:

- False if the maximum active sessions is equal to the length of the list of
  active sessions. Otherwise true.
- False if the node is an intermediary node and is known to be in a state in
  which attempting to create a new session would fail. Otherwise true.

shs96c:
You are correct. We can collapse this down to those two conditions. If we can phrase this using positive language, I think it'll be slightly clearer, but I could be wrong --- I'll trust your judgement.

jugglinmike:
I've had to amend the commit message and force-push, so I've made the original
state of this branch available here, in case that's useful:

> If we can phrase this using positive language, I think it'll be slightly
> clearer, but I could be wrong --- I'll trust your judgement.

If you prefer, this would look like:

> A remote end's readiness state is True if all of the following conditions are
> met:
>
> - the maximum active sessions is equal to the length of the list of active
>   sessions
> - the node is an endpoint node or is known to be in a state in which
>   attempting to create a new session would succeed

Of course, the other variable in this is whether to use the "switch"
construction or a list of conditions. Happy to re-phrase as you prefer!

shs96c:
Thanks for fixing this and bearing with the reviews!

jugglinmike:
Happy to help out :)

PR #889, opened 2017-04-06
Make Send Alert Text take a string text field

andreastt:
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/889)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -5695,8 +5695,8 @@ must run the following steps:
   of <a>getting a property</a> called "<code>text</code>"
   from the <var>parameters</var> argument.
 
- <li><p>If <var>text</var> is not a string, return an <a>error</a>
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -8630,15 +8630,13 @@
     <a>error code</a> <a>unsupported operation</a>.
   </dl>
 
- <li><p>Let <var>character array</var> be the result of
-  <a>getting the property</a> <code>value</code>
-  from the <var>parameters</var> argument.
-
  <li><p>Let <var>text</var> be the result of
-  joining together each item in <var>character array</var> to a string.
+  <a>getting the property</a> "<code>text</code>"
```
shs96c: We need to add a check that this is a `string`

andreastt: Oh, good call.

andreastt: Fixed!

PR #890, opened 2017-04-06
fix typo in capability matching section

jlipps:
I didn't see a variable `matched capability` anywhere and it seemed more reasonable to suppose we're setting a name and value on the capabilit_ies_ object, so I figured this was a typo.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/890)


shs96c:
Good catch. Thank you!

PR #894, opened 2017-04-17
Take <select multiple> into account in obscured element test

andreastt:
As the options of a `<select multiple>` are painted and represented as DOM
elements, the center point would end up targetting one of the `<option>`
elements.  We can take this into account by using `select.contains(option)`.

If the first pointer-interactable paint tree element is itself, this is
also passes the check, since the `DOMElement.contains(otherEl)` performs
an _inclusive_ descendant check.  This means `element.contains(element)`
is true.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/894)


shs96c:
I _ think_ this makes sense. Will correct the typos once the merge is in.

PR #898, opened 2017-04-19
Clarify how a UUID is meant to be serialized

### LINKED ISSUE #863: It's unclear how the session id is serialized
In chapter [8. Sessions](https://w3c.github.io/webdriver/webdriver-spec.html#new-session#h-sessions), section [8.1 New Session](https://w3c.github.io/webdriver/webdriver-spec.html#new-session#h-new-session):
> Let session id be a UUID. 

At some point (after setting it in the session, but before sending the response), this should be converted to a string representation as defined in https://www.ietf.org/rfc/rfc4122.txt

shs96c:
Closes #863.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/898)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -661,7 +661,13 @@ in the spec, as demonstrated in a (yet to be developed)
  a 128 bits long URN that requires no central registration process. [[!RFC4122]].
  <dfn>Generating a UUID</dfn> means Creating a
  <a>UUID</a> From Truly Random or Pseudo-Random Numbers [[!RFC4122]],
- and serializing it as appropriate for the callee.
+ and serializing it by converting the bit sequence to a string
+ representation, following the process outlined in [[!RFC4122]]. This
+ means that each field within the UUID is treated as an integer and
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -661,7 +661,13 @@
  a 128 bits long URN that requires no central registration process. [[!RFC4122]].
  <dfn>Generating a UUID</dfn> means Creating a
  <a>UUID</a> From Truly Random or Pseudo-Random Numbers [[!RFC4122]],
- and serializing it as appropriate for the callee.
+ and serializing it by converting the bit sequence to a string
+ representation, following the process outlined in [[!RFC4122]]. This
```
jgraham: "`and converting it to the string representation [[!RFC4122]].`" rather than all of this text.

jgraham:
Why is "Squash and merge" not enabled for this repo? I would just have merged this if it was.

shs96c:
I think we really need "squash, rebase, and merge"

PR #903, opened 2017-04-26
Document send keys

veith:
If you want to test keyboard navigation, you have to send keys "directly to the browser" without selecting an element before. 

In the jsonWireProtokoll was a command for doing that. (https://github.com/SeleniumHQ/selenium/wiki/JsonWireProtocol#sessionsessionidkeys) .

I suggest to place it in section 15 (Document Handling), because activeElement is on the document itself (https://developer.mozilla.org/en/docs/Web/API/Document/activeElement), but maybe this is totally wrong.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/903)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -6259,6 +6259,336 @@ must run the following steps:
 </ol>
 </section> <!-- /Execute Async Script -->
 </section> <!-- /Executing Script -->
+
+
+<section>
+ <h3>Send Keys</h3>
+
+ <table class="simple jsoncommand">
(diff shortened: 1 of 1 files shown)


andreastt:
You can use the actions API to send arbitrary keys without selecting the element first. But to be honest, you could also use Element Send Keys against the document element if you wanted to do that, with the added safety checks it provides.

veith:
@andreastt Thank you for the hint with the actions API.

PR #926, opened 2017-05-31
delete session: close session only if it is still active

andreastt:
For implementation compatibility we should not try to close the session
if the session does not exist.  This can be interpreted in different
ways because of the use of the "try" algorithm.

Instead we make the Delete Session algorithm explicitly call out that
the session must be active in order to attempt closing it.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/926)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -2759,7 +2759,8 @@ with a "<code>moz:</code>" prefix:
 <p>The <a>remote end steps</a> are:
 
 <ol>
- <li><p><a>Try</a> to <a>close the session</a>.
+ <li><p>If the <a>current session</a> is an <a>active session</a>,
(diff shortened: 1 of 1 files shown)


shs96c:
This makes closing a non-existent session a no-op. This seems to be the behaviour of the existing Selenium standalone server, and the IEDriverServer.

andreastt:
@shs96c Yes, that was the intention. This originated from https://github.com/mozilla/geckodriver/issues/732#issuecomment-303750567 because geckodriver currently returns a 500 Internal Server Error when trying to close a session that has already been closed.

E.g. if you close the last window and it implicitly ends the session, then calling `driver.quit()`/Delete Session shouldn’t return an error.

PR #929, opened 2017-06-01
Set the top-level browser context to use for a session.

### LINKED ISSUE #924: New Session command doesn't set the current browsing context
In chapter [8. Sessions](https://w3c.github.io/webdriver/webdriver-spec.html#new-session#h-sessions), section [8.1 New Session](https://w3c.github.io/webdriver/webdriver-spec.html#new-session#h-new-session):
> 8.1 New Session

Typically implementations start a new browsing context when the session is created and set that as the new [top-level] browsing  context. I don't think that creating a new one should be required, but it should either create a new one, or make an existing UA-specific browsing context the current browsing context (e.g. one that has focus).

shs96c:
This, by necessity, needs to be done in an
implementation-spefific way. I've avoided using
language that ties the spec to only automating
browsers, since there are already implementations
for mobile devices, and even automating Windows
and OS X.

Closes #924

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/929)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -2732,7 +2732,9 @@ with a "<code>moz:</code>" prefix:
  <li><p>Set the <a>webdriver-active flag</a> to true.</p>
 
  <li><p>Set the <a>current top-level browsing context</a>
-  for <var>session</var> in an implementation-specific way.
+  for <var>session</var> in an implementation-specific way. This
+  should be the <a>top-level browsing context</a> of the UA's
+  <a>current browsing context</a>.
 
   <p class="note">WebDriver implementations typically start a
    completely new browser instance, but there is no requirement in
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -2731,6 +2731,16 @@
 
  <li><p>Set the <a>webdriver-active flag</a> to true.</p>
 
+ <li><p>Set the <a>current top-level browsing context</a>
+  for <var>session</var> in an implementation-specific way.
```
andreastt: _Current top-level browsing context_ is a WebDriver concept, and not a browser concept. This state is what WebDriver implementations use to track which browsing context to interact with.

What we need to say is that the _current browsing context_ is the _top-level browsing context of the UA’s currently selected browsing context_, or something along those lines.

shs96c: I've mirrored the language from [Section 8](https://w3c.github.io/webdriver/webdriver-spec.html#sessions) that discusses the same thing.

PR #937, opened 2017-06-05
find: remove repeated algorithm step

jugglinmike:
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/937)


shs96c:
Good catch. Thank you!

PR #939, opened 2017-06-05
find: Correct implementation of "xpath" strategy

jugglinmike:
Specify all parameters as required by the `evaluate` function, and
normalize the return type according to the structure expected by the
"find" algorithm (i.e. a NodeList).

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/939)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -591,6 +591,8 @@ in the spec, as demonstrated in a (yet to be developed)
   <dd><p>The following terms are defined in the Document Object Model XPath standard [[!XPATH]]
     <ul>
       <!-- evaluate --><li><a href="https://www.w3.org/TR/DOM-Level-3-XPath/xpath.html#XPathEvaluator-evaluate"><dfn><code>evaluate</code></dfn></a>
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -4581,14 +4583,35 @@
...
-  selector</a>.
+ <li><p>Repeat, while <var>index</var> is less than <var>length</var>:
+
+  <ol>
+   <li>Let <var>node</var> be the result of calling <a>snapshotItem</a> with
+    argument <var>index</var>, with the <a>context object</a> equal to
+    <var>evaluateResult</var>.
+
+   <li><p>If <var>node</var> is not an <a>element</a> return an <a>error</a>
+    with <a>error code</a> <a>invalid selector</a>.
```
jugglinmike: The error handling in [the call site](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-find) is a little ambiguous:

> 6. If a DOMException, SyntaxError, XPathException, or other error occurs
>    during the execution of the element location strategy, return error
>    invalid selector. 

The intention is clear, but to the pedantically-minded (like me), it's not quite accurate to say that an error "occurs" in this case. I think it would be better if we were to explicitly "throw" in this patch. There is no convention for throwing within the WebDriver specification, so I could use an editor's input on that.

Alternatively, we could update "Find" with something like, "if an error is thrown or returned by the previous step."

shs96c: I think we'll be fine with updating the `Find` algorithm's step 5 to read:

```
Let elements returned be the result of *trying* the relevant element location strategy call with arguments start node, and selector.
```

That is, to make use of our `trying` mechanism for fallible algorithms, where the handling is already clear.

PR #948, opened 2017-06-09
Changes around navigator.webdriver

### LINKED ISSUE #922: Why Unforgeable
In chapter [4. Interface](https://w3c.github.io/webdriver/webdriver-spec.html#interface#h-interface):
> [Unforgeable]

I don't see how this makes sense, since navigator itself can be manipulated easily.

### LINKED ISSUE #923: Extend Navigator through a mixin
In chapter [4. Interface](https://w3c.github.io/webdriver/webdriver-spec.html#interface#h-interface):
> partial interface Navigator

The way you extend Navigator doesn't work for workers, which use WorkerNavigator. You need to define a [NoInterfaceObject] interface that is implemented by both Navigator and WorkerNavigator, similar to how the HTML Standard does it.

### LINKED ISSUE #947: Do not put navigator.webdriver on prototype chain
It was proposed in the [Intent To Ship](https://groups.google.com/a/chromium.org/d/msg/blink-dev/6GysDZCWwB8/U1RrBcsnBwAJ) of this feature in Chrome to expose `navigator.webdriver` only if UA is controlled by automation.

shekyan:
- expose `navigator.webdriver` if UA is controlled by WebDriver (fixes #947)
- drop [Unforgeable] (fixes #922)
- extend Navigator through mixin (fixes #923)

/cc @annevk, @foolip

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/948)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -344,6 +344,7 @@ in the spec, as demonstrated in a (yet to be developed)
    <!-- User prompt --> <li><dfn data-lt="user prompts"><a href=https://html.spec.whatwg.org/#user-prompts>User prompt</a></dfn>
    <!-- Value attribute --><li><dfn><a href=https://html.spec.whatwg.org/#dom-input-value><code>value</code> attribute</a></dfn>
    <!-- Value --> <li><dfn><a href=https://html.spec.whatwg.org/#concept-fe-value>Value</a></dfn>
+   <!-- WorkerNavigator --> <li><dfn data-lt="workernavigator"><a href=https://html.spec.whatwg.org/#workernavigator>WorkerNavigator object</a></dfn>
   </ul>
 
  <dd><p>The HTML specification also defines a number of elements
@@ -697,19 +698,25 @@ in the spec, as demonstrated in a (yet to be developed)
 </section> <!-- /Terminology -->
 
 <section>
-<h2>Interface</h2>
+
+<h2 data-dfn-for="NavigatorAutomationInformation">Interface</h2>
 
 <p>If the <dfn>webdriver-active flag</dfn> is true,
- the user agent must support the following partial interface:
+ the user agent must support <dfn>NavigatorAutomationInformation</dfn> interface:
+
+ <pre class=idl>
+ Navigator implements NavigatorAutomationInformation;
+</pre>
 
 <pre class=idl>
-	[NoInterfaceObject, Exposed=(Window,Worker)]
-	partial interface Navigator {
-	[Exposed=Window] readonly attribute boolean webdriver; // constant true
-};</pre>
+ [NoInterfaceObject, Exposed=(Window)]
+ interface NavigatorAutomationInformation {
+ [Exposed=Window] readonly attribute boolean webdriver; // constant true
+ };
+</pre>
 
-<p data-dfn-for="Navigator">If the <dfn><code>webdriver</code></dfn> IDL attribute
- of the <code><a>Navigator</a></code> interface is present,
+<p data-dfn-for="NavigatorAutomationInformation">Note that the <dfn>webdriver</dfn>
+ property should not be exposed on <code><a>WorkerNavigator</a></code>, and
  its value must be true.
 
 <p>This property defines a standard way for a co-operating user agent to inform a website
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -711,13 +711,13 @@ in the spec, as demonstrated in a (yet to be developed)
 <pre class=idl>
  [NoInterfaceObject, Exposed=(Window)]
  interface NavigatorAutomationInformation {
- [Exposed=Window] readonly attribute boolean webdriver; // constant true
+ readonly attribute boolean webdriver; // always returns true
  };
 </pre>
 
 <p data-dfn-for="NavigatorAutomationInformation">Note that the <dfn>webdriver</dfn>
- property should not be exposed on <code><a>WorkerNavigator</a></code>, and
- its value must be true.
+ property should not be exposed on <code><a>WorkerNavigator</a></code>, and it must be
+ a getter that always returns true.
 
 <p>This property defines a standard way for a co-operating user agent to inform a website
   that it is controlled by WebDriver.
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -699,13 +699,18 @@ in the spec, as demonstrated in a (yet to be developed)
 <section>
 <h2>Interface</h2>
 
-<pre class=idl>partial interface Navigator {
-	[Unforgeable] readonly attribute boolean webdriver;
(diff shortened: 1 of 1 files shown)


shekyan:
I don't like how IDL is rendered, is there a way to avoid putting comments on new line?

andreastt:
Paging @annevk for a sanity check on the WebIDL.

webdriver-spec.html
```diff
@@ -699,13 +699,18 @@
...
 <h2>Interface</h2>
 
-<pre class=idl>partial interface Navigator {
-	[Unforgeable] readonly attribute boolean webdriver;
+<p>If the <dfn>webdriver-active flag</dfn> is true,
+ the user agent must support the following interface:
+
+<pre class=idl>
+	[NoInterfaceObject, Exposed=(Window,Worker)]
+	interface Navigator {
```
foolip: This should still be a partial interface.

webdriver-spec.html
```diff
@@ -699,13 +699,18 @@
 <section>
 <h2>Interface</h2>
 
-<pre class=idl>partial interface Navigator {
-	[Unforgeable] readonly attribute boolean webdriver;
+<p>If the <dfn>webdriver-active flag</dfn> is true,
+ the user agent must support the following interface:
```
foolip: +partial

webdriver-spec.html
```diff
@@ -699,13 +699,18 @@
 <section>
 <h2>Interface</h2>
 
-<pre class=idl>partial interface Navigator {
-	[Unforgeable] readonly attribute boolean webdriver;
+<p>If the <dfn>webdriver-active flag</dfn> is true,
```
foolip: Is it possible for this flag to change at runtime, i.e. is it possible to use WebDriver to connect to an already-running browser?

shekyan: As I understand from https://w3c.github.io/webdriver/webdriver-spec.html#new-session there is a hypothetical possibility to attach to a browser instance at runtime. 
> WebDriver implementations typically start a completely new browser instance, but there is no requirement in this specification (or for WebDriver only to be used to automate only web browsers). Implementations might choose to use an existing browser instance, eg. by selecting the window that currently has focus.

Implementations that I looked at (WebKit, Chrome) require new instance.

foolip: OK, well in that case I think this is fine, could tweak this in the future if needed.

annevk:
This does not work. You need to define something similar to e.g. "NavigatorID" in HTML as you need to extend both `Navigator` and `WorkerNavigator`.

shekyan:
@annevk , is this closer to the right way of doing it?

webdriver-spec.html
```diff
@@ -699,13 +700,37 @@
...
-};</pre>
+<p>If the <dfn>webdriver-active flag</dfn> is true,
+ the user agent must support the following partial interfaces:
 
-<p data-dfn-for="Navigator">The <dfn><code>webdriver</code></dfn> IDL attribute
- of the <code><a>Navigator</a></code> interface must return
- the value of the <dfn>webdriver-active flag</dfn>, which is initially false.
+ <pre class=idl>
+ partial interface Navigator {
+ };
```
annevk: No need for this partial declaration.

webdriver-spec.html
```diff
@@ -699,13 +700,37 @@
...
+ Navigator implements AutomationInformation;
+</pre>
+
+<pre class=idl>
+ [Exposed=Worker]
+ interface WorkerNavigator {};
```
annevk: No need for this interface declaration either.

webdriver-spec.html
```diff
@@ -699,13 +700,37 @@
...
+ WorkerNavigator implements AutomationInformation;
+</pre>
+
+<section data-dfn-for="AutomationInformation" data-dfn-for="AutomationInformation">
+
+<h2><dfn>AutomationInformation</dfn> interface</h2>
```
annevk: It should have Navigator as a prefix. That's the pattern.

webdriver-spec.html
```diff
@@ -699,13 +700,37 @@
...
+
+<pre class=idl>
+ [NoInterfaceObject, Exposed=(Window,Worker)]
+ interface AutomationInformation {
+ [Exposed=Window] readonly attribute boolean webdriver; // constant true
```
annevk: If you only expose it on Window, why extend both WorkerNavigator and Navigator?

shekyan: Uh, that's how I interpreted your comments in https://github.com/w3c/webdriver/issues/923 .

shekyan: @annevk, addressed comments.

webdriver-spec.html
```diff
@@ -697,15 +698,26 @@
...
+ the user agent must support <dfn>NavigatorAutomationInformation</dfn> interface:
+
+ <pre class=idl>
+ Navigator implements NavigatorAutomationInformation;
+</pre>
+
+<pre class=idl>
+ [NoInterfaceObject, Exposed=(Window)]
+ interface NavigatorAutomationInformation {
+ [Exposed=Window] readonly attribute boolean webdriver; // constant true
```
annevk: You don't need the additional Exposed=Window line here. I'd make the comment "always returns true" as per precedent in DOM.

webdriver-spec.html
```diff
@@ -697,15 +698,26 @@
...
+ };
+</pre>
+
+<p data-dfn-for="NavigatorAutomationInformation">Note that the <dfn>webdriver</dfn>
+ property should not be exposed on <code><a>WorkerNavigator</a></code>, and
+ its value must be true.
```
annevk: It's a getter that returns true. It doesn't have a value like that.

annevk:
This looks good, though why does it always return true again? Wouldn't it normally return false?

shekyan:
Oh, because we don't want to expose this property unless UA is controlled by webdriver. That's what Chrome folks suggested to do in my [Intent to ship](https://groups.google.com/a/chromium.org/d/msg/blink-dev/6GysDZCWwB8/Az0byBmIBwAJ)

annevk:
I see, that kind of thing isn't really supported by IDL, but I guess if you address it in prose that should be fine. However, the IDL probably shouldn't suggest it always returns true then.

shekyan:
I followed the pattern in https://html.spec.whatwg.org/multipage/system-state.html#client-identification
`If the navigator compatibility mode is Gecko, then the user agent must also support the following partial interface:`

```
partial interface NavigatorID {
  [Exposed=Window] boolean taintEnabled(); // constant false
  [Exposed=Window] readonly attribute DOMString oscpu;
};
```

annevk:
Fair enough, all good then.

shs96c:
This looks fine to me. Is this ready to be merged in?

shekyan:
I don't have nothing more to add.

PR #952, opened 2017-06-20
Scope extension commands to current session

jugglinmike:
Extension commands are expected to be executed in the context of the
remote end's current session. This expectation is normatively enforced
by the remote end's processing model and further suggested by the first
example in the section titled "Protocol Extensions."

Update the extension mechanism to be phrased in terms of a URI template
that includes the current session ID (instead of a URL that does not).

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/952)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -1690,13 +1690,13 @@ in the spec, as demonstrated in a (yet to be developed)
  that is an <a>ASCII lower alpha</a> string,
  and which should bear some resemblance to what the command performs.
  The name is used to form an <a>extension command</a>’s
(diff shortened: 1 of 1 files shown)


shs96c:
This isn't correct: not all extensions are going to be at the session level. As an example, the selenium server has an extension command that allows new drivers to be registered at the server level ("Add Config"), and Grid has a number of extensions that allow users to figure out the health of the grid as a whole.

jugglinmike:
Understood. This has bearing on gh-954; I'll update that patch to account for the use cases you've described.

PR #954, opened 2017-06-20
Account for "Status" command in processing model

jugglinmike:
Similar to the "New Session" command, the "Status" command does not
specify a session identifier, nor does it depend on the "current
session" variable. The substeps related to those values should therefore
be skipped for the "Status" command.

Furthermore, processing either command should not involve extracting the
"session id" value from the variable named "url variables". Re-locate
that step to the branch that concerns only those commands where it is
relevant.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/954)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -889,12 +889,12 @@ in the spec, as demonstrated in a (yet to be developed)
   <p>Otherwise, let <var>command</var> and <var>url variables</var>
    be <var>request match</var>’s data.
 
- <li><p>Let <var>session id</var> be the corresponding variable
-  from <var>url variables</var>.
-
- <li><p>If <var>command</var> is not <a>New Session</a>:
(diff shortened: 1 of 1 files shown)


jugglinmike:
In gh-952, @shs96c wrote:

> [...] not all extensions are going to be at the session level. As an example,
> the selenium server has an extension command that allows new drivers to be
> registered at the server level ("Add Config"), and Grid has a number of
> extensions that allow users to figure out the health of the grid as a whole.

This means the current patch does not go far enough. Instead of extending the
condition to exclude the "Status" command explicitly, it must account for any
command that does not specify a "session id".

Although I originally intended to make the necessary adjustments within this
pull request, the fix also involves modifying the commit message, and that
change is easier to share/apply via a new pull request--see gh-956.

PR #956, opened 2017-06-21
Account for any session-independent commands

jugglinmike:
The "New Session" command does not specify a session identifier, nor
does it depend on the "current session" variable. This is also true of
the "Status" command, and any number of extension commands may likewise
operate outside the context of any particular session.

Skip the substeps related to the current session for any command that
does not specify a "session id" variable. Introduce a non-normative note
documenting the intention of this condition since it is partially
intended to satisfy externally-defined criteria.

Furthermore, processing such commands should not involve extracting the
"session id" value from the variable named "url variables". Re-locate
that step to the branch that concerns only those commands where it is
relevant.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/956)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -889,12 +889,15 @@ in the spec, as demonstrated in a (yet to be developed)
   <p>Otherwise, let <var>command</var> and <var>url variables</var>
    be <var>request match</var>’s data.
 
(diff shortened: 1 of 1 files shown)


shs96c:
LGTM. I think this properly captures the intent.

Issue #961, opened 2017-06-21
Allow extensions to specify session-based commands


jugglinmike:
Currently, the requirements on extension commands are too constrained to allow
for specification of algorithms tied to a specific session. This is due to the
following definition:

> The extension command’s extension command URL is a URL composed of the
> extension command prefix, followed by "/", and the extension command’s name.
> The extension command URL, along with the HTTP method and extension command,
> is added to the table of endpoints and thus follows the same rules for
> request routing as that of other built-in commands. 

...where "extension command prefix" and "extension command name" are both ASCII
lower alpha strings.

(Note that the text quoted above is erroneous because it the value named
"extension command URL" is described as a URL, but the constructed value is
actually a URI. In addition, the value is inserted into a table that includes
URI template values but does not include either URLs or URIs. The intent of the
text is clear, though, and the types may be corrected independently from a
resolution of the larger issue.)

Initially, I interpreted this definition as erroneously omitting text to insert
the "session id" URI template parameter. I learned that interpretation was
wrong in gh-952, where @shs96c wrote:

> not all extensions are going to be at the session level. As an example, the
> selenium server has an extension command that allows new drivers to be
> registered at the server level ("Add Config"), and Grid has a number of
> extensions that allow users to figure out the health of the grid as a whole.

But while some extensions may not require an active session, others will, and
the current text makes it impossible for these to be defined. The section
currently includes an example of such a command:
`/session/5d376174-36f0-11e5-9b9a-6bdf200a3f7f/ms/edge/context`.

I have two ideas for resolving this:

- Allow extension commands to specify URI template variables, either by
  defining compete URI templates, or by extending the character sets of
  "extension command prefix" and "extension command name" with the necessary
  characters
- Introduce a boolean attribute "requires session" (or similar) for extension
  commands, and insert `session/{ session id }` into the URI template for those
  commands that specify this as "true".

The latter approach may be easier to use since it does not require
specification editors to discover the implicit session-detecting logic of the
remote end's processing model. It also tends to make extensions more explicit:
the intention of a boolean attribute is more obvious than a pattern in a URI
template.

The former approach is much more powerful because it allows extensions to
specify additional URL parameters, e.g. `/session/{ session id }/bluetooth/{
device id }`.  There is no current use case for that functionality as far as
I'm aware, but it seems likely to be useful for some specifications.

I'd be happy to follow up with a patch implementing either of these changes,
or some third approach that I haven't considered.

shs96c:
[...]

PR #962, opened 2017-06-22
Define extension command URI parts as templates

jugglinmike:
Interpret the URI parts specified by extension commands as templates,
thereby allowing for extension commands such as `/session/{session
id}/permissions/grant`. Include a non-normative note to call attention
to the implicit "session retrieval" behavior of the command processing
model.

---

This is an initial pass at resolving gh-961. There, I pointed out that the command processing model's "session retrieval" behavior may not be apparent to those reading this section. I believe this justifies a non-normative note to call attention to that behavior.

@shs96c RFC6570 defines [distinct "levels" for templates](https://tools.ietf.org/html/rfc6570#section-1.2). So far, it seems as though WebDriver and its extensions may only need the capabilities of so-called "Level 1 templates". Do you see any value in restricting extension command templates to Level 1?

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/962)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -430,7 +430,6 @@ in the spec, as demonstrated in a (yet to be developed)
  <dd>The following terms are defined
   in the Infra standard: [[!INFRA]]
    <ul>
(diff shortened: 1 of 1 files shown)


AutomatedTester:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/w3c/webdriver/962)*

Issue #950, opened 2017-06-09
Timeout-based behavior may be untestable


jugglinmike:
While working on WebDriver tests in the Web Platform Tests project, I've
started to identify aspects of certain algorithms that appear to be untestable.
Specifically, it seems as though any behavior based on waiting/timing out
cannot be reliably observed from either the process issuing driver commands or
from the browsing context itself.

For example, from [the find algorithm](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-find):

> 1. Let end time be the current time plus the session implicit wait timeout.
>
> [...]
>
> 4. Let elements returned be the result of the relevant element location
>    strategy call with arguments start node, and selector.
> 5. If a DOMException, SyntaxError, XPathException, or other error occurs
>    during the execution of the element location strategy, return error
>    invalid selector.
> 6. If elements returned is empty and the current time is less than end time
>    return to step 4. Otherwise, continue to the next step. 

The element retrieval operation in step 4 is not directly observable. Tests can
only verify the result of the overall command--they cannot determine how many
retrieval attempts were made to reach that result (i.e. whether polling has
taken place).

It is possible to issue a "find element" command for a non-existent element and
subsequently execute a script to define that element. If the "find element"
command then returns the correct element reference, it seems likely that the
implementation is correct. However, this is not guaranteed. That's because of
the unobservable nature of the retrieval operation--such a test would only be
effective if the element were inserted *after* at least one attempt to retrieve
were made.

A "best effort" approach might involve setting a long (e.g. 3000 ms) implicit
wait timeout and deferring element insertion until half of that time had
passed.  Practically speaking, it seems likely that polling would be in
progress after 1.5 seconds... But this is still fundamentally a race condition.
And although longer timeouts would make expression of the race less likely,
WebDriver tests are already among the slowest in WPT; relying on such long
delays will make them even more so.

I believe this general problem applies to most (if not all) algorithm steps
that involve "waiting" (e.g. "wait for navigation to complete" in [the "Go"
command](https://w3c.github.io/webdriver/webdriver-spec.html#go)).

If this assessment is accurate, I can think of four possible responses:

1. Accept the functionality as untestable and do not make any attempt to verify
   it within WPT
2. Parameterize all timeout intervals such that test consumers can select a
   value that is appropriate for their environment
3. Approximate appropriate delays by issuing parallel "simulation" commands.
   For "find element," this would mean a second "find element" command for an
   element that is known to be present. It seems likely that by the time this
[...]

PR #965, opened 2017-06-27
Session request queue

jugglinmike:
This patch standardizes the expected behavior as described in the discussion of
gh-950:

> There's the text in the preamble for Section 6 that says that implementations
> must be HTTP compliant. An intermediary node is welcome to handle more than
> one command at a time. The only caveat is that each session should only have
> one command issued against it at a time, though we never spell this out in
> the spec.
>
> [...]
>
> Actually, this is also incorrect. It should always be possible to execute the
> "Delete Session" command even if another command is running for that session. 
>
> [...]
>
> There are three possible conditions that need to be tracked for an endpoint
> node:
>
> 1/ Sessionless commands can be processed regardless of whether or not there's
>    a currently active session.
> 2/ Delete Session can (optionally) be executed immediately
> 3/ All other commands should be pleased in a queue per session and executed
>    serially.
>
> I'm not aware of any implementations that don't merge those last two
> branches, as that hugely simplifies the implementation.

That last allowance for the "Delete Session" command has given me pause. It
doesn't seem sufficient to allow for the condition in the processing model
alone--I believe all commands will have to be updated to support interruption.
If this is true, and if all existing implementations queue Delete commands,
then it might be best to omit this special case behavior. (I've implemented
that part in a separate commit to simplify its removal.)

Note that the phrasing proposed here uses [a specification queue](https://infra.spec.whatwg.org/#queues) in a somewhat non-traditional way: algorithm execution blocks on the identity of "the first member of the queue." While the Infra spec does not describe this kind of operation, I'm taking a cue (please note the pun) from [the `ServiceWorkerContainer#ready` attribute of the Service Worker specification](https://w3c.github.io/ServiceWorker/#navigator-service-worker-ready), whose associated algorithm includes the following steps:

> 2. If registration is null, then:
>    1. Wait until scope to registration map has a new entry.
>    2. Jump to the step labeled CheckRegistration.

While the primitive is different in this case (i.e. a map and not a queue),
this text sets a precedent for the general operation of reacting to changes in
data structure membership.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/965)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -904,26 +904,29 @@ in the spec, as demonstrated in a (yet to be developed)
      list of <a>active sessions</a>, or <a><code>null</code></a> if
      there is no such matching <a>session</a>.
 
-   <li><p>If the <a>current session</a> is not <a><code>null</code></a>:
+   <li><p>If <var>command</var> is not <a>Delete Session</a>:
 
     <ol>
-      <li><p>Enqueue <var>request</var> in the <a>current session</a>'s <a>request queue</a>.
+     <li><p>If the <a>current session</a> is not <a><code>null</code></a>:
 
-      <li><p>Wait until the first element in the <a>current session</a>'s <a>request queue</a>
-       is <var>request</var>:
+      <ol>
+        <li><p>Enqueue <var>request</var> in the <a>current session</a>'s <a>request queue</a>.
 
-      <li><p>Dequeue <var>request</var> from the <a>current session</a>'s <a>request queue</a>.
+        <li><p>Wait until the first element in the <a>current session</a>'s <a>request queue</a>
+         is <var>request</var>.
 
-      <li><p>If the list of <a>active sessions</a> no longer contains the
-       <a>current session</a>, set the <a>current session</a> to
-       <a><code>null</code></a>.
-    </ol>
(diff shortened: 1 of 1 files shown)


AutomatedTester:
Ideally we dont want Delete to bypass the queue. The API is supposed to be "blocking". So each command should be completed and returned so delete shouldn't need to bypass anything.

jugglinmike:
I'm happy to hear that, as it seems to simplify things for everyone. If we removed the second commit on this branch, would that meet your expectations?

webdriver-spec.html
```diff
@@ -903,6 +904,21 @@
...
+      <li><p>Enqueue <var>request</var> in the <a>current session</a>'s <a>request queue</a>.
+
+      <li><p>Wait until the first element in the <a>current session</a>'s <a>request queue</a>
+       is <var>request</var>:
+
+      <li><p>Dequeue <var>request</var> from the <a>current session</a>'s <a>request queue</a>.
+
+      <li><p>If the list of <a>active sessions</a> no longer contains the
+       <a>current session</a>, set the <a>current session</a> to
+       <a><code>null</code></a>.
```
shs96c: Clever.

shs96c:
Closed with 2370a2179a9bebbd023ab2803654f2f660812bda

PR #968, opened 2017-06-30
Make pointer events include the aggregated key state

### LINKED ISSUE #595: The pointer section should state which keys are considered depressed as part of the action
From a comment by @jgraham in #500. 

Assigning to him to either flesh-out or close.

shs96c:
MouseEvents contain references to the modifier keys,
and these should be calculated as required.

https://www.w3.org/TR/DOM-Level-3-Events/#interface-mouseevent

Closes #595.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/968)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -7045,6 +7045,67 @@ is also removed.
  actions to the list and rely on the fact that the reset operations
  are idempotent.
 
+<p>The <dfn>calculated global key state</dfn> is the aggregated key
+ state from all <a>key input state</a> objects. It can be calculated
+ this way:
+
+<ol>
+ <li><p>Let <var>pressed</var> be a new Set.
+
+ <li><p>Let <var>alt</var>, <var>ctrl</var>, <var>meta</var>,
+  and <var>shift</var> be the <a>Boolean</a> <code>false</code> value.
+
+ <li><p>For enumerable <a>own property</a> in the <a>input state table</a>:
+
+ <ol>
+  <li><p>Let <var>source</var> be the value of the property.
+
+  <li><p>If <var>source</var> is not a <a>key input state</a>,
+   continue to the first step of this loop.
+
+  <li><p>Let <var>key state pressed</var> be the result of <a>getting
+   a property</a> named <code>pressed</code> from <var>source</var>.
+
+  <li><p>Add all strings from <var>key state pressed</var>
+   to <var>pressed</var>.
+
+  <li>Let <var>alt</var> be a logical OR of <var>alt</var> and the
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -7045,6 +7045,67 @@ <h2 data-dfn-for="NavigatorAutomationInformation">Interface</h2>
...
+   from <var>source</var>.
+
+  <li>Let <var>shift</var> be a logical OR of <var>shift</var> and the
+   result of <a>getting a property</a> named <code>shift</code>
+   from <var>source</var>.
+ </ol>
+
+ <li><p>Let <var>state</var> be a new JSON <a>Object</a>.
+
+ <li><o><a>Set a property</a> on <var>state</var> with
```
AutomatedTester: I am guessing `<o>` is supposed to be `<p>`

webdriver-spec.html
```diff
@@ -7045,6 +7045,67 @@ <h2 data-dfn-for="NavigatorAutomationInformation">Interface</h2>
...
+  name <var>pressed</var> and value <var>pressed</var>.
+
+ <li><o><a>Set a property</a> on <var>state</var> with
```
AutomatedTester: I am guessing `<o>` is supposed to be `<p>`

webdriver-spec.html
```diff
@@ -7045,6 +7045,67 @@ <h2 data-dfn-for="NavigatorAutomationInformation">Interface</h2>
...
+  name <var>alt</var> and value <var>alt</var>.
+
+ <li><o><a>Set a property</a> on <var>state</var> with
```
AutomatedTester: I am guessing `<o>` is supposed to be `<p>`

webdriver-spec.html
```diff
@@ -7045,6 +7045,67 @@ <h2 data-dfn-for="NavigatorAutomationInformation">Interface</h2>
...
+  name <var>ctrl</var> and value <var>ctrl</var>.
+
+ <li><o><a>Set a property</a> on <var>state</var> with
```
AutomatedTester: and here

webdriver-spec.html
```diff
@@ -7045,6 +7045,67 @@ <h2 data-dfn-for="NavigatorAutomationInformation">Interface</h2>
...
+  name <var>meta</var> and value <var>meta</var>.
+
+ <li><o><a>Set a property</a> on <var>state</var> with
```
AutomatedTester: and here...

shs96c:
I was on fire that day.

PR #969, opened 2017-06-30
Don't set expiry on a session cookie

### LINKED ISSUE #957: Define the expiry for a session cookie.
The specification doesn't define what the expiry should be when a session cookie is returned.

The former behavior was to omit the field:

https://github.com/SeleniumHQ/selenium/wiki/JsonWireProtocol

> 1 When returning Cookie objects, the server should only omit an optional field if it is incapable of providing the information.

There's only this rule for adding a cookie:

https://w3c.github.io/webdriver/webdriver-spec.html#dfn-table-for-cookie-conversion

> When the cookie expires, specified in seconds since Unix Epoch. Defaults to 20 years into the future if omitted when adding a cookie.

I recommend to omit the field when a session cookie is returned to make it clear that there's no expiration  date.

I also think that setting the expiry to 20 years into the future when adding a cookie is a mistake, first because 20 years in the future is a valid expiry date and second because it will prevent the cookie from being deleted once the browser is closed.

shs96c:
Also Americanises the spelling of "serialised cookie".

Closes #957

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/969)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -6356,13 +6356,13 @@ must run the following steps:
 <p>The following <dfn>table for cookie conversion</dfn>
  defines the cookie concepts relevant to WebDriver,
  how these are referred to in [[!RFC6265]],
- what keys they map to in a <a>serialised cookie</a>,
+ what keys they map to in a <a>serialized cookie</a>,
  as well as the attribute-value keys needed
  when constructing a list of arguments for <a>creating a cookie</a>.
 
 <p>For informational purposes,
  the table includes a legend of whether the field is optional
- in the <a>serialised cookie</a> provided to <a>Add Cookie</a>,
+ in the <a>serialized cookie</a> provided to <a>Add Cookie</a>,
  and a brief non-normative description of the field
  and the expected input type of its associated value.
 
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -6443,14 +6443,14 @@
   <td>"<code>expiry</code>"
   <td>"<code>Max-Age</code>"
   <td>✓
-  <td>When the cookie expires,
-   specified in seconds since <a>Unix Epoch</a>.
-   Defaults to 20 years into the future
-   if omitted when <a>adding a cookie</a>.
+  <td>When the cookie expires, specified in seconds since
+   <a>Unix Epoch</a>.  Must not be set if omitted when
+   <a>adding a cookie</a>.
```
andreastt: I think this is incompatible with the description in the ‘table for cookie conversion’, which claims cookie expiry is non-optional.

The problem here is that we need to describe how to _deserialise_ a JSON Object we get from the client to a cookie representation as well as how to _serialise_ one from the browser’s cookie store. These two have different expectations with regards to required fields, as I understand it?

shs96c: The expiry is already marked as [optional in the table](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-table-for-cookie-conversion).

shs96c:
I think I've now responded to the review comment.

PR #987, opened 2017-07-27
minimize window: make it possible to restore the window

andreastt:
It is the intention that calling the Minimize Window command a second
time will restore it to its original dimensions on the visible screen.
It looks like an oversight that the specification does not cater for this.

This patch also makes it crystal clear that minimizing or restoring
the window must happen synchronously and not return before the
visibilitychange DOM event has fired.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/987)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -434,11 +434,11 @@ in the spec, as demonstrated in a (yet to be developed)
     <!-- queue --> <li><dfn><a href=https://infra.spec.whatwg.org/#queues>queue</a></dfn>
    </ul>
 
- <dt>Page Visibility
+ <dt>Interaction
  <dd>The following terms are defined in the
    Page Visibility Specification [[!PAGE-VISIBILITY]]
    <ul>
-     <li><dfn><a href="https://www.w3.org/TR/page-visibility/#dom-document-hidden">Document <code>hidden</code> attribute</a></dfn>
+    <!-- visibility hidden --> <li><dfn data-lt="visibility hidden"><a href="https://www.w3.org/TR/page-visibility/?csw=1#pv-page-hidden">Visibility state <code>hidden</code></a></dfn>    <!-- visibility state --> <li><dfn>Visibility state</dfn> being the <a href="https://www.w3.org/TR/page-visibility/?csw=1#dom-document-visibilitystate"><code>visibilityState</code></a> attribute on <a>Document</a>    <!-- visibility visible --> <li><dfn data-lt="visibility visible"><a href="https://www.w3.org/TR/page-visibility/?csw=1#pv-page-visible">Visibility state <code>visible</code></a></dfn>
    </ul>
 
  <dt>Selenium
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -434,11 +434,15 @@
     <!-- queue --> <li><dfn><a href=https://infra.spec.whatwg.org/#queues>queue</a></dfn>
    </ul>
 
- <dt>Page Visibility
+ <dt>Performance
```
shs96c: This doesn't seem like a correct change --- the very next line refers to the `Page Visibility Specification`.

webdriver-spec.html
```diff
@@ -3785,6 +3789,28 @@ <h2 data-dfn-for="NavigatorAutomationInformation">Interface</h2>
...
+<p>To <dfn>iconify the window</dfn>,
+ given an operating system level window
+ with an associated <a>top-level browsing context</a>,
+ run implementation-specific steps
+ to iconify, minimize, or hide the window
+ from the visible screen.
+ Do not return from this operation
+ until the <a>visibility state</a>
+ of the <a>top-level browsing context</a>’s <a>active document</a>
+ has reached the <a data-lt="visibility hidden">hidden</a> state.
```
shs96c: Or until the operation times out? Or will we caveat this out in the various window commands?

webdriver-spec.html
```diff
@@ -3785,6 +3789,28 @@ <h2 data-dfn-for="NavigatorAutomationInformation">Interface</h2>
...
+<p>To <dfn>restore the window</dfn>,
+ given an operating system level window
+ with an associated <a>top-level browsing context</a>,
+ run implementation-specific steps
+ to restore or unhide the window
+ to the visible screen.
+ Do not return from this operation
+ until the <a>visibility state</a>
+ of the <a>top-level browsing context</a>’s <a>active document</a>
+ has reached the <a data-lt="visibility visible">visible</a> state.
```
shs96c: As above with the timeout.

webdriver-spec.html
```diff
@@ -4018,9 +4043,18 @@ <h2 data-dfn-for="NavigatorAutomationInformation">Interface</h2>
...
+  of the <a>current top-level browsing context</a>’s <a>active document</a>:
+
+  <dl class=switch>
+   <dt><a data-lt="visibility normal">normal</a></dt>
+   <dt><a data-lt="visibility prerender">prerender</a></dt>
+   <dt><a data-lt="visibility unloaded">unloaded</a></dt>
+   <dd><p><a>Iconify the window</a>.
+
+   <dt><a data-lt="visibility hidden">hidden</a>
+   <dd><p><a>Restore the window</a>.
```
shs96c: Seems like this could be the main case, and the others handled with an `otherwise` case. This would handle the case of a vendor-prefixed DOMString, which is allowed as a valid visibility state.

webdriver-spec.html
andreastt: The Page Visibility spec is part of the Web Performance WG, but arguably this header should maybe say _Interaction_ since that’s the domain it represents.

Cf. the headers of the other sections: they’re not grouped by the name of the specification but by what general area/domain they are in.

webdriver-spec.html
andreastt: Yes, this is missing. Added.

webdriver-spec.html
andreastt: And done.

webdriver-spec.html
andreastt: The visibility state is always one of these, but it makes sense to use “otherwise”.

shs96c:
Once this validates, it's good to go. Thank you!

andreastt:
Thanks for the review (-:

Issue #989, opened 2017-07-29
Stale Element distinction


titusfortner:
Shouldn't this example throw a stale element reference error and not an invalid element state error?

In chapter [6. Protocol](https://w3c.github.io/webdriver/webdriver-spec.html#handling-errors#x6.-protocol), section [6.6 Handling Errors](https://w3c.github.io/webdriver/webdriver-spec.html#handling-errors#x6.6-handling-errors):
> A command could not be completed because the element is in an invalid state, e.g. attempting to click an element that is no longer attached to the document.
>

andreastt:
Yes, or more precisely: the example for the _invalid element state_ error is wrong.

PR #990, opened 2017-07-29
invalid element state: correct description

andreastt:
That an element is in an invalid state does not mean it has been detached
from the document.  It usually means the element is in a state in which
it is impossible for WebDriver to interact with it, e.g. if you try to use
the high-level Element Click command to click an element that is disabled.

Fixes: https://github.com/w3c/webdriver/issues/989
Reported-by: @titusfortner

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/990)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -1550,10 +1550,10 @@ in the spec, as demonstrated in a (yet to be developed)
   <td><dfn>invalid element state</dfn>
   <td>400
   <td><code>invalid element state</code>
-  <td>A <a>command</a> could not be completed
-   because the element is in an invalid state,
-   e.g. attempting to click an element
-   that is no longer attached to the <a>document</a>.
+  <td>The target <a>element</a> is in an invalid state,
+   rendering it impossible to interact with,
+   for example if you <a data-lt="element click">click</a>
+   a <a>disabled</a> <a>element</a>.
  </tr>
 
  <tr>


titusfortner:
Drivers currently do not error when clicking an element in a disabled state. When I last asked about it, I was told that was intended design. Is this not correct? 

It would simplify my Watir code if there is an error, which is why I remember specifically asking about it. I can't find my question in #selenium logs, so I'm not sure where/when I asked. Pretty sure it was @jimevans who answered me, though. Maybe I asked in person when we were both in SF.

andreastt:
You’re right, the only two references I can find to _invalid element state_ is under Element Clear:

> If element is not editable and is not resettable return an error with error code invalid element state. 

and

> If element is disabled, read only, or has pointer events disabled, return error with error code invalid element state. 

I _thought_ the intention was for high-level commands to alert you when something is potentially wrong. This is the reason why the low-level action click primitive doesn’t error on clicking a disabled element.

shs96c:
The high-level "do as I mean" commands are meant to emulate user behaviour. For an `Element Click` while it's a waste of effort to click on a disabled element, it's a thing a user could do. For `Element Send Keys` if there's no way that the intention of actually setting content can be fulfilled we're clearly violating expectations, so we do the check for keyboard interactability. 

It may seem counter-intuitive to only do the check when typing and not clicking, but bug reports and the way that WebDriver is used suggests that this behaviour matches user expectations most closely.

shs96c:
Closing, as the example given isn't correct.

PR #994, opened 2017-08-03
Clearly indicate that webdriver can access HttpOnly cookies

### LINKED ISSUE #971: "All associated cookies" must include http-only cookies.
In chapter [16. Cookies](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-all-associated-cookies#x16.-cookies):
> all associated cookies

It's not clear to me that this is the case from our definition.

shs96c:
Closes #971

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/994)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -6574,8 +6574,10 @@ must run the following steps:
 <p>To get <dfn data-lt="associated cookies">all associated cookies</dfn> to a <a>document</a>,
  the user agent must return the enumerated set of <a>cookies</a>
  that meet the requirements set out in the first step of the algorithm in [[RFC6265]] to
- <a>compute <code>cookie-string</code></a> for an ‘HTTP API’,
- from the <a>cookie store</a> of the given <a>document</a>’s <a>address</a>.
+ <a>compute <code>cookie-string</code></a> for an ‘HTTP API’, from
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -6574,8 +6574,10 @@ <h2 data-dfn-for="NavigatorAutomationInformation">Interface</h2>
 <p>To get <dfn data-lt="associated cookies">all associated cookies</dfn> to a <a>document</a>,
  the user agent must return the enumerated set of <a>cookies</a>
  that meet the requirements set out in the first step of the algorithm in [[RFC6265]] to
- <a>compute <code>cookie-string</code></a> for an ‘HTTP API’,
- from the <a>cookie store</a> of the given <a>document</a>’s <a>address</a>.
+ <a>compute <code>cookie-string</code></a> for an ‘HTTP API’, from
+ the <a>cookie store</a> of the
+ given <a>document</a>’s <a>address</a>. The returned cookies must
+ include <code>HttpOnly</code> cookies.
```
andreastt: Use `<a>cookie HTTP only</a>` here. We have a reference to it from the relevant RFC.

andreastt:
r+wc

webdriver-spec.html

PR #995, opened 2017-08-03
Allow Set Window Rect to accept any valid Number

### LINKED ISSUE #988: {Get,Set} Window Rect type parity
The Get Window Rect command can potentially return `width` and `height` as floats, which in JSON gets encoded as a JSON Number (can be pretty much anything). The Set Window Rect command, however, only accepts positive integers for these two fields which means you can’t directly take the return value from Get Window Rect and pass it on to Set Window Rect.

As far as I can tell, there is however no reason why the remote end can’t accept a float value, since this is supported by `window.resizeTo`.

shs96c:
Closes #988

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/995)


andreastt:
Thanks for patching this, @shs96c!

Issue #912, opened 2017-05-09
Behaviour of unhandled prompt handler diverges from Selenium


shs96c:
In Selenium, setting the unhandled prompt handler behaves in the same way as in the spec, but with the important difference that if a prompt is present, an error will always be returned. In the spec, we return `success` if the handler is invoked.

The intention is to allow a user to re-use their existing WebDriver instance without needing to create a new one (which can be a slow and expensive process), but they should still be notified that the driver is in an unexpected state. It would seem that the Selenium behaviour better meets that intention, and I suspect that the [wording in the user prompts section](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-handle-any-user-prompts) is an oversight.

Test for Selenium behaviour:
https://github.com/SeleniumHQ/selenium/blob/master/dotnet/test/common/AlertsTest.cs#L681

Existing implementation in the legacy FirefoxDriver (all code paths lead to an error being returned to the user if an unexpected modal is open):
https://github.com/SeleniumHQ/selenium/blob/master/javascript/firefox-driver/js/nsCommandProcessor.js#L478
https://github.com/SeleniumHQ/selenium/blob/master/javascript/firefox-driver/js/modals.js#L261

shs96c:
Relevant Selenium issue: https://github.com/SeleniumHQ/selenium/issues/3972

andreastt:
> The intention is to allow a user to re-use their existing WebDriver instance without needing to create a new one (which can be a slow and expensive process), but they should still be notified that the driver is in an unexpected state.

The driver doesn’t get into an unrecoverable state when a user prompt is present. You can call the Dismiss Alert command to close any prompt if a command returns with an ‘unexpected alert open’ error, so I don’t understand why a user would have to create a new session?

> It would seem that the Selenium behaviour better meets that intention, and I suspect that the wording in the user prompts section is an oversight.

From what you’re writing, I don’t understand what behaviour you’re seeking.

What I find mildly surprising about the user prompt handler as it is defined right now, is that the prompt is dismissed if a handler is not set. I would only expect it to dismiss a prompt if the handler is in the "dismiss" state. In the default/unknown state, I would expect it to just return an error.

shs96c:
Consider these tests, which share a common WebDriver instance (written in the Java style, but the intent should be clear):

```
@Test
public void failsBecauseOfAnUnexpectedAlert() {
  driver.get("http://www.example.com/");
  WebElement element = driver.findElement(By.id("cake"));
[...]

PR #996, opened 2017-08-04
Add states to prompt handling to allow notification of handled prompts

jimevans:
Setting a user prompt handler accepts or dismisses a prompt with no
notification to the user that the prompt was handled. In the primary use
case of testing, this hides important information from the user, that a
prompt the test did not expect was encountered, and handled. If the user
wishes to have their test ignore this case, they should make a fully
informed and conscious decision to do so. This change adds options to the
user prompt handler to allow for the case to inform the user of the
handled propmt via an error. To maintain compatibility with previous spec
language, the added values are the ones that describe the new behavior.

Closes issue #912.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/996)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -1,4 +1,4 @@
-<!doctype html>
+﻿<!doctype html>
 <html
  data-issue-url="https://github.com/w3c/webdriver/"
  data-issue-param-milestone="Level 1">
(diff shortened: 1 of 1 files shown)


shs96c:
I'm in favour of this change, as it allows implementations to bring themselves into sync with the existing behaviour used by selenium implementations.

shs96c:
Discussed with @AutomatedTester on the IRC. Approved and landing.

PR #998, opened 2017-08-09
Make definition of `editable` clearer

### LINKED ISSUE #884: Wrong use of hasAttribute for editable element
In chapter [11. Elements](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-editable#h-elements):
> element is an input element for which hasAttribute "checked" is false and for which the input event applies. 

We should reuse the definitions from the HTML spec to refer to what it means for an element not be editable. That `<input checked>` `hasAttribute` "checked" does not make it a checkbox.

shs96c:
Closes #884

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/998)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -4280,9 +4280,8 @@ by following these steps:
  if one or more of the following conditions are met:
  <ul>
   <li><p><var>element</var> is an <a><code>input</code> element</a>
-   for which <a><code>hasAttribute</code></a> <code>"checked"</code>
-   is <code>false</code> and for which the <a><code>input</code>
-   event applies</a>.
+   for which the <a><code>input</code> event applies</a>, and which is
+   neither in the <a>checkbox</a> state nor <a>radio button</a> state.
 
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -4280,9 +4280,8 @@ <h2 data-dfn-for="NavigatorAutomationInformation">Interface</h2>
  if one or more of the following conditions are met:
  <ul>
   <li><p><var>element</var> is an <a><code>input</code> element</a>
-   for which <a><code>hasAttribute</code></a> <code>"checked"</code>
-   is <code>false</code> and for which the <a><code>input</code>
-   event applies</a>.
+   for which the <a><code>input</code> event applies</a>, and which is
+   neither in the <a>checkbox</a> state nor <a>radio button</a> state.
```
AutomatedTester: I am not sure what "state" means here.

shs96c: It's from the HTML5 spec, and is a term we've used elsewhere in the spec.

Issue #895, opened 2017-04-17
Need to wait for a paint before taking a screenshot


jgraham:
https://github.com/jugglinmike/chrome-screenshot-race/issues/1 details an issue where using ChromeDriver to take screenshots on page load produces unreliable behaviour because it isn't waiting for the content to paint before taking the screenshot. The specification should require webdriver implementations to wait for a paint before taking the screenshot, otherwise the implementation isn't reliable.

cc @jugglinmike

gsnedders:
Related to #893.

jgraham:
This is related to #893 but somewhat different because it isn't about resources that aren't loaded by `load` but about the fact that `load` firing doesn't guarantee a paint has happened.

andreastt:
Would adding the step

> Wait until the user agent event loop has spun enough times to process the DOM events generated by the previous step.

from the [Element Click](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-element-click) command help?

We are somewhat vague about what that means, but in practice, what Marionette does wait for `window.requestAnimationFrame` to return:

```js
interaction.flushEventLoop = function* (win) {
  let unloadEv;

  return new Promise((resolve, reject) => {
    if (win.closed) {
      reject();
      return;
    }

    unloadEv = reject;
    win.addEventListener("unload", unloadEv, {once: true});

    win.requestAnimationFrame(resolve);
  }).then(() => {
    win.removeEventListener("unload", unloadEv);
  });
};
```

jgraham:
No. In gecko terms we should wait for MozAfterPaint (although I think that rAF is equivalent). But I'm not sure if it's guaranteed that a paint actually happens before that runs. The spec sort of implies that it probably ought to, but maybe in practice it doesn't.

shs96c:
We now using the same language as `requestAnimationFrame`, so there shouldn't (ha!) be pending paint events. Waiting for all content to be loaded (including non-critical ones) is out of scope for Level 1. Pushing to level 2.

foolip:
@Hexcles and I are discussing how to deal with https://crbug.com/507054 and need to understand the exact timing of [Take Screenshot](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-take-screenshot) to do that.

> In gecko terms we should wait for MozAfterPaint (although I think that rAF is equivalent)

[...]

PR #999, opened 2017-08-09
Layout should be complete before screenshots are taken.

shs96c:
Starts to address #895

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/999)


webdriver-spec.html
```diff
@@ -9114,7 +9114,9 @@ <h2 data-dfn-for="NavigatorAutomationInformation">Interface</h2>
  <li><p>If the <a>current top-level browsing context</a> is <a>no longer open</a>,
   return <a>error</a> with <a>error code</a> <a>no such window</a>.
 
- <li>Let <var>root rect</var> be the <a>current top-level browsing context</a>’s
+ <li><p>Wait in an implementation-specific way for layout to complete.
```
AutomatedTester: I think we need to talk about painting and compositor here instead of layout. We might want to base it around requestAnimationFrame which I believe would fire when "layout" is done

shs96c: Some help with the language would be useful.

Issue #893, opened 2017-04-10
Unable to take a screenshot of a fully loaded page


gsnedders:
At the moment, with WebDriver, you can trivially wait until after the load event has fired before taking a screenshot. However, not everything delays the load event (see HTML's notion of "critical subresources" and "delays the load event"). Anything you can detect whether or not it has loaded using the DOM can equally be dealt with through the script APIs.

However, CSS is especially awkward because of https://github.com/w3c/csswg-drafts/issues/1088. At the moment, Edge doesn't block the load event on background-image and list-style-image and these are undetectable through the DOM.

We need some proper solution to this problem in a general sense. As a web developer, to do any sort of visual testing, you want to be able to take a screenshot of a page after everything has loaded (of course, this can get complicated by long polling XHR, WebSockets, etc.).

I think there's at least a couple of options options:

 * Convince everyone (WGs, implementers) that the page should be in a stable state after the load event has fired
 * Add some custom notion to WebDriver of loadedness (OperaDriver has something like this with `opera.idle`)

Notably, this is causing problems for wptrunner running reftests, as we need some way to take a screenshot after everything has loaded. (https://github.com/w3c/web-platform-tests/issues/5412)

shs96c:
I realise that your first solution was probably flippant, and your second the desired outcome, but I feel the first _is_ the correct solution. 

The problem is that for anyone to actually implement the second suggestion, the first must be true. Worse, as new standards emerge, _someone_ would have to be responsible for making sure that the `load` event actually meant `loaded`. This would either mean endless minor updates to this spec (through the w3c approval process to get to Recommendation) or the relevant specs should include language to indicate how they influence the `load` event.

From the linked WPT bug, it seems like at least some browser vendors have considered this and are treating all "remote loads" as impacting the `load` event (which seems sensible to me), and from w3c/csswg-drafts#1088 it sounds like the CSS folks are considering the matter too, if I read @fantasai's [comment](https://github.com/w3c/csswg-drafts/issues/1088#issuecomment-290451355) correctly.

Closing this here, but please feel free to reopen if you feel that more discussion is necessary or more reasoning is lax.

gsnedders:
@shs96c I'd on the whole agree the first solution is probably the correct solution!

[...]

PR #1000, opened 2017-08-09
Add back the `conservative` page loading strategy

shs96c:
This waits for all resources of a page to have finished
loading, and not just for the readyState to reach
"complete".

Addresses #893.

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -3034,6 +3034,12 @@ with a "<code>moz:</code>" prefix:
   <td><dfn data-lt="normal page loading strategy">normal</dfn>
   <td>"<code>complete</code>"
  </tr>
+
+ <tr>
+  <td>"<code>conservative</code>"
+  <td><dfn data-lt="conservative page loading strategy">conservative</dfn>
+  <td>"<code>complete</code>"
+ </tr>
 </table>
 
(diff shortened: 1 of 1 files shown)


shs96c:
Marking as an L2 fix. It's deemed to be a new feature, which is not in keeping with the CR period.

jgraham:
Per the testing policy, this PR can't be accepted without a corresponding test in the web-platform-tests reposiotry.

andreastt:
@shs96c Do you have time to also write the tests for this?

Because we don’t currently have any tests for the other page loading strategies it may be required to do those first, which could be a bit of work. @whimboo has written some excellent Marionette navigation tests that we would be happy to share with WPT, but we need to CC-0 license it first. @whimboo, would that be easy to do?

whimboo:
> @whimboo has written some excellent Marionette navigation tests that we would be happy to share with WPT, but we need to CC-0 license it first. @whimboo, would that be easy to do?

I wrote those tests as employee of Mozilla, so you can automatically handle those as CC-0. The tests can be found here:

https://hg.mozilla.org/mozilla-central/file/default/testing/marionette/harness/marionette_harness/tests/unit/test_navigation.py#l615

PR #1001, opened 2017-08-09
Element interaction: BODY only interactable in HTML docs

### LINKED ISSUE #888: <body> element should only be keyboard-interactable in HTML or XHTML
In chapter [11. Elements](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-keyboard-interactable-element#h-elements), section [11.1 Element Interactability](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-keyboard-interactable-element#h-element-interactability):
> A keyboard-interactable element is defined to be any element that has a focusable area, or is the body element.

The `<body>` element should only be considered keyboard-interactable when the document is in HTML- or XHTML mode. It does not make sense for it to be keyboard-interactable in SVG, for example.

shs96c:
Closes #888

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1001)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -4319,7 +4319,7 @@ by following these steps:
 
 <p>A <dfn data-lt="keyboard-interactable">keyboard-interactable element</dfn>
  is any <a>element</a> that has a <a>focusable area</a>,
- or is the <a><code>body</code> element</a>,
+ is a <a><code>body</code> element</a>,
  or is the <a>document element</a>.
 
 <p>An <a>element</a>’s <dfn data-lt="center point">in-view center point</dfn>


webdriver-spec.html
```diff
@@ -4319,7 +4319,7 @@ <h2 data-dfn-for="NavigatorAutomationInformation">Interface</h2>
 
 <p>A <dfn data-lt="keyboard-interactable">keyboard-interactable element</dfn>
  is any <a>element</a> that has a <a>focusable area</a>,
- or is the <a><code>body</code> element</a>,
+ is a <a><code>body</code> element</a>,
```
AutomatedTester: I don't think that fixes the issue. It appears to make body or document elements keyboard interactable. I think we just need to say that `<a><code>body</code> element</a> if the document's namespace is html` linking to https://infra.spec.whatwg.org/#html-namespace

shs96c: The "body" that we link to is in the HTML5 spec. By definition it'll be an HTML document.

PR #1004, opened 2017-08-10
Made numerous hygiene passes over the doc

InstyleVII:
Added lang=en to html element
Updated links to https where appropriate
Added quotes to HTML properties where appropriate, this should allow for editor's to parse them properly
Renamed Getting Page Source to Get Page Source to be in line with other command names

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1004)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -642,7 +642,7 @@ in the spec, as demonstrated in a (yet to be developed)
 <p>WebDriver aims to allow other standards
  to <a data-lt="extension commands">provide extensions</a> to support new functionality,
  and to allow conformance tests that cannot be implemented
- entirely in <a href="https://tc39.github.io/ecma262"/>ECMAScript</a>
+ entirely in <a href="https://tc39.github.io/ecma262">ECMAScript</a>
  to be written in a vendor-neutral way.
 </section>
 </section> <!-- /Design Notes -->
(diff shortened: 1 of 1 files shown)


AutomatedTester:
there are travis failures, please check it before I do a review

InstyleVII:
@AutomatedTester Fixed now! Adding proper quotes highlighted that some anchor tags had ending '/' in the opening tag (I'm guessing the Travis parser couldn't parse them properly before which is why it never came up).

andreastt:
Why do you need your editor “parse links”? AIUI links without quotes are not invalid, unless they contain an ampersand.

Disregarding the search-and-replace for links across the doc, I would prefer if you did smaller changes in individual commits in the future. When you rename _Getting Page Source_ to _Get Page Source_ in the middle of an enormous commit, it is hard to follow the commit history back in time to know when the change was made.

InstyleVII:
I don't need my editor to parse links, it's just a quality of life thing. It was also inconsistent throughout the doc with regards to using/not using them, although maybe that was only for links with ampersands. It was also for readability of the markup, having quotes surrounding the links makes it a little more clear where they start/stop, and it was a personal preference. Haven't seen markup before so consistently never use quotes for html property values.

As for doing the "Getting Page Source" change with this one, fair point. I can split that out if you'd like. It was more that I just noticed it while I was giving the doc a pass and updated it. Also didn't want to add to the workload by adding another PR but will do so going forward.

shs96c:
Don't worry about the workload :) I'd much rather many focused PRs than fewer PRs without focus.

AutomatedTester:
@InstyleVII Can you redo this PR so that we can see individual commits for what you're fixing. It will significantly improve the chances of this PR landing

andreastt:
@InstyleVII Do you have time to rework your hygine passes? I suggest submitting separate PRs with smaller changes so they are easier to review.

I’m going to close this PR because it can’t be landed in its current form anyway. I still think these things should be fixed, however.

PR #1012, opened 2017-08-16
Socks Proxy requires Socks Version

### LINKED ISSUE #1008: Clarify which proxy capabilities are required for each proxy type
Currently the spec is not clear about which proxy capabilities are required or optional depending on the type of proxy selected. 

Should we have another column in the [proxy caps table](https://w3c.github.io/webdriver/webdriver-spec.html#proxy) or better clarify with wording inside the remote steps?

@shs96c and @AutomatedTester what do you prefer?

shs96c:
If someone tries to set the socks proxy they
also need to declare the socks version.

Closes #1008

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1012)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -2048,6 +2048,12 @@ with a "<code>moz:</code>" prefix:
   "<code>proxyType</code>" return an <a>error</a> with <a>error
   code</a> <a>invalid argument</a>.
 
+ <li><p>If the result of <a>getting a property</a>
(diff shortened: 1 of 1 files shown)


whimboo:
This does not close issue #1008. It just fixes a single part.

shs96c:
Discussed on IRC. This PR improves things by making sure that a proxy is valid.

PR #1017, opened 2017-08-17
Allow getting and setting of implementation-specific timeouts

### LINKED ISSUE #1016: Cannot get implementation-specific timeouts
The definition for session timeouts allows for implementation-specific timeouts: 

https://w3c.github.io/webdriver/webdriver-spec.html#dfn-session-timeouts-configuration

We should be returning these implementation-specific timeouts whenever we return a timeouts object (so in New Session and Get Timeous)

shs96c:
This is clearly meant to be allowed, since the spec
has language to say that implementation-specific
timeouts are supported.

https://w3c.github.io/webdriver/webdriver-spec.html#dfn-session-timeouts-configuration

Closes #1016

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1017)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -2716,6 +2716,17 @@ with a "<code>moz:</code>" prefix:
       set the <a>current session</a>’s <a>session script timeout</a>
       to the value of property "<code>script</code>".
       Otherwise, set the <a>session script timeout</a> to 30,000 milliseconds.
+
+     <li><p>For every <var>timeout</var> in this implementation's implementation-specific timeouts:
+      <ol>
+       <li>Let <var>value</var> be the numeric property with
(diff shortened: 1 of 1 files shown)


shs96c:
The specification doesn't provide any mechanism for getting the implementation-specific timeouts. It makes sense to return these as part of the timeouts object. `Get Timeouts` doesn't actually return them, and that blob is also what gets returned from `New Session`.

Presumably if someone sets implementation-specific timeouts in `New Session` they want them to be used. It goes no further than saying timeouts need to numbers.

shs96c:
If this PR lands, then #1018 should be closed.

andreastt:
@AutomatedTester This should get reverted. Per @shs96c’s comment _either_ this or https://github.com/w3c/webdriver/pull/1018 should land; not both.

PR #1018, opened 2017-08-17
Timeouts: tighten definition to exclude implementation-specific timeouts

### LINKED ISSUE #1016: Cannot get implementation-specific timeouts

shs96c:
Closes #1016

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1018)


shs96c:
If this PR lands, then #1017 should be closed.

PR #1022, opened 2017-08-19
set window rect, maximize window: restore window first

andreastt:
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1022)


andreastt:
@shs96c It had a conflict with the previous patch to Maximize Window.

PR #1024, opened 2017-08-22
Return an unknown error if an http request makes no sense

shs96c:
Resolves issue 1 on the spec. The language is wooly, but then
all the language around how to handle the network connection
is wooly. The key benefit of unknown error is that it sets the
http status code to 500, which is what I'd expect to happen.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1024)


andreastt:
This makes sense to me.

PR #1028, opened 2017-08-23
[Element Click]: modify how we wait for navigation

shs96c:
Avoids wooly language about waiting for navigation to begin
by adding woooly language about wait for navigation to begin.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1028)


webdriver-spec.html
```diff
@@ -5603,20 +5603,13 @@ <h2 data-dfn-for="NavigatorAutomationInformation">Interface</h2>
...
-
-   <li><p><a>Try</a> to <a>wait for navigation to complete</a>.
-  </ol>
-
-  <p class=example>Note that one may not know
-   if a click causes <a>navigation</a>.
-   It is safe to run the <a>post-navigation checks</a>
-   and <a>waiting for the navigation to complete</a>
-   even though navigation <em>did not</em> occur.
+ <li><p><a>Try</a> to <a>wait for navigation to complete</a>.
```
whimboo: Please note that there can be a delay in when the navigation actually get started. How should this be handled, and that we do not return too early? In Marionette we currently use a hardcoded delay of 200ms for now. Maybe we can spec something better?

whimboo: Also waiting in general is not such a great idea given that not all clicks will trigger a navigation.

andreastt: There ought to be an issue blocking L1 if this is to make it into the spec before it is frozen for REC next Thursday.

whimboo: Ok, I will file one.

whimboo: https://github.com/w3c/webdriver/issues/1046

PR #1043, opened 2017-08-24
Status Codes: mark not found resources as 404

shs96c:
And mark "unknown command" as 501 ("not implemented")

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1043)


shs96c:
We may just have to live with our choices of status codes.

andreastt:
Patched the webdriver crate in https://github.com/mozilla/webdriver-rust/pull/117.

PR #1044, opened 2017-08-24
Clarify that noProxy expects a list of comma-separated entries.

whimboo:
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1044)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -1986,8 +1986,9 @@ with a "<code>moz:</code>" prefix:
  <tr>
   <td><dfn><code>noProxy</code></dfn>
   <td>array
-  <td>Lists the address for which the proxy should be bypassed when
-   the <a><code>proxyType</code></a> is "<code>manual</code>".
+  <td>Comma-separated list of hosts for which the proxy should be bypassed if
+   <a><code>proxyType</code></a> is equal to "<code>manual</code>".
+  <td>
   <td>A <a>List</a> containing any number of any
    of <a>domains</a>, <a>IPv4 addresses</a>, or <a>IPv6 addresses</a>.
  </tr>
@@ -6049,7 +6050,7 @@ must run the following steps:
 
    <dt><var><a>element</a></var> is <a>content editable</a>
    <dd>Set the text insertion caret after any child content.
-   
+
    <dt>Otherwise
    <dd>
     <ol>


shs96c:
@whimboo is a Mozilla employee, which is a w3c member organisation.

whimboo:
@shs96c I got my account validated and connected it to Github. Looks like we are set for a review now.

webdriver-spec.html
```diff
@@ -1986,8 +1986,9 @@ <h2 data-dfn-for="NavigatorAutomationInformation">Interface</h2>
  <tr>
   <td><dfn><code>noProxy</code></dfn>
   <td>array
-  <td>Lists the address for which the proxy should be bypassed when
-   the <a><code>proxyType</code></a> is "<code>manual</code>".
+  <td>Comma-separated list of hosts for which the proxy should be bypassed if
```
shs96c: The `Valid Values` column tells us that this is not a comma-separated list, but a JSON `List` type:

> A List containing any number of any of domains, IPv4 addresses, or IPv6 addresses.

If the driving force behind this is to make the implementation less fussy (for example, just a `String` that the implementation can do what it wants with), then you also need to modify the `Valid Value` column.

whimboo: Oh, so it means we are requiring something like `{"noProxy": ["foo.bar", "127.0.0.1"]}`?

With that the different driver implementations would have to join all entries into a single string with a separator the browser supports. I would be fine with that, and would update the PR to only make this entry sound similar to `proxyAutoconfigUrl`.

whimboo: @shs96c any further feedback? If the last comment applies to how it should look like, we can close this issue. Or should we mention "JSON list" similar to other parts of the spec?

shs96c: Yes, the spec is saying that we should have something like `{"noProxy": ["foo.bar"]}`

whimboo: So should we update to say "JSON list"? I can update my PR appropriately.

shs96c:
Closing this PR, as @whimboo requested. The spec is sufficiently clear.

whimboo:
@shs96c so no need to make it a "JSON List" what is the term we use at other places?

andreastt:
It already says that the valid input is a JSON List?

shs96c:
Yes. Under `noProxy`'s "valid values"

> A List containing any number of any of domains, IPv4 addresses, or IPv6 addresses.

With `List` linking to the JSON List.

PR #1048, opened 2017-08-25
Make note about platform names less confusing

### LINKED ISSUE #1047: platformName uses normative text in a note
screenshot-2017-08-25-155107

Under the `platformName` capability there is a note which is supposed to give an indication what input values the remote end should accept for different classes of systems.

First of all, I don’t understand why this is a note, and second I don’t understand why the note uses normative text.

shs96c:
Closes #1047

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1048)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -2387,8 +2387,9 @@ with a "<code>moz:</code>" prefix:
 
       <div class=note>
        <p>The following platform names are in common usage with
-        well-understood semantics and are to be treated as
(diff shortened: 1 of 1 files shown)


Debanjan-B:
@andreastt I think without a valid set of credentials I won't be able to review the Code Merge. Can you please update the highlights of the merge here or in #1047 or in [/mozilla/geckodriver/issues/897](https://github.com/mozilla/geckodriver/issues/897). Thanks

PR #1053, opened 2017-08-29
Define what response is in post-navigation steps. Fixes #1030

### LINKED ISSUE #1030: Post-navigation checks need to define what "response" is.

AutomatedTester:
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1053)


AutomatedTester:
@andreastt @jgraham do you remember the idea behind the issue in #1030 because having to introduce the idea of `<var>response</var>` feels fraught with lots of little issues that will continue to creep in.

PR #1054, opened 2017-08-29
Specify default ports in proxies if omitted.

### LINKED ISSUE #1049: Optional port for proxy host (manual mode) should default to schemes default port
As specified in the [spec](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-host-and-optional-port):

> A host and optional port is defined as being a valid host, optionally followed by a colon and port.

It's unclear what behavior is expected when the port is not given or 0. 

Specifically for Firefox in both cases the proxy host would not be used for the given scheme. It doesn't fallback automatically to the default port of the scheme. Other browsers seem to do that.

Should we add that if the port is not in range of 1 - 65535 the scheme's default port is used?

shs96c:
By linking to the URL spec.

Closes #1049

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1054)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -566,6 +566,7 @@ in the spec, as demonstrated in a (yet to be developed)
   <ul>
    <!-- Absolute URL --> <li><dfn><a href=https://url.spec.whatwg.org/#syntax-url-absolute>Absolute URL</a></dfn>
    <!-- Absolute URL with fragment --> <li><dfn><a href=https://url.spec.whatwg.org/#syntax-url-absolute-with-fragment>Absolute URL with fragment</a></dfn>
+   <!-- Default port --> <li><dfn><a href="https://url.spec.whatwg.org/#default-port">Default port</a></dfn>
    <!-- Domain --><li><dfn data-lt="domains"><a href="https://url.spec.whatwg.org/#concept-domain">Domain</a></dfn>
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -2018,10 +2022,12 @@ <h2 data-dfn-for="NavigatorAutomationInformation">Interface</h2>
 
 </table>
 
-<p>A <dfn>host and optional port</dfn> is defined as being a
-  valid <a>host</a>, optionally followed by a colon
-  and <a>port</a>. The <a>host</a> may
-  <a data-lt="includes credentials">include credentials</a>.
+<p>A <dfn>host and optional port</dfn> for a <var>scheme</var> is
```
AutomatedTester: I really think we can drop this whole definition because it is a redefinition really of [URL](https://url.spec.whatwg.org/#concept-url). This means that we can, for the other parts, instead of
`<td>A <a>host and optional port</a> with an <a>undefined</a> scheme.` 
we can say 
`<td>A <a>URL</a> with an <a>undefined</a> scheme.`

shs96c: But it's not a complete URL, since we omit the scheme in valid values for a proxy. In order to specify the default port, we then need to perform linguistic shenanigans like saying `a URL with an undefined scheme but with port set to the default port for "scheme" if the port is omitted`, which ain't elegant at all.

PR #1072, opened 2017-08-30
is stale: take document into account

andreastt:
The check for whether an element is connected does not take into account
that the document may have changed.  This is a special case for WebDriver,
where element references are preserved across documents.

To work around this issue we compare the element's node document
(element.ownerDocument) to the current browsing context's active document
(element.ownerDocument.defaultView.document).

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1072)


webdriver-spec.html
```diff
@@ -4352,8 +4352,9 @@ <h2 data-dfn-for="NavigatorAutomationInformation">Interface</h2>
  <li><p>Return <a>success</a> with data <var>element</var>.
 </ol>
 
-<p>A <a>element</a> <dfn>is stale</dfn>
- if its <a>context object</a> is not <a>connected</a>.
+<p>An <a>element</a> <dfn>is stale</dfn>
+ if its <a>node document</a> is not the <a>active document</a>
```
shs96c: `node document` is not defined.

andreastt: Added a reference to DOM.

PR #1075, opened 2017-08-30
stop styling null and undefined types

andreastt:
In specifications we want to compare ourselves with, null and undefined
are not given a special <code> tag.  Null and undefined types are no
different to strings or arrays.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1075)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -939,28 +939,28 @@ in the spec, as demonstrated in a (yet to be developed)
     from <var>url variables</var>.
 
    <li><p>Let the <a>current session</a> be the <a>session</a>
-     with <a data-lt="session id">ID</a> <var>session id</var> in the
-     list of <a>active sessions</a>, or <a><code>null</code></a> if
-     there is no such matching <a>session</a>.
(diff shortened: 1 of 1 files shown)


shs96c:
I'm unsure about this. Although the other specs may not highlight them, both `null` and `undefined` are values, and we are styling values. They are different from String and Array, as those are both types (as @jgraham pointed out on IRC, the associated types are `Null` and `Undefined`)

andreastt:
At the moment we use the computing science terminology of null and undefined interchangeably with the JSON types. If you wanted to be super-correct here you would want to distinguish between nulls and undefineds used internally for algorithms and those passed in as input or emitted as output.

In any case we are referencing the JSON definitions of both where it matters, which we also do for String and Array. I can certainly go through the specification and add plumbing to allow internal-nulls and internal-undefineds to be implicitly coerced to JSON Nulls and Undefineds when a response is being sent, as well as fix s/null/Null and s/undefined/Undefined/ where that matters for output?

shs96c:
Let's look like everyone else for now.

PR #1081, opened 2017-09-01
Differentiate connected and unconnected elements for getting a web element.

jgraham:
In the existing spec, getting an element will return any element,
possibly in any browsing context, regardless of whether it's connected
to the document tree. This is clearly wrong:

 * Only the list of known elements for the current browsing context
   should be used. That is intended by the current spec, but the text
   is unclear.

* For operations that depend on layout, interation, scrolling, etc. we
  should ignore elements that are not connected to the document tree
  (e.g. the result of createElement). For operations that only depend
  on DOM or scripting, the possibility of working with such
  disconnected elements makes sense. So we need to differentiate these
  cases.

This commit fixes both the above issues.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1081)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -4282,28 +4282,42 @@ with a "<code>moz:</code>" prefix:
 <p>An ECMAScript <a>Object</a> <dfn>represents a web element</dfn>
  if it has a <a>web element identifier</a> <a>own property</a>.
 
-<p>Each <a>browsing context</a> has an associated list of
- <dfn data-lt="known element">known elements</dfn>.
+<p>Each <a>browsing context</a> has an associated <dfn>list of
+ known elements</dfn>.
  When the <a>browsing context</a> is <a>discarded</a>,
- the list of <a>known elements</a> is discarded along with it.
+ the <a>list of known elements</a> is discarded along with it.
 
-<p>To <dfn>get a known element</dfn>
- with argument <var>reference</var>,
- let <var>element</var> be the <a>known element</a>
- which <a>web element reference</a> matches <var>reference</var>.
- If there are none, return <a>error</a>
- with <a>error code</a> <a>no such element</a>.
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -4282,36 +4282,50 @@ <h2 data-dfn-for="NavigatorAutomationInformation">Interface</h2>
...
+
+<p>To <dfn>get a known connected element</dfn> with
+argument <var>reference</var>, run the following steps:
+<ol>
+  <li>Let <var>element</var> be the result of <a>trying</a> to <a>get
+  a known element</a> with argument <var>reference</var>.
+  <li>If <var>element</var> is not <a>connected</a>
+  return <a>error</a> with error code <a>stale element reference</a>.
+  <li>Return <a>success</a> with <var>element</var>.
+</ol>
```
andreastt: @jgraham @shs96c A connected element doesn’t take into account that the document may have changed, see the definition of ‘is stale’.

WebDriver is special in this regard because element references are kept across documents and the DOM definition of ‘connected’ doesn’t take this into account.

jgraham: We should change the definition of "is stale" to just check if the node's document is the active document, or just inline that check.

andreastt: I added that earlier this week:

> An element is stale if its node document is not the active document or if its context object is not connected.

And step 2 of ‘get a known element’ says:

> If element is stale, return error with error code stale element reference. 

And the new ‘get a known connected element’ you added here calls that in step 1, so I think we are OK here! 👍 

Sorry for the false warning.

jgraham: The problem here is that "get a known element" is checking for connectedness, when that isn't necessary. It should be possible to do

```
elem = executeScript("window.e = document.createElement('div'); return e")
getElementTagName(e)
```

So stale should just check that the element exists, not that it's connected to the document.

Issue #1090, opened 2017-09-11
Make clear that window.print user prompt 


whimboo:
Via https://github.com/mozilla/geckodriver/issues/925 we got a report that geckodriver doesn't work for `window.print` dialogs. And that is expected because we cannot handle this OS-wide dialog. Interestingly our spec mentions this as a valid user prompt. I think we should get this clarified or removed.

https://w3c.github.io/webdriver/webdriver-spec.html#user-prompts

andreastt:
I can’t find a reference to `window.print` as a “valid user prompt”?

whimboo:
```
The current user prompt is said to be the active user prompt, which can be one of the entries on the table of simple dialogs, or the window.print dialog.
```

andreastt:
Sorry, I see the example to `window.print` further down. We should remove that.

PR #1091, opened 2017-09-11
user prompts: drop mention of window.print

andreastt:
window.print creates an OS level dialogue that no drivers should be
allowed to interact with.

Fixes: https://github.com/w3c/webdriver/issues/1090

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1091)


andreastt:
Technically speaking the specification is correct because the print dialogue isn’t handled, but I can see how the mention of is as a current user prompt is confusing.

PR #1109, opened 2017-09-22
CODEOWNERS: add

andreastt:
Suggest reviewers from the WebDriver admin group.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1109)

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,1 @@
+*	@w3c/webdriver


andreastt:
So _I think_ this will automatically suggest reviewers when you submit a PR, so you don’t have to search for them in the top-right box of the PR page.

PR #1114, opened 2017-09-22
back: make description non-normative

andreastt:
The patch makes the first paragraph describing the semantics of
the command non-normative, because it could easily be mistaken for
a normative requirement.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1114)


andreastt:
Dropped the rename.

PR #1115, opened 2017-09-22
forward: make intro non-normative

andreastt:
The first paragraph is made into a note as to not confuse readers
about its normative requirements.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1115)


andreastt:
Dropped the rename.

PR #1118, opened 2017-09-22
command contexts: make intros non-normative

andreastt:
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1118)


shs96c:
There's some inconsistency about whether or not "command" links to the definition or not. We can fix that up in a later diff.

PR #1123, opened 2017-10-06
'Get Window Handles' returns window handles ordered by timestamp of creation

JonasRudolph:
I suggest, that the Get Window handles command returns the window handles in ascending order, sorted by the timestamp of the creation of their associated top-level browsing context.
By doing this, testers can rely on the fact, that the first window handle in the list represents the window, with which the test was started and the last window handle in the list represents the most recently opened window.

There also is a StackOverflow question [[0]] with ~200k views from which the accepted answer suggests to just switch to the last window handle when one wants to switch to the most recently opened browsing context. While researching I also found other humans (examples: [[1]][[2]]) wrongly asserting that behaviour of the command.

I think, this improvement would be really helpful .

What do you think?

[0]: https://stackoverflow.com/a/9597714/4099380
[1]: https://sqa.stackexchange.com/a/4993
[2]: https://stackoverflow.com/a/19116017/4099380

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -3698,7 +3698,8 @@ with a "<code>moz:</code>" prefix:
 <p>The <dfn>Get Window Handles</dfn> <a>command</a>
  returns a list of <a>window handles</a>
  for every open <a>top-level browsing context</a>.
- The order in which the window handles are returned is arbitrary.
+ The window handles must be returned in ascending order, sorted by the 
+ timestamp of the creation of their associated <a>top-level browsing context</a>.
 
 <p>The <a>remote end steps</a> are:
 


sideshowbarker:
Marked as non substantive for IPR from ash-nazg.

andreastt:
The last time we discussed this, we landed on explicitly calling out arbitrary ordering of window handles because it is difficult to achieve what you are describing in some user agents. The reality of this may have changed, however, so I think it could be worth revisiting this discussion for level 2.

Certainly I would love for there to be some determinism about how window handles are ordered.

jgraham:
Per the testing policy, this PR can't be accepted without a corresponding test in the web-platform-tests repository.

andreastt:
@JonasRudolph Do you have time to follow up on @jgraham’s note? I’d be happy to review tests to WPT.

shs96c:
I'm not sure whether this will provide a stable ordering in all cases (for example, start a timer that opens a window once it expires, then have a loop that opens windows --- what's the expected stable ordering), and requires the browser to track when the window was opened.

Admittedly, the example given here is a pathological case.

PR #1127, opened 2017-10-11
add definition of 'script' variable

jlipps:
In the definition for [extracting the script arguments from a request](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-extract-the-script-arguments-from-a-request) for Execute Script, we refer to a `script` variable, but it has not been defined. Presumably it is meant to match a `script`  property in the parameters?

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1127)


andreastt:
@AutomatedTester This is also a bug in level 1.

PR #1139, opened 2017-11-02
Change dispatch pointerUp's x, y handling

JKereliuk:
x and y should come from the input state and not the action object since passing in x and y values in the payload of a pointerUp doesn't really make sense :)

cc/ @jgraham 

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1139)


JKereliuk:
Friendly ping for review @andreastt @shs96c :)

PR #1144, opened 2017-11-10
Remove the ability to return a window or frame from execute script

shs96c:
This has not been implemented in any webdriver implementation
and so is being lifted to Level 2.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1144)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -3537,32 +3537,6 @@ with a "<code>moz:</code>" prefix:
  identifies it. This must be a <a>String</a> and must not be
  "<code>current</code>".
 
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -3540,29 +3540,6 @@ <h2 data-dfn-for="NavigatorAutomationInformation">Interface</h2>
 <p>The <dfn>web window identifier</dfn>
```
jgraham: This can be deleted too.

shs96c: Done.

PR #1145, opened 2017-11-10
Remove the configurability of the user prompt behaviour

shs96c:
This is not implemented in two separate implementations of
webdriver. Punting to level 2, but specifying unconfigurable
default behaviour for handling user prompts.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1145)


shs96c:
This is based on feedback from @AutomatedTester that this behaviour (dimiss and error) will work, and from @jimevans that this behaviour is implemented in the IEDriverServer. This brings us into line with Selenium's existing behaviour.

PR #1148, opened 2017-11-13
actions: input state is a variable, not a definition

andreastt:
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1148)


andreastt:
Fixes two warnings from ReSpec.

Issue #1147, opened 2017-11-13
This is redundant with delete all cookies section immediately below.


DrMarcII:
In chapter [16. Cookies](https://w3c.github.io/webdriver/webdriver-spec.html#add-cookie#x16-cookies), section [16.4 Delete Cookie](https://w3c.github.io/webdriver/webdriver-spec.html#add-cookie#x16-4-delete-cookie):
> or all the cookies associated with the active document’s address if name is undefined.

PR #1149, opened 2017-11-14
cookies: remove non-normative introductions

andreastt:
Drops the non-normative introductions that can be confusing to
implementors from the Delete Cookie and Delete All Cookies commands.

Fixes: https://github.com/w3c/webdriver/issues/1147

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1149)


andreastt:
Ping @AutomatedTester

Issue #1129, opened 2017-10-12
Element Clear isn't clear on what requirements the element must provide


InstyleVII:
In the summary for Element Clear we state (emphasis mine):
>The Element Clear command scrolls into view **an editable or resettable element** and then attempts to clear its selected files or text content.

Then later on in step 3 we state (emphasis mine):
>If element is **neither content editable nor both editable and resettable** return an error with error code invalid element state.

This is confusing as content editable is a subset of editable per the definition of editable. It also makes later sections confusing as we don't reference editable at all:

>When required to clear a **content editable** element a remote end must run the following steps: 

and

>When required to clear a **resettable element** a remote end must run the following steps:

andreastt:
The first paragraph is meant to be informative, but this has turned out to be misleading, see e.g. https://github.com/w3c/webdriver/issues/1105. I’ve submitted a number of PRs to fix this but they do not appear to have been merged yet. Maybe @AutomatedTester can give us an update on when they will be.

On the topic of the _Element Clear_ command, I tried implementing it a few months ago and got stuck on the same issues you mention here. See my comment back then in https://bugzilla.mozilla.org/show_bug.cgi?id=1354211#c1.

I think we need to revisit this command to make it implementable. I haven’t got time to do that right now, but I leave this here as a note that I’ve observed the same problems as @InstyleVII.

barancev:
One extra note: the main step to clear a resettable element is:
```
Run element's reset algorithm.
```
This action does not "clear" element, it does not always set "empty" value. For example, input elements have this reset algorythm:
```
The reset algorithm for input elements is to set the dirty value flag and dirty checkedness flag back to false, set the value of the element to the value of the value content attribute, if there is one, or the empty string otherwise
```

andreastt:
@shs96c and I had a conversation on IRC about this, and our conclusion was that Element Clear itself is reasonable, with the following exceptions:

- The non-normative introduction text needs to be yanked
- The reset algorithm from HTML resets the value to the initial value, and we need to adapt it to always set an empty string
- There are possibly some missing DOM events (change?)
- Keyboard-interactable, which is a substep of telling if the element is interactable, is hard to implement but HTML has improved since I last looked at it

andreastt:
[...]

PR #1152, opened 2017-11-22
element clear: drop intro

andreastt:
Drop the non-normative introductions that can be confusing to
implementors from the Element Clear command.

Fixes: https://github.com/w3c/webdriver/issues/1147
Fixes: https://github.com/w3c/webdriver/issues/1129

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1152)


andreastt:
Actually, let me put all the patches for fixing Element Clear into a single PR.

PR #1156, opened 2017-11-24
element clear: clear instead of resetting elements

andreastt:
The Element Clear command currently _resets_ interactable elements to their default value, e.g. what was defined in the `value` IDL attribute in the case of `<input>`. The intention of this command is to _clear_ the value entirely. These patches fixes the problem, but unfortunately had to reorganise dependencies a bit to make it happen.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1156)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -306,6 +306,8 @@ in the spec, as demonstrated in a (yet to be developed)
    <!-- Color state --> <li><dfn><a href="https://html.spec.whatwg.org/#color-state-(type=color)"><code>color</code> state</a></dfn>
    <!-- Cookie-averse Document object --> <li><dfn><a href=https://html.spec.whatwg.org/#cookie-averse-document-object>Cookie-averse <code>Document</code> object</a></dfn>
    <!-- Current entry --> <li><dfn><a href=https://html.spec.whatwg.org/#current-entry>Current entry</a></dfn>
+   <!-- Dirty checkedness flag --> <li><dfn><a href=https://html.spec.whatwg.org/#concept-input-checked-dirty-flag>Dirty checkedness flag</a></dfn>
+   <!-- Dirty value flag --> <li><dfn><a href=https://html.spec.whatwg.org/#concept-fe-dirty>Dirty value flag</a></dfn>
    <!-- Disabled --> <li><dfn><a href=https://html.spec.whatwg.org/#concept-fe-disabled>Disabled</a></dfn>
    <!-- Document address --> <li><dfn data-lt=address><a href=https://dom.spec.whatwg.org/#concept-document-url>Document address</a></dfn>
    <!-- Document readiness --> <li><dfn><a href=https://html.spec.whatwg.org/#current-document-readiness>Document readiness</a></dfn>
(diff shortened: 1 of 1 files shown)


andreastt:
FWIW this doesn’t require any tests because the tests for this are correct.

webdriver-spec.html
```diff
@@ -5512,6 +5522,33 @@ <h2>Element Interaction</h2>
  they will implicitly <a data-lt="scroll into view">scroll elements into view</a>
  and check that it is an <a>interactable element</a>.
 
+<p>Some <a>resettable elements</a> define their own <dfn>clear algorithm</dfn>.
+ Unlike their associated <a>reset algorithms</a>,
+ changes made to form controls as part of these algorithms
+ <em>does</em> count as changes caused by the user
```
shs96c: Nit: s/does/do/

webdriver-spec.html
```diff
@@ -5512,6 +5522,33 @@ <h2>Element Interaction</h2>
...
+ (and thus, e.g. does cause <a><code>input</code></a> events to fire).
```
shs96c: And here.

webdriver-spec.html
```diff
@@ -5512,6 +5522,33 @@ <h2>Element Interaction</h2>
...
+ When the <a>clear algorithm</a> is invoked
+ for an element that does not define its own <a>clear algorithm</a>,
+ its <a>reset algorithm</a> must be invoked instead.
+
+<p>The <a>clear algorithm</a> for <a><code>input</code></a> elements
```
shs96c: This may be clearer as a switch. When I did the initial read, I though all cases applied in all cases. Maybe I need more coffee?

webdriver-spec.html
```diff
@@ -5512,6 +5522,33 @@ <h2>Element Interaction</h2>
...
+ is to set the <a>dirty value flag</a> and <a>dirty checkedness flag</a> back to false,
+ set the <a>value</a> of the element to an empty string,
```
shs96c: What if the empty value doesn't pass the correctness tests for the new value?

webdriver-spec.html
andreastt: Fixed.

webdriver-spec.html
andreastt: Fixed.

webdriver-spec.html
andreastt: I did first write this as a switch, but switches make it hard for other specifications to reference WebDriver. The other reason I chose this approach is that it is the same that HTML uses for ‘resettable elements’ and ‘reset algorithm’, which gives it a nice parity.

webdriver-spec.html
andreastt: An empty string should always be correct for `<input>` elements, seeing as `<input value="">` is always allowed. `<input required>` may not allow the form to be _submitted_, but that’s unrelated.

andreastt:
Ping @AutomatedTester.

I believe the approach I’ve taken with this patch is the right one because it gives parity to the HTML spec and re-uses the same concepts. It would be easier for other specs to depend on WebDriver if element clear algorithms are defined this way if you consider we couldn’t have extended HTML this way if it was using switches.

PR #1157, opened 2017-11-24
get active element: return error if activeElement is null

andreastt:
It is not currently clear what WebDriver should do when
document.activeElement returns null, which may happen if the document
element is deleted.  This makes it clear that it will return a no
such element error in such a case.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1157)


andreastt:
This PR requires test changes as well.

andreastt:
The test will be updated in https://bugzilla.mozilla.org/show_bug.cgi?id=1420431.

andreastt:
Fixed the test.

andreastt:
Ping @AutomatedTester

PR #1162, opened 2017-11-29
Refer to Window.outerWidth/outerHeight normatively

gsnedders:
Contrary to the note, they aren't proprietary, and have been in CSSOM View all the way back to its first draft a decade ago (see https://github.com/w3c/csswg-drafts/commit/fdd7f32848f891251bffd7134c004f783bf6e970).

Also add a note to the Set Window Rect resizing.

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1162)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -4082,6 +4082,14 @@ with a "<code>moz:</code>" prefix:
 
     <p>Particular implementations may have other limitations
      such as not being able to resize in single-pixel increments.
+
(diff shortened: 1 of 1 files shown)


gsnedders:
Ideally for the setter side of this, we could just refer to CSSOM View's "client window", but that hasn't actually been defined. (That's https://github.com/w3c/csswg-drafts/issues/430.)

PR #1167, opened 2017-11-30
acknowledgements: add gsnedders

andreastt:
@gsnedders

---
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1167)

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -9416,6 +9416,7 @@ argument <var>value</var>:
  <!-- Eran Messeri --> Eran Messeri,
  <!-- Erik Wilde --> Erik Wilde,
  <!-- Gábor Csárdi --> Gábor Csárdi,
+ <!-- Geoffrey Sneddon --> Geoffrey Sneddon,
  <!-- James Graham --> James Graham,
  <!-- Jason Juang --> Jason Juang,
  <!-- Jason Leyba --> Jason Leyba,


webdriver-spec.html
```diff
@@ -9401,6 +9401,7 @@ <h2>Acknowledgements</h2>
  <!-- David Burns --> <a href=http://www.theautomatedtester.co.uk/>David Burns</a>,
  <!-- Eran Messeri --> Eran Messeri,
  <!-- Erik Wilde --> Erik Wilde,
+ <!-- Geoffrey Sneddon --> Geoffrey Sneddon,
  <!-- Gábor Csárdi --> Gábor Csárdi,
```
gsnedders: Actually, shouldn't my name be after Gábor? What locale are we using for the sorting? Under both English and Hungarian rules, I'd expect Geoffrey and Gábor.

andreastt: I think the problem here is that sort(1) uses `LC_COLLATE=C`, which sorts byte by byte.

I don’t understand why this doesn’t work though:

    % LC_ALL=en_GB.UTF-8 sort -
    Geoffrey Sneddon
    Gábor Csárdi^D
    Geoffrey Sneddon
    Gábor Csárdi

andreastt: I have manually fixed this.

andreastt:
Ping @AutomatedTester

PR #1169, opened 2017-12-05
Define the correct rect for Take Screenshot

gsnedders:
@zcorpan is this the right way to do this with what we have defined in CSSOM View? (please wait for him or someone else who understands CSSOM View better to comment!)

***
<a href="https://pr-preview.s3.amazonaws.com/gsnedders/webdriver/pull/1169.html" title="Last updated on Dec 21, 2017, 2:36 PM GMT (a35157e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1169/c9a1ddc...gsnedders:a35157e.html" title="Last updated on Dec 21, 2017, 2:36 PM GMT (a35157e)">Diff</a>

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -9218,8 +9218,34 @@ argument <var>value</var>:
 
  <li><p>When the user agent is next to <a>run the animation frame callbacks</a>:
   <ol>
-   <li><p>Let <var>root rect</var> be the <a>current top-level browsing context</a>’s
-    <a>document element</a>’s <a>rectangle</a>.
+   <li><p>Let <var>window</var> be the <a>associated window</a>
+     of the <a>current browsing context</a>’s <a>active document</a>.
+
+   <li><p>Let <var>x</var> be <var>window</var>'s
+    <a><code>scrollX</code></a> attribute.
+
+   <li><p>Let <var>y</var> be <var>window</var>'s
+    <a><code>scrollY</code></a> attribute.
+
+   <li><p>Let <var>y</var> be <var>window</var>'s
+    <a><code>scrollY</code></a> attribute.
+
+   <li><p>Let <var>width</var> be the width of the <a>top-level
+    browsing context</a>'s <a>viewport</a> excluding the width of the
+    scroll bar, if any.
+
+   <li><p>Let <var>height</var> be the height of the <a>top-level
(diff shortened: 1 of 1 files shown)


zcorpan:
Do I understand correctly that you want to get the rectangle of the top-level viewport?

What you have is basically scrollX, scrollY, innerWidth, innerHeight, yes?

Looking at https://w3c.github.io/webdriver/webdriver-spec.html#screen-capture it's not super-clear to me if the coordinate space it's expecting is the same as cssom-view uses for scrollX et al. Note "relative to the initial containing block origin" -- IIRC that is intended to match up with coordinate space for absolute positioning. This becomes interesting with `dir=rtl` (or vertical writing modes), where scrolling left from default scroll position makes `scrollX` negative.

Here's a test to play around with http://software.hixie.ch/utilities/js/live-dom-viewer/saved/5607

Also see https://lists.w3.org/Archives/Public/www-style/2014Apr/0438.html for information on interop problems to watch out for.

So... what to do. Figure out what coordinate space the screenshotter is expecting, then figure out how to get that rectangle, and test with rtl/vertical.

Hope that helps :)

gsnedders:
> What you have is basically scrollX, scrollY, innerWidth, innerHeight, yes?

innerWidth/Height include the scrollbars, which we don't want to.

andreastt:
Closing and reopening to trigger pr-preview.

webdriver-spec.html
```diff
@@ -9218,8 +9218,31 @@ <h3>Take Screenshot</h3>
 
  <li><p>When the user agent is next to <a>run the animation frame callbacks</a>:
   <ol>
-   <li><p>Let <var>root rect</var> be the <a>current top-level browsing context</a>’s
-    <a>document element</a>’s <a>rectangle</a>.
+   <li><p>Let <var>window</var> be the <a>associated window</a>
+     of the <a>current browsing context</a>’s <a>active document</a>.
```
shs96c: The original wording of this captured the visible contents of the OS window. Would `current top level browser context` be more in keeping with the spirit of this?

gsnedders:
Closing this in favour of just an issue (#1251), because I don't really know what the right thing here is, nor have the time to better understand it.

Issue #1142, opened 2017-11-08
Improve wording around extension commands


andreastt:
In chapter [6. Protocol](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-extension-command#x6-protocol), section [6.7 Protocol Extensions](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-extension-command#x6-7-protocol-extensions):
> The protocol is designed to allow extension to meet vendor-specific needs. Commands that are specific to a user agent are called extension commands and behave no differently than other commands; each has a dedicated HTTP endpoint and a set of remote end steps. 

Extension commands are also useful in providing other standards with the ability to extend WebDriver. Extension commands for vendor-specific needs is only one aspect of it, and we should improve the wording around this to not make it sound exclusionary. The current wording is confusing to other specification authors as can be seen in https://github.com/w3c/permissions/pull/151/files?diff=unified#r123109981.

We use a better explanation under the _Design notes_ chapter:

> WebDriver aims to allow other standards to provide extensions to support new functionality, and to allow conformance tests that cannot be implemented entirely in ECMAScript to be written in a vendor-neutral way.

Issue #1170, opened 2017-12-06
Specifying extension commands with no "name"


jugglinmike:
In [our efforts to integrate the WebDriver in the Permissions specification](https://github.com/w3c/permissions/pull/151), we've identified a need that isn't addressed by the "protocol extension" mechanism today: defining a command that has no associated name.

In this case, we're looking to support setting a Permission in automation. We'd like to define a command for the HTTP POST method to the URI template `/session/{session id}/permissions`.

The definition of the "extension command prefix" defines its intent as: "separating extension commands from other commands in order to avoid potential resource conflicts with other implementations." Based on this definition, `session/{session id}/permissions` seems like the correct choice, but in this case, there is no particular "name" we'd like to assign.

We could circumvent this by defining the "prefix" as `session/{session id}` and the "name" as `permissions`, but this conflicts with the stated intent of those segments.

I'd like to propose that the "extension command name" be made optional. Beyond documenting it as such, I believe the only other change to the WebDriver specification text would be in the instruction for how the "extension command URI Template" is constructed.

Would this be acceptable to the editors?

jugglinmike:
We resolved via gh-1177 :smile_cat:

PR #1177, opened 2017-12-22
Generalize protocol extension mechanism

jugglinmike:
The facilities offered for protocol extensions are intended to be used
by both web standards and browser vendors. Prior to the application of
this commit, the specification text inconsistently described these two
distinct use cases [1]. The mechanism itself was biased towards the
"browser vendor" use case because it required the specification of an
"extension command prefix" which is not appropriate for all "web
standards" use cases [2].

Explicitly describe both use cases in the informative text and
generalize the normative text to better-support extensions by web
standards.

[1] https://github.com/w3c/webdriver/issues/1142
[2] https://github.com/w3c/webdriver/issues/1170

---

This is an attempt to resolve gh-1142 and gh-1170 as [recently discussed on IRC](https://w3.logbot.info/webdriver/20171207#c366572-c366621).

[The Permissions specification](https://w3c.github.io/permissions/) was [recently patched](https://github.com/w3c/permissions/pull/151) to define an extension command according to the specification text today. As described in gh-1170, this meant defining a resource of the form `session/{session id}/permissions/set`. If this patch is accepted, I plan to update the template to `session/{session id}/permissions`.

***
<a href="https://pr-preview.s3.amazonaws.com/bocoup/webdriver/pull/1177.html" title="Last updated on Jan 6, 2018, 2:48 AM GMT (a6ff1fb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1177/d50bdf0...bocoup:a6ff1fb.html" title="Last updated on Jan 6, 2018, 2:48 AM GMT (a6ff1fb)">Diff</a>

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -1812,7 +1812,7 @@ in the spec, as demonstrated in a (yet to be developed)
  as that of other built-in <a>commands</a>.
 
 <p>In order to avoid potential resource conflicts with other implementations,
- vendor-specific <a>extension command URI Template</a>s SHOULD begin with one
+ vendor-specific <a data-lt="extension command URI Template">extension command URI Templates</a> must begin with one
  or more path segments which uniquely identifies the vendor and UA.
  It is suggested that vendors use their vendor prefixes
  without additional characters as outlined in [[CSS21]],
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -671,11 +671,11 @@ in the spec, as demonstrated in a (yet to be developed)
 <section>
 <h3>Extensions</h3>
 
-<p>WebDriver aims to allow other standards
- to <a data-lt="extension commands">provide extensions</a> to support new functionality,
- and to allow conformance tests that cannot be implemented
- entirely in <a href=https://tc39.github.io/ecma262/>ECMAScript</a>
- to be written in a vendor-neutral way.
+<p>WebDriver provides a mechanism for others to define extensions to the protocol
+ for the purposes of automating functionality that cannot be implemented entirely
+ in <a href=https://tc39.github.io/ecma262/>ECMAScript</a>. This allows other
+ web standards to support the automation of new platform features. It also
+ allows vendors to expose functionality that is specific to their browser.
 </section>
 </section> <!-- /Design Notes -->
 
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -1790,29 +1790,36 @@ <h3>Handling Errors</h3>
...
- and the <a>extension command</a>’s <a data-lt="extension command name">name</a>.
- The <a>extension command URI Template</a>,
+ This value,
  along with the HTTP method and <a>extension command</a>,
  is added to the <a>table of endpoints</a>
  and thus follows the same rules for <a>request routing</a>
  as that of other built-in <a>commands</a>.
 
+<p>In order to avoid potential resource conflicts with other implementations,
+ vendor-specific <a>extension command URI Template</a>s SHOULD begin with one
```
andreastt: You can add `<dfn data-lt="extension command URI Templates">…</dfn>` for the plural form on :1805.

andreastt: I believe the vendor namespace was a requirement before and we shouldn’t loosen this by saying vendors _should_ be using a vendor prefix.

I suggest s/SHOULD/must/ here or just rephrasing it to have a normative sound to it:

> … vendor-specific _extension command URI templates_ beginw tih on or more path segments …

webdriver-spec.html
```diff
@@ -1790,29 +1790,36 @@ <h3>Handling Errors</h3>
...
+ or more path segments which uniquely identifies the vendor and UA.
+ It is suggested that vendors use their vendor prefixes
+ without additional characters as outlined in [[CSS21]],
```
andreastt: I filed https://github.com/w3c/webdriver/issues/1179 about defining what ‘additional characters’ mean.

andreastt:
@jugglinmike Do you have time to address the nits?

webdriver-spec.html
jugglinmike: It looks like the convention is to prefer the singular for term definitions, so I've done the inverse:

    <a data-lt="extension command URI Template">extension command URI Templates</a>

Does that seem okay to you?

jugglinmike: Sure thing!

webdriver-spec.html
jugglinmike: Ah, very good

jugglinmike:
Thanks for the review, @andreastt! I've pushed up a commit to implement your suggestions--sorry for the delay.

andreastt:
Thanks! It looks great!

jugglinmike:
Agreed :) Thanks again for the help--I'll be following up with a patch for the Permissions specification soon

PR #1182, opened 2018-01-08
Removing tests that are not included in Level 1

AutomatedTester:
this removes results from tests that are not part of level 1 from the reports


andreastt:
I’m not sure we agreed Element Clear was out of scope of level 1, but I do see they are failing in all UAs currently.

PR #1185, opened 2018-01-08
Add clarification about why we wait in `click`

### LINKED ISSUE #1046: For click it's not clear how long to wait for a navigation to happen
The current spec says the following for `click`:

> 9. Perform implementation-defined steps to allow any navigations triggered by the click to start.
> 10. Try to wait for navigation to complete. 

We can perform those implementation-defined steps but it's not clear how long it will take until the navigation starts. It can be from some milliseconds up to more then 100ms. It all depends on the CPU load, performance of the machine, and type of build of the browser (debug builds are generally slower). 

Also not each `click` triggers a navigation, which would mean that we loose a given number of milliseconds each time a `click` is not causing a navigation.

How can the spec cover that, or don't we want to do that? 

For Marionette I implemented that we are waiting for `beforeunload` events which are send first. So a hard-coded 200ms delay works just fine in all cases. But even then a navigation request can be prevented by a custom listener registered for `beforeunload`, or something else delays it.

Given that navigation is optional for `click` how can we define the duration the client has to wait? Also to not affect those `click` commands that much, which don't cause a navigation.

shs96c:
Closes #1046.

***
<a href="https://pr-preview.s3.amazonaws.com/shs96c/webdriver-1/pull/1185.html" title="Last updated on Jan 8, 2018, 2:48 PM GMT (741e829)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1185/69b5bad...shs96c:741e829.html" title="Last updated on Jan 8, 2018, 2:48 PM GMT (741e829)">Diff</a>

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -5712,7 +5712,10 @@ argument <var>reference</var>, run the following steps:
   algorithm to wait, and it is acknowledged that some implementations
   may have unavoidable race conditions. The intention is to allow a
   new attempt to <a>navigate</a> to begin so that the next step in the
-  algorithm is meaningful.
+  algorithm is meaningful. It is possible the click does not cause an
+  attempt to <a>navigate</a>, in which case the implementation-defined
+  steps can return immediately, and the next step will also return
+  immediately.
 
  <li><p><a>Try</a> to <a>wait for navigation to complete</a>.
 
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -5708,6 +5708,12 @@ argument <var>reference</var>, run the following steps:
   any <a data-lt="navigation">navigations</a> triggered by the click
   to start.
 
+  <p class="note">It is not always clear how long this will cause the
+  algorithm to wait, and it is acknowledged that some implementations
+  may have unavoidable race conditions. The intention is to allow a
+  new attempt to <a>navigate</a> to begin so that the next step in the
+  algorithm is meaningful.
+
  <li><p><a>Try</a> to <a>wait for navigation to complete</a>.
 
  <li><p><a>Try</a> to run the <a>post-navigation checks</a>.


webdriver-spec.html
```diff
@@ -5708,6 +5708,12 @@ <h3>Element Click</h3>
   any <a data-lt="navigation">navigations</a> triggered by the click
   to start.
 
+  <p class="note">It is not always clear how long this will cause the
+  algorithm to wait, and it is acknowledged that some implementations
+  may have unavoidable race conditions. The intention is to allow a
+  new attempt to <a>navigate</a> to begin so that the next step in the
+  algorithm is meaningful.
```
andreastt: I think this misses the point @whimboo raised that not all clicks induce navigation. For example clicking an anchor won’t cause navigation, but the point of this step is to either allow enough time for a navigation attempt to begin in cases you can’t tell up front or for steps to be taken in cases where you can.

So the point here is not that this causes “unavoidable race conditions”, but that there is a need to run some steps to tell whether the click is going to cause or have caused navigation.

shs96c: Surely that's handled by the "any navigations" language? I'll amend note to indicate that not all clicks cause navigation.

andreastt: Thanks, that makes it clearer.

But is it really true that “implementations may have unavoidable race conditions” with running the Go command?

andreastt:
It would be good if @whimboo looked at whether this addresses his concerns in https://github.com/w3c/webdriver/issues/1046.

whimboo:
The changes here are very generic. I wonder how different browser vendors will implement it based on this wording. I would assume that we will see different behavior in handling that. And some might even fail in some conditions others don't. 

So we don't want to set a maximum amount of time the algorithm has to wait before returning from the click command? I feel it will be hard to explain to users why one is waiting longer and is able to catch a specific navigation request, while another browser doesn't for the same situation.

andreastt:
@shs96c talked a little bit about different implementation techniques in https://github.com/w3c/webdriver/issues/1046#issuecomment-324958842.

PR #1186, opened 2018-01-08
Update HTML attributes to use quotes

InstyleVII:
Added quotes to all HTML attributes to improve editor tooling.

***
<a href="https://pr-preview.s3.amazonaws.com/InstyleVII/webdriver/pull/1186.html" title="Last updated on Jan 8, 2018, 8:48 PM GMT (39de95d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1186/e9d00b7...InstyleVII:39de95d.html" title="Last updated on Jan 8, 2018, 8:48 PM GMT (39de95d)">Diff</a>

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -5,7 +5,7 @@
 <meta charset=utf-8>
 <title>WebDriver</title>
 
-<script src=https://www.w3.org/Tools/respec/respec-w3c-common async class=remove></script>
+<script src="https://www.w3.org/Tools/respec/respec-w3c-common" async class=remove></script>
 <script class=remove>
 var respecConfig = {
   specStatus: "CR",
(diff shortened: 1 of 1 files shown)


InstyleVII:
This is an updated version of my previous PR with just the addition of quotes to all HTML attributes as well as one minor fix to an incorrect class name.

andreastt:
I don’t see the value of this PR at all. This is not incorrect usage of attributes.

InstyleVII:
Two mains reasons:
Editor tooling - Some editors won't read links as links allowing you to nav to them from the editor if they aren't in quotes
Consistency - We had some attributes using quotes and others not using quotes, now the document is consistent

andreastt:
The document is pretty much consistent in not using quotes where it is unnecessary already, so I wouldn’t go as far as calling that an inconsistency. I would also consider it a bug that your editor doesn’t recognise valid HTML attributes.

Now I’m not going to block a patch that adds quotes everywhere, but I would like to reiterate that they are completely pointless.

InstyleVII:
@shs96c mind assigning a reviewer?

InstyleVII:
Going to close as consensus is this isn't needed.

Issue #1187, opened 2018-01-09
Element Click allows <option disabled> to be selected


andreastt:
It was reported to us in https://github.com/mozilla/geckodriver/issues/1107 that the [Element Click](https://w3c.github.io/webdriver/webdriver-spec.html#dfn-element-click) command lets you select `<option disabled>`.

It’s not clear to me what the expected behaviour is. Since we don’t check disabled/readonly for any other elements when clicking, I’m guessing WebDriver should return silently and not change the element’s selectedness?

barancev:
A user can't select a disabled option, neither with the mouse nor with the keyboard. So I suppose Selenium should not be able to select it too. Yes, a user can click a disabled option, but this action does not change the selected element, it's effectively a no-op.

andreastt:
Thanks for the clarification on behaviour. Just checking it is not meant to return an error when trying to click a disabled element.

whimboo:
Please note that there is a difference between a select with and without the multiple flag. For multiple selects a disabled option can receive focus, while for the others it doesn't work. Here an example jsbin: https://jsbin.com/zebisixudu/1

barancev:
@whimboo In this sample the disabled option can take focus but it still can't be selected.

(By the way, in Chrome and IE it can't take focus even in multi-select, it's Firefox specific behavior)

whimboo:
Interesting. I filed a [Firefox bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1429350) for that.

PR #1189, opened 2018-01-10
element click: prevent selectedness change on <option disabled>

andreastt:
WebDriver currently toggles the selectedness state of <option
disabled> elements.  Because we can interact with the containing
<select> element we should simulate the mouse interaction steps but
prevent it from changing the selectedness state.

Fixes: https://github.com/w3c/webdriver/issues/1187

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -5617,23 +5617,26 @@ argument <var>reference</var>, run the following steps:
 
        <li><p>Run the <a>focusing steps</a> on <var>parent node</var>.
 
-       <li><p><a>Fire</a> an <a><code>input</code></a> event at <var>parent node</var>.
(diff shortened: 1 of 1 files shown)


andreastt:
@barancev This changes it to simulate the mouse interaction steps on `<select>`, but prevents `<option disabled>`’s selectedness from changing. Does this match the behaviour you were expecting?

andreastt:
I’ve written tests in https://bugzilla.mozilla.org/show_bug.cgi?id=1429384.

andreastt:
@AutomatedTester Any chance you can have a look at this soon? The implementation changes have landed in m-c, and the tests in m-c and upstream in WPT.

I would like this merged before I work on https://github.com/w3c/webdriver/issues/1190.

PR #1192, opened 2018-01-15
Update results

AutomatedTester:
Waiting on updated results from MS


carlosgcampos:
I've just released WebKitGTK+ 2.18.6, updated results are [here](https://people.igalia.com/cgarcia/wk-2.18.6.json)

Issue #1203, opened 2018-01-23
Clarify how firstMatch capabilities are validated against alwaysMatch


carlosgcampos:
According to the spec, once capabilities have been validated, we iterate firstMatch list, trying to merge and match the capabilities. As soon as we have a match we return the merged capabilities. This means that there might be duplicated entries in first and always as long as any previous one matches. This is what happens in test merge.py::test_merge_invalid, that includes an empty object as the first item of every firstMatch list. The test passes in firefox, because it first checks that there isn't duplicated entries in first and always and then it does the merge and match. If that was the idea, then we should fix the spec, probably adding a step before the iteration to do the check, and remove the check from the merge algorithm. Otherwise we should fix the test by removing the initial empty object from the firstMatch list.

PR #1204, opened 2018-01-23
Try to merge capabilities before selecting them.

jgraham:
This means that the session will fail if any capabilities can't be
merged, rather than failing only if the capabilities that can't be merged
are before the ones that are selected. The early failure is consistent
with the early failure for other kinds of validation errors elsewhere.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/pull/1204.html" title="Last updated on Jan 23, 2018, 2:00 PM GMT (fa4d3d1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1204/877030f...fa4d3d1.html" title="Last updated on Jan 23, 2018, 2:00 PM GMT (fa4d3d1)">Diff</a>

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -2236,17 +2236,25 @@ with a "<code>moz:</code>" prefix:
     <var>validated first match capabilities</var>.
   </ol>
 
+  <li><p>Let <var>merged capabilites</var> be an empty <a>List</a>.
+
   <li><p>For each <var>first match capabilities</var> corresponding
    to an indexed property in <var>validated first match capabilities</var>:
    <ol>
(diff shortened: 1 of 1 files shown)


jgraham:
Fixes #1203

carlosgcampos:
This looks good to me. From the implementation point of view I think it would be simpler to check the first match capabilities while validating them (see [patch](https://people.igalia.com/cgarcia/wd-pr1204.diff)). That's what I've done in WebKit, see bug [180407](https://bugs.webkit.org/show_bug.cgi?id=180407). From the spec point of view either way is fine, anyway.

PR #1211, opened 2018-01-26
Add a command to minimize the window

rude9:
# # 

> [`**It**`](

- 

- [ ] 

- 

1. url
Fixes @darobin 

)

 is likely that users are going to want to minimize a window during
their testing. Adding that here.


andreastt:
This is stale and we specified Minimize Window elsewhere.

Issue #1153, opened 2017-11-23
Clarify the expected behavior when an alert is shown in the middle of a script execution


carlosgcampos:
The spec says:

> The rules to execute a function body are as follows. The algorithm will return success with the JSON representation of the function’s return value, or an error if the evaluation of the function results in a JavaScript exception being thrown or at any point during its execution an unhandled user prompt appears.

> If at any point during the algorithm a user prompt appears, the user prompt handler must be invoked. If its return value is an error, it must immediately return with that error and abort all subsequent substeps of this algorithm.

It explicitly says the subsequent substeps, so the current step that is the script execution itself should not be aborted according to the spec. So, for example:

<pre>
execute_script("window.alert(\"Hello World\");window.alert(\"Good bye\");"
switch_to_alert().accept()
</pre>

is showing the second alert in both firefox and chrome. If the expected behavior was to abort the script execution too then the second alert should not be shown.

andreastt:
This should be fixed by https://github.com/w3c/webdriver/pull/1248.

shs96c:
The spec doesn't say anything about what should happen to the JS engine in the browser, so this behaviour is currently undefined. The most inclusive way of implementing this is to say that the underlying script behaves as if the user had handled the prompt, which implies we'll need to handle both of the alerts.

whimboo:
During TPAC we discussed this issue. Here an overview: https://www.w3.org/2018/10/26-webdriver-irc#T14-35-50

Most of the drivers (geckodriver, chromedriver, safaridriver, and probably Edge) support immediately aborting the currently running script if a user prompt appears, and as such seem to currently fail running the `webdriver` test because it was written for the IE driver:

https://wpt.fyi/results/webdriver/tests/execute_script/execute.py?label=experimental

As @shs96c already mentioned above we should make it more clear in the spec that the script has to be aborted immediately when a user prompt opens.

whimboo:
Also CC'ing @JohnChen0 who actually works on the chromedriver code. Maybe he can have a look at my last comment and run the check with chromedriver and Chrome? Thanks

JohnChen0:
The behavior of Chrome/ChromeDriver is similar to geckodriver/Marionette. Calling `window.alert` from JavaScript code causes `execute_script` to return immediately. Once `alert.accept()` closes the user prompt, the JavaScript code continues executing from the point of `window.alert` call.

andreastt:
[...]

Issue #1135, opened 2017-10-24
Make default unhandled prompt behavior crystal clear


p0deje:
In chapter [18. User Prompts](https://w3c.github.io/webdriver/webdriver-spec.html#user-prompts#x18-user-prompts):

> By default user prompts are not handled automatically unless a user prompt handler has been defined. When a user prompt appears, it is the task of the subsequent command to handle it. If the subsequent requested command is not one listed in this chapter, an unexpected alert open error will be returned.

This essentially means that by default remote end should not dismiss prompts. However, once reading spec more it becomes obvious that by default it has to dismiss prompt and return error. In the step 2 of the steps remote end should follow, the default behavior is:

> mssing value default state
> not in the table of simple dialogs
>   1. Dismiss the current user prompt.
>   2. Return an annotated unexpected alert open error.

So, essentially, the paragraph above should be changed to something like this:

> When a user prompt appears, it is the task of the subsequent command to handle it. If the subsequent requested command is not one listed in this chapter, a user prompt will be dismissed and an unexpected alert open error will be returned.

Also, a bit of logs from #selenium IRC:

```
[17:06:02]  simonstewart p0deje: Lemme pull up the spec
[17:06:32]  simonstewart That's the bit
[17:06:51]  simonstewart So, unless it's been set, the value for the current session's user prompt handler is "null"
[17:07:25]  simonstewart So we fall through to the bottom condition in step 2
[17:07:39]  simonstewart Where we
[17:07:44]  simonstewart Dismiss the current user prompt.
[17:07:49]  simonstewart Return an annotated unexpected alert open error.
[17:07:54]  simonstewart That seems pretty clear
[17:09:19]  p0deje ah
[17:09:33]  p0deje I understood this is done only when "not in the table of simple dialogs"
[17:09:59]  simonstewart "missing value, default state"
[17:10:03]  simonstewart too
```

PR #1248, opened 2018-04-04
Updating prompt handling during user script to be consistent

jimevans:
**NOTE** This PR is intended to spark discussion. It touches on at least two issues already filed against the spec (#1153 and #1135), and should be considered, since user prompt handlers was largely removed from the Level 1 specification.

Section 18 of the spec states that, "When a user prompt appears, it is
the task of the subsequent command to handle it." The algorithm for
"Executing a function body" (Section 15.2) violates this statement,
insisting that the user prompt handler be invoked during the current
command, rather than by the subsequent command. Instead, we should
immediately abort the execute[Async]Script command, and let the subsequent
command handle the user prompt, if any.

***
<a href="https://pr-preview.s3.amazonaws.com/jimevans/webdriver/pull/1248.html" title="Last updated on Apr 10, 2018, 2:59 PM GMT (3fa9de8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1248/582ae36...jimevans:3fa9de8.html" title="Last updated on Apr 10, 2018, 2:59 PM GMT (3fa9de8)">Diff</a>

### DIFF
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -1,4 +1,4 @@
-﻿<!doctype html>
+<!doctype html>
 <html
  data-issue-url="https://github.com/w3c/webdriver/"
  data-issue-param-milestone="Level 1">
--- a/webdriver-spec.html
+++ b/webdriver-spec.html
@@ -1,4 +1,4 @@
-<!doctype html>
+﻿<!doctype html>
 <html
  data-issue-url="https://github.com/w3c/webdriver/"
  data-issue-param-milestone="Level 1">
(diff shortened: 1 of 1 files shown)


webdriver-spec.html
```diff
@@ -1,4 +1,4 @@
-<!doctype html>
+﻿<!doctype html>
```
andreastt: I can’t tell from the raw patch what has actually changed here.

jimevans: It’s probably the UTF-8 byte marker from the editor I was using, since it’s at the very beginning of the file. I can fix that up if need be, I suppose.

andreastt: Yeah, we don’t want a BOM there.

jimevans: Removed in subsequent commit.

andreastt:
Spec change and tests look good to me, but it would be good if someone from another vendor and an editor would look at this before continue.

Issue #1245, opened 2018-03-29
Take Element Screenshot tries to use request params, but uses HTTP verb GET


burg:
In chapter [19. Screen Capture](https://w3c.github.io/webdriver/webdriver-spec.html#take-element-screenshot#x19-screen-capture), section [19.2 Take Element Screenshot](https://w3c.github.io/webdriver/webdriver-spec.html#take-element-screenshot#x19-2-take-element-screenshot):
> Let scroll be the result of getting the property scroll from parameters if it is not undefined. Otherwise let it be true.

However, in §6.3 Processing Model, Step 5, we only process parameters if the HTTP verb is POST. So this will always not work. I suggest changing the HTTP verb to POST, or drop the optional scrollIntoView and have clients do that some other way.

shs96c:
Good catch. My vote is for making it a POST. I guess the alternative is to allow query parameters to be used in a GET.

andreastt:
We can’t make a backwards-incompatible change to an endpoint now.

I believe it was Microsoft who wanted the optional `scrollIntoView` behaviour, but if that’s not used in practice, is it a controversial to propose dropping that instead?

Alternatively we need to add a _second_ POST endpoint for the same command which acts as an alias that can receive data. It is not an option to remove the GET endpoint.

burg:
Either default for scrollIntoView is acceptable to me.

thejohnjansen:
I think I'm being slightly misquoted. Microsoft wanted scrollIntoView to be OPTIONAL on Take Element Screenshot because I don't think Take Element Screenshot should do a scrollIntoView first. I think doing so will cause us to miss bugs. I think my example was for anchors on the page. If webdriver clicks an anchor that points to a div that is currently below the fold, the browser should scroll that div into view. If your browser (or your webpage) has a bug in it that prevents that div from scrolling into view, you'd miss that bug because WebDriver would scroll it into view for you. So, I'm OK with removing it completely and forcing webdevs to explicitly scrollIntoView before they take a screenshot if that's what they want.

andreastt:
@thejohnjansen The argument isn’t that anyone disagrees with the _reasoning_ behind it, but that providing data with a `GET` request isn’t possible.

InstyleVII:
In which case I'd imagine removing the body entirely and leaving the command as a GET would be fine.

shs96c:
It's entirely possible to have GET request by allowing query parameters. For a boolean parameter this seems like a reasonable thing to do..

[...]

PR #1249, opened 2018-04-06
take element screenshot: drop scroll feature

andreastt:
Issue #1245 points out that you cannot pass data parameters to a
GET method request, so it is currently impossible to pass a "scroll"
field to Take Element Screenshot.

It was also discovered in the discussion that no vendors implement
this feature.  Because we want the specification to match reality,
this patch removes the "scroll" field for Take Element Screenshot.

Fixes: https://github.com/w3c/webdriver/issues/1245

***
<a href="https://pr-preview.s3.amazonaws.com/andreastt/webdriver/pull/1249.html" title="Last updated on Apr 6, 2018, 3:40 PM GMT (81c89fa)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1249/582ae36...andreastt:81c89fa.html" title="Last updated on Apr 6, 2018, 3:40 PM GMT (81c89fa)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -9311,17 +9311,10 @@ sets the text field of a <a>window.<code>prompt</code></a>
 The <a>Take Element Screenshot</a> <a>command</a>
 takes a screenshot of the visible region encompassed
 by the <a>bounding rectangle</a> of an <a>element</a>.
-If given a parameter argument <code>scroll</code>
-that evaluates to false, the <a>element</a> will not be <a>scrolled into view</a>.
 
 <p>The <a>remote end steps</a> are:
 
 <ol>
- <li><p>Let <var>scroll</var> be the result
-  of <a>getting the property</a> <code>scroll</code> from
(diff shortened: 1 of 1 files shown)


jugglinmike:
This patch removes the un-implementable opt-in mechanism. That seems critical for folks trying to implement today. However, it also enables the feature unconditionally. As @shs96c [pointed out](https://github.com/w3c/webdriver/pull/1249#pullrequestreview-110311183), that's contentious.

@ato as a first step, do you think we could land this by removing all references to scrolling within this command? The discussion about the opt-in mechanism and/or changes to the default behavior can proceed from there.

carlosgcampos:
What's the status of this? I agree with removing the scroll parameter from the spec since it's impossible to use. I'm not sure I understand the expected behavior when not scrolling into view, would we end up capturing the viewport when the element is not visible? or should we generate an error because the element is not visible?

PR #1257, opened 2018-04-26
Clean-up

plehegar:
A few clean up for the spec.

Btw, can we move the spec to use index.html (to match other EDs)?


shs96c:
We could make the existing document be blank but with a meta-refresh in it
to point to the new location.

PR #1259, opened 2018-05-06
Trivial grammatical fix (missing preposition)

nbloomf:
Change "existing users Selenium WebDriver" to "existing users of Selenium WebDriver".


andreastt:
Thanks!

PR #1261, opened 2018-05-28
Remove superfluous 0 length check for files of element_send_keys

whimboo:
Step 10 of [Element Send Keys](https://w3c.github.io/webdriver/#dfn-element-send-keys) contains a 0-length check for files:

> 1. Let files be the result of splitting text on the newline (\n) character.
> 2. If files is of 0 length, return an error with error code invalid argument. 

This is superfluous because splitting a string for a separator which is not part of the string, or even if it is an empty string will always result in a single-entry array which contains the original string.

### DIFF
--- a/index.html
+++ b/index.html
@@ -6124,9 +6124,6 @@ must run the following steps:
      <li><p>Let <var>files</var> be the result of splitting <var>text</var>
       on the newline (<code>\n</code>) character.
 
-     <li><p>If <var>files</var> is of 0 length,
-      return an <a>error</a> with <a>error code</a> <a>invalid argument</a>.
-
      <li><p>Let <var>multiple</var> equal the result
       of calling <code><a>hasAttribute</a>("multiple")</code>
       on <var>element</var>.


whimboo:
Can someone please review? I will open a separate issue in regards of using `\n` as separator is not valid given that filenames can also contain such a character.

twalpole:
Splitting an empty string results in an empty Array, which is what this check is checking for.

whimboo:
No, it's always at minimum 1 element present, which is the part before the separator. And in case of an empty string it will be `[""]`.

twalpole:
@whimboo In Ruby, for instance,  `"".split("\n")   => []` so unless "splitting text" is defined somewhere in the spec that I'm missing I don't think your assertion is a valid assumption

whimboo:
In Javascript it's not: https://jsbin.com/gijemuqowe/edit?html,js,output

So it might depend on the language the driver is implemented in.

twalpole:
@whimboo Exactly, so that means the 0 length check may not be superfluous.

whimboo:
In that case lets close this PR. Thanks!

PR #1262, opened 2018-05-30
Element Send Keys for "<input type=file>" has to also send a change event.

whimboo:
As given by the HTML spec the `<input type=file>` element sends both the `input` and `change` event. The spec currently only requires to fire the `input` event.

https://html.spec.whatwg.org/#file-upload-state-(type=file):event-input-3
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file

Testcase: https://output.jsbin.com/vijukozuro/1

### DIFF
--- a/index.html
+++ b/index.html
@@ -6145,7 +6145,11 @@ must run the following steps:
       If <var>multiple</var> is <code>true</code>
       <var>files</var> are be appended to <var>element</var>’s <a>selected files</a>.
 
-     <li><p><a>Fire</a> an <a><code>input</code></a> event.
+     <li><p><a>Fire</a> these events in order on <var>element<var>:
+      <ol>
+       <li><a><code>input</code></a>
+       <li><a><code>change</code></a>
+      </ol>
 
      <li><p>Return <a>success</a> with data <a>null</a>.
     </ol>


index.html
```diff
@@ -6145,7 +6145,8 @@ <h3><dfn>Element Send Keys</dfn></h3>
       If <var>multiple</var> is <code>true</code>
       <var>files</var> are be appended to <var>element</var>’s <a>selected files</a>.
 
-     <li><p><a>Fire</a> an <a><code>input</code></a> event.
+     <li><p><a>Fire</a> an <a><code>input</code></a> and
+      <a><code>change</code></a> event.
```
andreastt: This might be OK, but I think it would be better to call out the ordering explicitly:

  1. Fire an input event on element.
  2. Fire a change event on element.

Or:

  1. Fire these events in order on element:
  1.1. input
  1.2. change

We should also change this to be specific about where to fire the element (note the use of _element_ above).

whimboo: I've made those changes.

whimboo:
I will take care of adding the test via https://bugzilla.mozilla.org/show_bug.cgi?id=1448792

whimboo:
@andreastt shouldn't we label that PR for `level 2`?

PR #1273, opened 2018-06-29
More direct links to implementation docs for Firefox and Edge

timsutton:
The link for Firefox was a manual "this content has moved to.." redirect to this bugzilla dependency tree page, so this just skips that.

For the link for Edge, I usually had to do some clicking around on the platform status page before I could finally get to the MS docs page with the implementation details for Webdriver in Edge, so this link just goes directly there.

Thanks!


andreastt:
Thanks!

PR #1274, opened 2018-07-03
Correct variables in Execute Async Script

jugglinmike:
WebDriver uses a Promise to implement a callback-based API, so we need to expose the resolving function. I couldn't find an example of this in the Promises guide or ServiceWorkers, though. This is my best guess for how it should be specified.

@Domenic What do you think?

### DIFF
--- a/index.html
+++ b/index.html
@@ -6601,7 +6601,7 @@ This is a function that, when called, returns its first argument as the response
 
   <li><p>Run the following substeps <a>in parallel</a>:
   <ol>
-    <li><p>Append resolve to <var>script arguments</var>.
+    <li><p>Append resolve to <var>arguments</var>.
 
     <li><p>Let <var>result</var> be the result of <a>promise-calling</a>
       <a>execute a function body</a>, with arguments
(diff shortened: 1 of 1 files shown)


AutomatedTester:
Gentle reminder for @domenic

jugglinmike:
I believe this patch is incomplete; gh-1275 includes further change which seems necessary

domenic:
I find the flow here pretty confusing. "execute a function body" doesn't appear to be a JS function, so promise-calling it doesn't make much sense to me.

Let's assume you don't promise-call it, since that part seems wrong. OK, then... in that case, I would not mix the promises-guide shortcut "a new Promise" with the ES CreateResolvingFunctions. Instead I'd use NewPromiseCapability(%Promise%).

The other issue is that you can't do main-thread stuff from "in parallel". That includes running JS code. See https://html.spec.whatwg.org/multipage/webappapis.html#event-loop-for-spec-authors

It's also confusing how you're returning from algorithms asynchronously ("Upon fulfillment .. return success"). Like in JS, that doesn't really work. Maybe a separate issue though.

jugglinmike:
Thanks, @domenic! I think that the more aggressive change I mentioned above (gh-1275) may be closer to the mark.

AutomatedTester:
closing as this is in #1275

PR #1275, opened 2018-07-06
Improve language describing script execution

jugglinmike:
gh-1274 includes a couple fixes for the definition of variables in the Execute
Async Script command. After further review, I found a few more changes that may
be necessary, and they effect both Execute Async Script and Execute Script.

If I'm interpreting gh-381 correctly, then both of these commands should honor
any Promise value returned by the script. That is the behavior I've tried to
implement here.

This patch subsumes gh-1274.

### DIFF
--- a/index.html
+++ b/index.html
@@ -6601,7 +6601,7 @@ This is a function that, when called, returns its first argument as the response
 
   <li><p>Run the following substeps <a>in parallel</a>:
   <ol>
-    <li><p>Append resolve to <var>script arguments</var>.
+    <li><p>Append resolve to <var>arguments</var>.
 
     <li><p>Let <var>result</var> be the result of <a>promise-calling</a>
       <a>execute a function body</a>, with arguments
--- a/index.html
+++ b/index.html
@@ -6628,29 +6628,6 @@ This is a function that, when called, returns its first argument as the response
     Otherwise, return <a>error</a> with <a>error code</a> <a>javascript error</a>
     and data <var>r</var>.
 </ol>
-
-<p>The <dfn>execute async script callback</dfn> algorithm
- is initialized with a single argument <var>webdriver callback state</var>.
- It defines a <a>function</a> with a single optional argument <var>result</var>.
- When this function is called, the following steps are run:
-
-<ol>
- <li><p>If <var>webdriver callback state</var>
-  is not in the <code>unset</code> state,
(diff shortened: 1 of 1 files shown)


jugglinmike:
Thinking about this more, I realized that fully specifying the intended behavior of Execute Async Script would require even more text. If we simply Promise-call the provided script, we won't be able to differentiate between a script that return a promise and a script that does not. Gating the algorithm on that promise would make using the "callback" function very unintuitive. Developers who wanted to use it would have to return an unresolved Promise in order to force the command to wait for their invocation of the callback. In code:

```python
# This function returns `undefined`, so Promise-calling it will create a
# Promise which is immediately fulfilled with `undeinfed`. If Execute Async
# Script observed that promise, it would continue before the callback was
# invoked.
session.execute_async_script('''
  var done = arguments[0];

  // Wait 1 second and signal completion
  setTimeout(done, 1000);
''')

# The only way to use the callback effectively would be to return a Promise
# which never settles.
session.execute_async_script('''
  var done = arguments[0];

  // Wait 1 second and signal completion
  setTimeout(done, 1000);

  // Return a promise which never settles
  return new Promise(() => {});
''')
```

I've reviewed the online discussions around adding promise support. While the desirability of "Execute Async Script" was debated, it was ultimately accepted, and I don't see anything about removing support for the callback function. [The patch that originally introduced promise support maintained that function](https://github.com/w3c/webdriver/pull/756), so it's been my assumption that we want to keep that behavior around. Removing it would certainly simplify the implementation (making it identical to the latest version of "Execute Script"), but I expect that was and is considered too drastic a change.

This is why I've extended the algorithm to only resolve on the script's return value if it is found to be a "thenable" object. Beyond making the algorithm longer, it also increases coupling with ECMAScript. I don't think this can be avoided, though I did check [the Promises guide](https://www.w3.org/2001/tag/doc/promises-guide) and [WebIDL](https://heycam.github.io/webidl/#es-extended-attributes) for shorthands. I'm suggesting a non-normative note to explain why we're effectively re-implementing "Promise call."

As before, @domenic's familiarity with using Promises in specifications would help here. Do you have any time to look this over, Domenic?

AutomatedTester:
mergedf by hand so I could preserve the commit series

jugglinmike:
Thanks for taking the time to do that!

Issue #1287, opened 2018-08-15
HTTP status code 408 causes clients to retry


barancev:
[Section 6.6](https://w3c.github.io/webdriver/#handling-errors) contains two error codes "script timeout"and "timeout" mapped to HTTP status code 408 Request Timeout.

It might be logically correct to use this error code, but we have a collision with HTTP semantics implemented in HTTP clients. To support keep-alive we allow retries in HTTP clients, and if a client sees code 408 it thinks that the server has gone and attempts to retry the request. As a results, we have executeScript or get/refresh duplication after timeout.

What can we do?

1) Stop using keep-alive, and hit the famous port exhaustion problem.
2) Stop using code 408 and replace it with a more neutral one that does not cause retries.

andreastt:
Which HTTP code do you propose?

barancev:
Unfortunately there is no HTTP code that would perfectly fit the purpose. We can use 400, as we do for the most other errors.

barancev:
Or 500, because timeout is a server error, not a client one.

andreastt:
I would prefer 500 over 400, since the latter gives the impression of a problem with the client’s request (e.g. malformed data or invalid arguments).

Can we get some input from @kereliuk, @burg, and @jimevans?

andreastt:
Sorry, forgot @InstyleVII.

burg:
I would prefer 500 over 400, per Andreas' argument.

jimevans:
I don't know that I have a preference, but @andreastt makes a good argument. I'm perfectly happy using 500.

andreastt:
For the interested, I also wrote a blog post about this change: https://sny.no/2018/08/webdriver-keepalive

PR #1292, opened 2018-08-21
errors: change timeout errors from 408 to 500 HTTP status code

andreastt:
Whilst it is logically correct to use 408 for the timeout and script
timeout errors, it causes a collision with HTTP semantics implement
in HTTP clients.  To support Keep-Alive we allow retries in HTTP
clients and if a client sees code 408 it thinks that the server has
gone away and retries the request.

This causes Execute Script, Navigate To, and Refresh commands to
be sent twice with some HTTP clients.

This is a backwards incompatible change to WebDriver in order to
not break HTTP/1.1 Keep-Alive.

Fixes: https://github.com/w3c/webdriver/issues/1287

### DIFF
--- a/index.html
+++ b/index.html
@@ -1726,7 +1726,7 @@ It is acknowledged that this is complementary to the Evil Bit [[!RFC3514]].
 
  <tr>
   <td><dfn>script timeout</dfn>
-  <td>408
+  <td>500
   <td><code>script timeout</code>
   <td>A script did not complete before its timeout expired.
  </tr>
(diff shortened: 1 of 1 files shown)


andreastt:
This change is __backwards incompatible__ but we need a fix for this in order to support Keep-Alive and persistent connections. This PR can’t land until we have tests ready and I will supply these downstream in https://bugzilla.mozilla.org/show_bug.cgi?id=1484941.

0matko0:
zeah

andreastt:
@AutomatedTester The tests have landed.

Issue #1266, opened 2018-06-11
Need a browser.frame() variant that uses getSVGDocument() instead of contentDocument to work with svg embedded documents; no way to click on elements in an <svg> inside an <embed> tag


osmolyar:
## The problem

There is currently no way (with webdriverio) to click on elements inside an <svg> element inside of a #document in an \<svg\>  \<embed\> tag, wuth html as seen in example #3 at http://srufaculty.sru.edu/david.dailey/javascript/frames/objectTest.html.

The elements inside the svg are not accessible to Selenium/Webdriverio because they are not in the main document DOM but are relative to the #document inside the embed tag.  Any attempt to access them normally results in 'element not found'.

Trying to use browser.execute as described in examples #2 or #4 at http://aksahu.blogspot.com/2015/05/dealing-with-object-tags-in-selenium-webdriver.html doesn't work for webdriverio either (using getSVGDocument() in place of .contentDocument); any variant of example #2, using a JS click() inside the execute script, results in a "***.click() is not a function" error, even though the element the click is being called on is retrieved successfully either manually in the console via the same command or at runtime.

Trying to use variant #4, first retrieving the element into a variable and then calling browser.elementIdclick() on its .value.ELEMENT results in a "Element is no longer attached to the DOM" error, I assume because it is looking for the element relative to the main document again.

Trying to use browser.frame to switch to the 2nd 'document' as if it's a frame doesn't work either, giving  the error  "no such frame: element is not a frame" Error: A request to switch to a frame could not be satisfied because the frame could not be found.[chrome #0-0]     at frame([object Object])".

This should be supported via an analog to browser.frame() that allows switching to a document that is not defined inside an IFrame, e.g., using "getSVGDocument()" to switch to the internal document frame instead of "contentDocument" (or some other way).

My understanding is that Webdriverio depends on WebDriver's browser.frame command, hence submitting the request here.

See also https://github.com/webdriverio/webdriverio/issues/2753 

## Environment

* WebdriverIO version: 4.6.2
* Node.js version: 8.9.0
* [Standalone mode or wdio testrunner](http://webdriver.io/guide/getstarted/modes.html):  standalone
* if wdio testrunner, running synchronous or asynchronous tests: synchronous
* Additional wdio packages used (if applicable):

## Details

Describe in more detail the problem you have been experiencing, if necessary.

In addition to the above information, here are examples of the code used in the described attempts. 

html is like:
[...]

PR #1320, opened 2018-10-21
Add Switch to Shadow DOM command

AutomatedTester:
This allows the ability to move between a Shadow DOM context and the
document context.

If there is no Shadow DOM to switch to an error
No Such Shadow DOM is returned.

### DIFF
--- a/index.html
+++ b/index.html
@@ -158,6 +158,7 @@ dl.subcategories { margin-left: 2em }
    <!-- NodeList --> <li><dfn><a href=https://dom.spec.whatwg.org/#nodelist><code>NodeList</code></a></dfn>
    <!-- querySelectorAll --> <li><dfn><a href="https://dom.spec.whatwg.org/#dom-parentnode-queryselectorall"><code>querySelectorAll</code></a></dfn>
    <!-- querySelector --> <li><dfn><a href="https://dom.spec.whatwg.org/#dom-parentnode-queryselector"><code>querySelector</code></a></dfn>
+   <!-- shadowRoot --> <li><dfn><a href="https://dom.spec.whatwg.org/#dom-element-shadowroot"><code>shadowRoot</code></a></dfn>
    <!-- tagName --> <li><dfn><a href=https://dom.spec.whatwg.org/#dom-element-tagname>tagName</a></dfn>
    <!-- Text node --> <li><dfn><a href=https://dom.spec.whatwg.org/#text><code>Text</code> node</a></dfn>
   </ul>
@@ -1326,6 +1327,12 @@ It is acknowledged that this is complementary to the Evil Bit [[!RFC3514]].
  </tr>
 
  <tr>
+   <td>POST
+   <td>/session/{<var>session id</var>}/shadow
+   <td><a>Switch to Shadow DOM</a>
+ </tr>
+
+ <tr>
   <td>GET</td>
   <td>/session/{<var>session id</var>}/window/rect</td>
   <td><a>Get Window Rect</a></td>
@@ -1722,6 +1729,13 @@ It is acknowledged that this is complementary to the Evil Bit [[!RFC3514]].
   <td><code>no such element</code>
   <td>An element could not be located on the page
    using the given search parameters.
+ </tr>
+
+ <tr>
+   <td><dfn>no such shadow dom</dfn>
+   <td>404
+   <td><code>no such shadow dom</code>
+   <td>The element does not have a Shadow DOM attached to it.
  </tr>
 
  <tr>
@@ -3760,6 +3774,73 @@ make the tab containing the <a>browsing context</a> the selected tab.
   and compare it with the set after the action is performed.
 </aside>
 </section> <!-- /Get Window Handles -->
+
+<section>
+<h3><dfn>Switch to Shadow DOM</dfn></h3>
+
+<table class="simple jsoncommand">
+  <tr>
+    <th>HTTP Method</th>
+    <th>URI Template</th>
+  </tr>
+  <tr>
+   <td>POST</td>
+   <td>/session/{<var>session id</var>}/shadow</td>
+  </tr>
+</table>
+
+<p>The <a>remote end steps</a> are:
+
+<ol>
+
+  <li><p>Let <var>id</var> be the result of
+   <a>getting the property</a> "<code>id</code>"
+   from the <var>parameters</var> argument.
+
+  <li><p>If <var>id</var> is not <a>null</a>,
+   or an <a>Object</a> that <a>represents a web element</a>,
+   return <a>error</a> with <a>error code</a> <a>no such shadow dom</a>.
+
+  <li><p>If the <a>current browsing context</a> is <a>no longer open</a>,
+   return <a>error</a> with <a>error code</a> <a>no such window</a>.
+
+  <li><p><a>Handle any user prompts</a>
+   and return its value if it is an <a>error</a>.
+
+  <li><p>Run the substeps of the first matching condition:
+
+    <dl class=switch>
+     <dt><var>id</var> is <a>null</a>
+     <dd>
+      <ol>
+       <li><p>Set the <a>current browsing context</a> to
+        the <a>current top-level browsing context</a>.
+      </ol>
+
+      <dt><var>id</var> <a>represents a web element</a>
+      <dd>
+       <ol>
+        <li><p>Let <var>element</var> be the result
+         of <a>trying</a> to <a>get a known element</a>
+         by <a>web element reference</a> <var>id</var>.
+
+        <li><p>If <var>element</var> <a>is stale</a>,
+         return <a>error</a> with <a>error code</a> <a>stale element reference</a>.
+
+        <li><p>Let <var>shadowRoot</var> be the result of getting
+        <var>element</var>'s <a><code>shadowRoot</code></a>.
+
+        <li><p>If <var>shadowRoot</var> is <a>null</a>
+          return <a>error</a> with <a>error code</a> <a>no such shadow dom</a>.
+
+        <li><p>Set the <a>current browsing context</a> to <var>element</var>’s
+         <a><code>shadowRoot</code></a>.
+       </ol>
+    </dl>
+
+  <li><p>Return <a>success</a> with data <a>null</a>.
+</ol>
+</section> <!-- /Switch to Shadow DOM -->
 
 <section>
 <h3><dfn>Switch To Frame</dfn></h3>


index.html
```diff
@@ -3761,6 +3775,73 @@ <h3><dfn>Get Window Handles</dfn></h3>
...
+         return <a>error</a> with <a>error code</a> <a>stale element reference</a>.
+
+        <li><p>Let <var>shadowRoot</var> be the result of getting
+        <var>element</var>'s <a><code>shadowRoot</code></a>.
+
+        <li><p>If <var>shadowRoot</var> is <a>null</a>
+          return <a>error</a> with <a>error code</a> <a>no such shadow dom</a>.
+
+        <li><p>Set the <a>current browsing context</a> to <var>element</var>’s
+         <a><code>shadowRoot</code></a>.
```
andreastt: What is the `shadowRoot` type? What if it is closed/open?

index.html
```diff
@@ -1325,6 +1326,12 @@ <h3>List of Endpoints</h3>
   <td><a>Switch To Parent Frame</a></td>
  </tr>
 
+ <tr>
+   <td>POST
+   <td>/session/{<var>session id</var>}/shadow
```
gsnedders: needs to be on the element

index.html
```diff
@@ -3761,6 +3775,73 @@ <h3><dfn>Get Window Handles</dfn></h3>
...
+<h3><dfn>Switch to Shadow DOM</dfn></h3>
+
+<table class="simple jsoncommand">
+  <tr>
+    <th>HTTP Method</th>
+    <th>URI Template</th>
+  </tr>
+  <tr>
+   <td>POST</td>
+   <td>/session/{<var>session id</var>}/shadow</td>
```
andreastt: This should be on a web element namespace.

index.html
```diff
@@ -3761,6 +3775,73 @@ <h3><dfn>Get Window Handles</dfn></h3>
...
+  <li><p>If the <a>current browsing context</a> is <a>no longer open</a>,
+   return <a>error</a> with <a>error code</a> <a>no such window</a>.
+
+  <li><p><a>Handle any user prompts</a>
+   and return its value if it is an <a>error</a>.
+
+  <li><p>Run the substeps of the first matching condition:
+
+    <dl class=switch>
+     <dt><var>id</var> is <a>null</a>
```
andreastt: The `id` will always be defined if it’s a web element.

justinfagnani:
The switching API for Shadow DOM is really suboptimal and adds a lot of unnecessary overhead to testing Web Components. a ShadowRoot is not an iframe, nor even like an iframe, and it shouldn't have a similar API.

Please see my objections in #5869: https://github.com/SeleniumHQ/selenium/issues/5869#issuecomment-400854187

jlipps:
So far I think @justinfagnani has a good point; I'm curious what your response would be, @andreastt and @AutomatedTester.

jgraham:
This was briefly discussed in the working group meeting today. I think there was some agreement that this might not be the best API approach, but some thought is needed about how interactability checks should work with shadow roots.

christian-bromann:
I am curious how easy it is for browser vendors to find an element within a shadow DOM. In the CDP I don't find any method like `querySelectorAll` that includes finding an element within a shadow DOM. Not sure about Gecko. I see two scenarios:

1. We include searching in shadow DOMs when using the findElement command and leave it up to the browser driver to also check within every shadow DOM of every element on the page that has one.
    - __Pro__: almost no work for WebDriver bindings to add this functionality
    - __Con__: heavy work for driver to look into every shadow DOM to find the element
2. We do it as proposed here and let the WebDriver bindings ensure that they implement it correctly
    - __Pro__: more efficient and easier to implement for drivers
    - __Con__: WebDriver bindings need to implement an API that takes a selector for the element containing the shadow DOM and the selector to find the element within the shadow DOM.

The question is also whether or not we need to switch context in order to execute an action on an shadow DOM element. This can also add a lot of work on the client libraries having to flag elementIds as shadow elements or non shadow elements and having to perform a context switch before using that element id.

TLDR: depending on how easy it is for browser vendors to find a shadow dom and execute an action on it we should either abstract it away or have the WebDriver bindings have to implement an abstraction on top of it.

jgraham:
Looking inside shadow doms should be an explicit action because it's breaking encapsulation. It should be quite acceptable to treat a `<my-button>` element with associated shadow dom as a single entity which can be interacted with just like a UA-defined element. But it should also be possible to pierce the veil and interact with the implementation of the component when that's a necessary thing to do. I don't think that implemenation complexity is the overriding concern here, but presenting an API that is both easy to use and has the desired featureset. The problem with the model where you scope all commands to a session-global shadow DOM is that it fails the test of being easy to use. The counter-proposal (roughly) is to have two possible web-elements corresponding to a DOM node with a shadow tree; one represeting the host document view of the element, and one representing the shadow root. So you would have an endpoint like `/session/{session id}/element/shadow` which would return a `Shadow Root Web Element`. Passing the `element id` of that `shadow root web element` to `/session/{session id}/element/{element id}/element` would allow searching for elements within the shadow tree (which would then return normal `web element` references that could be passed to other commands).

andreastt:
Treating web elements inside a shadow DOM the same as any other web
element in the document makes a great deal of sense, thanks to
@jgraham for this idea.  Scoping element retrieval commands to act
on a shadow root solves the problem of overreaching global state
that is both hard to define, maintain, as well as understand.

The assumption of this PR was that introspecting a shadow DOM would
behave like switching frames, where you need to set up a certain
amount of state before you can interact with the web element, and
then clean up the same state to interact with a web element outside
the shadow DOM.

That would require an awful amount of context managing, and because
shadow DOM elements essentially belong to the same document it’s
not clear we need the extra complexity of statefully changing the
“interaction mode” as we do for browsing contexts.

jgraham:
> If an action is then called on a shadow root web element reference the context can be switched by the driver automatically.

Note that in the proposal there is no session-global state associated with shadow DOMs. The change is that for elements that represent a shadow root, there are two associated `Web Elements`, one that represents the shadow host and cannot see inside the shadow dom, and one that represents the shadow root and can. Once you have a web element for an element inside a shadow DOM, it's assumed that you can use it just like one for an element inside the host document.

That said, I could get behind adding parameters to the existing endpoints rather than adding new endpoints. One deciding factor there could be how the fallback works in existing implementations. In the case where there's a separate endpoint to get the shadow root then old implementations will fail explicitly. In the case that there's a parameter, it will be ignored by old implementations and they will silently have incorrect behaviour.

christian-bromann:
thanks for clarifying @jgraham @andreastt 

> One deciding factor there could be how the fallback works in existing implementations

Are there already existing implementations for that endpoint? I was under the assumption that there aren't any. A brief research resulted that many implementations are allowing actions in the shadow dom by using the `execute` command.

jgraham:
> Are there already existing implementations for that endpoint?

I meant the `Find Element` endpoint(s), which are implemented. If the way to get the shadow root is to call `/session/{session id}/element` with data `{"using": "css selector", "value": "my-element", "shadow": true}` then an implementation with no support for getting shadow roots would silently return the `my-element` shadow host rather than the associated shadow root, which would be incorrect behaviour. If instead we make the endpoint `/session/{session id}/element/shadow` (or something) with the data `{"using": "css selector", "value": "my-element"}` then the method would return unsupported command in non-supporting implementations. But the usability of the first does seem superior so I could be convinced it's the right thing to do.

christian-bromann:
> then an implementation with no support for getting shadow roots would silently return the `my-element` shadow host rather than the associated shadow root

Yes, which is why the default is set to `false` which results in all implementation won't change their behavior at all. Having this default also make sense because in 95% of all use cases you are not interested in looking into the shadow dom (this number can change if web components become more popular).

jgraham:
The default doesn't matter. The question is what happens in an existing implementation in the case that a user specifically requests looking at the shadow dom. With a new parameter, that parameter is ignored and we return the wrong result. With a new endpoint, that endpoint returns a `404` and an unsupported command exception.

AutomatedTester:
The one reason why this can't be on `findElement` directly is we need to take into account that if people are expecting a `shadowRoot` and don't get one then we need to handle that case. We also don't have any mechanisms on the web platform to allow us to penetrate shadow DOM from a find. This has been removed (see https://github.com/w3c/csswg-drafts/issues/640). I don't want WebDriver to go do it's own thing here.

The other reason for doing this as an explicit switch is that all interaction commands need to know the context of where it is executing. For example:

```html
<html>
  <my-list>
    <my-button>
       <fancy-list-that-is-data-driven>
    </my-button>
  </my-list>
```

If we have a mechanism that allows us to traverse down to `fancy-list-that-is-data-driven` when we do try do a click we will get either a `stale element reference` or `element click intercepted` since currently we do this from `document` where with each shadow DOM we need to set the context to `shadowRoot`. This means we need to have a similar context to where `findElement` is starting.

In Web API terms if we do `document.elementsFromPoint(...)` we get the top most element with a `shadowRoot` and then we need to move to that and redo the query. Doing it without explicit transitions means we could start adding recursive code to find what we mean which would be less than ideal.

jgraham:
> We also don't have any mechanisms on the web platform to allow us to penetrate shadow DOM from a find.

That doesn't matter; the proposed API doesn't depend on the ability to do anything other than select a shadow host from css (which is necessarily possible since it's just a normal element) and get a reference to its associated shadow root (which is possible via content APIs for the case of an open shadow root and must be possible from browser-internal APIs for a closed shadow root since that's required for the page to actually function).

> If we have a mechanism that allows us to traverse down to fancy-list-that-is-data-driven when we do try do a click we will get either a stale element reference or element click intercepted since currently we do this from document where with each shadow DOM we need to set the context to shadowRoot

Interaction APIs will presumably start from the element's nearest shadow root for the purposes of click. So in your example I would have to write code like:

```
var myList = webdriver.select("my-list", shadowRoot=true);
var myButton = myList.select("my-button", shadowRoot=true);
var myFancyList = myButton.select("my-fancy-list-which-is-data-driven");
myfancyList.click()
```

(in practice you would chain the calls rather than making lots of intermediates at least in languages that support that style, so I don't think the programmer burden of this is too high). Note that this is basically exactly the same as you could do today using script APIs to access the contents of open shadow hosts:

```
var myListHost = webdriver.select("my-list");
var myList = webdriver.execute("return arguments[0].shadowRoot");
var myButtonHost = myList.select("my-button");
var myButton = webdriver.execute("return arguments[0].shadowRoot");
var myFancyList = myButton.select("my-fancy-list-which-is-data-driven");
myfancyList.click()
```

andreastt:
> Are there already existing implementations for that endpoint? I was
> under the assumption that there aren't any.

For the sake of complete transparency there is some prior art in
Marionette, but I don’t think it should govern the direction of
this API discussion:

https://searchfox.org/mozilla-central/rev/8848b9741fc4ee4e9bc3ae83ea0fc048da39979f/testing/marionette/driver.js#2608-2630

christian-bromann:
> we need to handle that case

This could be handled similar to if no element was found.

> I don't want WebDriver to go do it's own thing here

I am not sure how to interpret this. I see that we are aligned to some web APIs (e.g. querySelector -> findElement or querySelectorAll -> findElements) but in many cases we already do our own "thing". Aligning to existing standards is definitely good but if that means it would add unnecessary complexity I would avoid it.

> This means we need to have a similar context to where `findElement` is starting.

I think this can be easily solved when we say that every `shadow root web element reference` is connected to another `web element reference` or `shadow root web element reference`. With that the driver can easily resolve all contexts as shown in @jgraham example.

gsnedders:
> Looking inside shadow doms should be an explicit action because it's breaking encapsulation.

I'm super strongly +💯 on this, to the extent I'll probably object to doing anything else.

`/session/{session id}/element/shadow` seems a bit weird to me: what exactly would it do? Find an element *and* return its shadow root?

To me it makes more sense to have `/session/{session id}/element/{element id}/shadow` which returns a Shadow Root Web Element (which is just a special type of Web Element, and Find Element from Element etc. work normally on it; or null) for the given element id, splitting the finding and returning the shadow root into separate actions.

I also think the harder part is defining element interactibility (given a `<input type=text>` should still be considered interactable if it's within an shadow tree).

justinfagnani:
I want to point out that in the standards groups the assumption has been that testing frameworks will provide a means to pierce ShadowRoots in searches, which is why it was ok to remove the shadow-piercing selectors like `/deep/`. I can try to dig up minutes if you like. I know that it was discussed at TPAC 2017 in the Web Components breakout.

For open ShadowRoots, the encapsulation isn't intended to be strong enough to disallow code that knows that it's doing from reaching in. It's possible, and definitely not discouraged for testing, to write a `querySelectorDeep(selectors: string[])` function that recurses through ShadowRoots applying each selector in turn.

jgraham:
`/session/{session id}/element/{element id}/shadow` looks more like the DOM APIs, but typically has more overhead since if you find an element then you have to make a second call to get the shadow root. It has the advantage that it works when you return a (closed) shadow host from a script and want to get into the shadow dom.

For reference CDP seems to do something more like "give me the children of this node, including any shadow children", which does make sense for a debugger, but doesn't seem to map well to WebDriver since WebDriver only has selection by (various kinds of) selector not direct DOM traversal.

gsnedders:
I guess my suggestion is entirely redundant, anyway, given `wd.execute_script("return arguments[0].shadowRoot", element)` will do the same.

Oh well. I don't know well enough what makes sense here.

Given @justinfagnani said the Web Platform WG expected testing frameworks to provide a means to pierce shadow roots in searches, it would seemingly imply we either need to alter the semantics of CSS/XPath selectors or invent a new API to do something else?

justinfagnani:
> alter the semantics of CSS/XPath selectors

One option is to just implement `/deep/` and `::shadow`. I think Ryosuke Niwa from Apple has a userland implementation of that somewhere, but it's not too difficult. Another is to take a list of selections, with an implied `/deep/` between each.

christian-bromann:
@gsnedders any comments on `/session/{session id}/element/{element id}/shadow` endpoint vs. `/session/{session id}/element` with data `{"using": "css selector", "value": "my-element", "shadow": true}`? I guess it depends on finding a solution for what @jgraham said [here](https://github.com/w3c/webdriver/pull/1320#issuecomment-433935315)

christian-bromann:
Latter could be also defined as `/session/{session id}/element` with data `{"using": "css selector", "value": "my-element", "deep": true}`

gsnedders:
@christian-bromann We should add a new path, because silently returning wrong data is Bad.

jlipps:
I agree that sending a new JSON param to an older endpoint, and having it happily and silently ignore the request for a shadow element, would lead to confusion.

@christian-bromann I also don't see how your proposal makes it any easier for client library authors to support the new feature either; maybe I'm missing something?

justinfagnani:
I don't know WebDriver details that well, but hopefully my drive-by comments on Shadow DOM are helpful...

Re: `{"using": "css selector", "value": "my-element", "deep": true}`
Aside from the comments about a new parameter to an existing endpoint, I think it's worth a note deep query semantics. You can't easily just add a flag to a selector to say that it should pierce shadow roots because you need to know what parts of a query should pierce.

Given a compound selector like `a b c`, a "deep" flag could be interpreted to mean `/deep/ a b c` (select `a b c` in any shadow root), or `/deep/ a /deep/ b /deep/ c` (basically ignore shadow boundaries completely). The former isn't expressive enough, the later is too loose (and too expensive to execute).

Specific shadow-piercing combinators allow the query to specify exactly what sub-selectors should be matched in a shadow root. This is why you'd either need to actually support a combinator, or take a list of selectors, with `/deep/` implied between each. Then a query could be `a b /deep/ c`, which a single flag wouldn't be able to support.

jgraham:
I think we shoulc do this in two phases.

1. Have a way to get the shadow root attached to a shadow host element (by either writing an endpoint that returns the shadow root given an element, or by writing one that selects elements like normal but returns shadow roots rather than the host elements).
2. Consider extending selectors so it's possible to pierce the shadow boundaries directly.

The first part is a relatively simple change that gives us all the functionality but at some possible cost of usability. The second is a much more complex proposition which makes some cases more efficient but required a selector that only works from test code, which afaik is a rather new construct.

andreastt:
> Consider extending selectors so it's possible to pierce the shadow
> boundaries directly.


I think it’s a novel idea to allow (CSS/XPath?) selectors in WebDriver
to pierce the shadow DOM boundary, but it would (certainly in Gecko’s
case) require buy-in from the people working on DOM as it does come
with some security implications.

codingphil:
> I think we shoulc do this in two phases.
> 
> 1. Have a way to get the shadow root attached to a shadow host element (by either writing an endpoint that returns the shadow root given an element, or by writing one that selects elements like normal but returns shadow roots rather than the host elements).
> 2. Consider extending selectors so it's possible to pierce the shadow boundaries directly.
> 
> The first part is a relatively simple change that gives us all the functionality but at some possible cost of usability. The second is a much more complex proposition which makes some cases more efficient but required a selector that only works from test code, which afaik is a rather new construct.

In my opinion if Web Components and shadow DOM find it's way to broader usage and you have an application which heavily uses shadow DOM with deep hierarchies (at each component boundary), for example see the Chrome settings page, then the 1st approach with which you have to be explicit when crossing the shadow DOM boundary would be just a usability nightmare.

jgraham:
> the 1st approach with which you have to be explicit when crossing the shadow DOM boundary would be just a usability nightmare.

Perhaps, but we should let experience rather than supposition guide us here. It makes a great deal more sense to me to do the simple thing now and the complex thing if it proves necessary rather than the complex thing upfront.

> I think it’s a novel idea to allow (CSS/XPath?) selectors in WebDriver to pierce the shadow DOM boundary, but it would (certainly in Gecko’s case) require buy-in from the people working on DOM as it does come with some security implications.

Are you thinking of browser UI implemented in shadow DOM? I assume that would remain opaque in any case.

christian-bromann:
> Perhaps, but we should let experience rather than supposition guide us here.

I agree. Especially since WebDriver implementations (e.g. Selenium or WebdriverIO) can build some nice API interface on top of it that hides the complexity for the purpose of testing.

justinfagnani:
> I think we shoulc do this in two phases.
> 
> 1. Have a way to get the shadow root attached to a shadow host element (by either writing an endpoint that returns the shadow root given an element, or by writing one that selects elements like normal but returns shadow roots rather than the host elements).
> 2. Consider extending selectors so it's possible to pierce the shadow boundaries directly.

This sounds reasonably pragmatic to me. I think a workaround for the initial usability hit would be to ensure a script evaluated via execute can return a ShadowRoot. Then a test could load userland shadow-piercing code of their choice.

christian-bromann:
We have a similar use case with SVGs here: https://github.com/w3c/webdriver/issues/1266

AutomatedTester:
It appears that how to handle `elementsFromPoint` is currently unspec'ed see https://github.com/w3c/csswg-drafts/issues/556

This is important as it plays into the `interactability checks` used for interactions

AutomatedTester:
The consensus from CSSWG is that we should not be allowed to pierce shadow DOMs that are closed. This is why piercing queryselectors were dropped. 

> AutomatedTester> hey smaug, Do you remember the reason why shadow dom piercing query selectors were dropped?
> 9:39 AM 
> <•smaug> AutomatedTester: because they reveal closed shadow doms
> 9:39 AM 
> <•smaug> which breaks encapsulation 

This doesn't suggest they were leaving it up to testing frameworks. I can't find scribed notes that suggest that.

AutomatedTester:
Mozilla's implementation bug around `elementsFromPoint` is https://bugzilla.mozilla.org/show_bug.cgi?id=1502369

jgraham:
To be clear, if `elementsFromPoint` doesn't pierce into shadow DOMs, we have a problem with the definition of interactable irrespective of the model we use for switching into the shadow tree. So that's definitely a problem that needs to be addressed writing the spec here, but afaict it doesn't obviously place additional constraints on how this should work.

justinfagnani:
> The consensus from CSSWG is that we should not be allowed to pierce shadow DOMs that are closed. This is why piercing queryselectors were dropped.

That's closed ShadowRoots. The vast majority are open.

justinfagnani:
Having test frameworks use useland shadow-piercing selector libraries was discussed at TPAC 2017, and several other times in in-person conversations:

> Ryosuke: We have lots of test APIs in WebKit. 
> ... Turns out a lot of apps end up using those test APIs. 
> ... That's not good, not thread safe. 
> ... If we expose it in static profile, someone's going to use it in production somewhere. 
> ... Maybe we should standardize polyfill or something (for test frameworks).

https://www.w3.org/2017/11/10-webplat-minutes.html

AutomatedTester:
> > The consensus from CSSWG is that we should not be allowed to pierce shadow DOMs that are closed. This is why piercing queryselectors were dropped.
> 
> That's closed ShadowRoots. The vast majority are open.

The point is the piercing would break encapsulation.

coreyfarrell:
What is the desired outcome of `getText()` run on an element which only contains shadowRoot?  Currently `chromedriver` returns the text contents from within the shadowRoot (including recursively), `geckodriver` returns an empty string.  Minimum example is posted at mozilla/geckodriver#1417.

andreastt:
The behaviour in relation to Get Element Text is somewhat undefined. However the browser implements shadow DOM, the expected outcome should be whatever running `bot.dom.getVisibleText` as defined in https://github.com/SeleniumHQ/selenium/blob/e09e28f016c9f53196cf68d6f71991c5af4a35d4/javascript/atoms/dom.js#L981 gives.

Get Element Text is _horribly_ underdefined because we couldn’t reach agreement on what constituted visible text.

coreyfarrell:
Even so I if I have:
```js
const fragment = '<div>any html fragment</div>';
document.getElementById('test1').innerHTML = fragment;
document.getElementById('test2').attachShadow({mode:'open'}).innerHTML = fragment;
```
I think `getText` should return the same value for `test1` and `test2`.  Even if different elements of W3C disagree on what constitutes visible text, each browser should agree with itself.  Putting `fragment` in a shadowRoot does not change the [element’s text “as rendered”](https://w3c.github.io/webdriver/#dfn-get-element-text).

gsnedders:
Neither `textContent` nor `innerText` get anything from shadow trees, do they? Should we expect Get Element Text to differ?

coreyfarrell:
They do not.  I think `textContent` is not worth comparing since it returns the contents of non-visible childNodes such as `<style>`.

In reality I'm unable to get rendered text from within a shadow tree from any browser, only chromedriver `Get Element Text` does what I need.  `innerText` returns a blank string on both Chrome and Firefox.  As is a test helper would be needed to reimplement the logic of innerText by iterating `ele.shadowRoot.childNodes` and ignoring non-visual and hidden elements.  Unfortunately `innerText` does not return blank when requested directly from non-visual childNodes so the following prints `div{display:none}test1test2` to the console twice.
```js
ele.attachShadow({mode: 'open'}).innerHTML = '<style>div{display:none}</style><div>test1</div>test2'
console.log(Array.from(ele.shadowRoot.childNodes).map(e => e.innerText || e.wholeText).join(''));
console.log(ele.shadowRoot.textContent);
```

I'm not sure what the correct answer is to how `Get Element Text` should respond for elements containing shadow trees.  I can understand the argument for not wanting the existing command to pierce shadow trees, but I also feel that the ability to do so (somehow) is important for testing.  I tried using [Take Element Screenshot](https://w3c.github.io/webdriver/#take-element-screenshot) with a test that compared the screenshot to expected (previous) image, unfortunately the fonts on Travis-CI are slightly different than my workstation so the screenshots cannot match.  I can't think of any other options.

jgraham:
I think the behaviour of get element text should be a separate issue. This issue should be about how to access elements in the shadow tree, and the other issue about how to access text under the shadow tree. Depending on the constraints we find, the solutions may be unrelated (i.e. if it's possible to update Get Element Text to at least optionally return the shadow tree's text, that won't depend at all on the ability to access elements in the shadow tree).

jgraham:
Filed https://github.com/w3c/webdriver/issues/1350 for the Get Element Text issue.

lukebjerring:
@AutomatedTester - what's the status on this?
Ran into this issue myself testing wpt.fyi, which I worked around using execute script of `arguments[0].shadowRoot.querySelector('foo')` to fetch foo from a web-component element.

AutomatedTester:
I need to pick this up again. I will look at it next week.

David
<http://www.theautomatedtester.co.uk/>

jrobinson01:
I'm not sure this is the best place for this, but here goes anyway. I'm using webdriverio and have added this simple custom command:
```javascript
browser.addCommand('getShadowRoot', function(el) {
  return browser.execute(el => el.shadowRoot ? el.shadowRoot : el, el.value);
});
```
Usage is sort of clunky:
```javascript
get myButton() {
    browser.getShadowRoot(browser.element('my-container')).element('my-button');
}
```

but I think it could be much nicer, if in webdriverio it were a getter function on the `element` object returned by `browser.element('selector')`. This would allow the following usage (as part of a page object in webdriverio for example):

```javascript
// get the custom element
get myElement() {
    return browser.element('my-element');
}
// query inside it's shadowRoot
get myButton() {
    return this.container.shadowRoot.element('my-button');
}
```
The last example basically lines up with how things work inside the browser:
```javascript
document.querySelector('my-element').shadowRoot.querySelector('my-button');
```

osmolyar:
Any update on this?  Still stuck on

https://github.com/webdriverio/webdriverio/issues/2753 
 https://github.com/w3c/webdriver/issues/1266, 
https://github.com/FuckTheWorld/chromedriver/issues/427#issuecomment-498290042

junguchina:
Is there any update here？Who can push this pr？

osmolyar:
@junguchina , not sure if this workaround may help your case:  https://github.com/webdriverio/webdriverio/issues/2753#issuecomment-499949444

letsgotomars:
Any updates for us on the status of this PR @AutomatedTester, @shs96c, or @jgraham? It's been two years and it would be nice to know what the path forward is.

I feel that this is a very important issue that needs to be addressed in order for developers to truly adopt and embrace Shadow DOM seriously. Shadow DOM enabled web components are one most exciting and powerful features of the modern web, but sadly if developers cannot properly or easily write tests for them without going through kludgy or complex workarounds then their adoption will be crippled since the cons will outweigh the pros in many cases. More on this [here](https://github.com/SeleniumHQ/selenium-ide/issues/702).

AutomatedTester:
> Any updates for us on the status of this PR @AutomatedTester, @shs96c, or @jgraham? It's been awhile since this PR started and it would be nice to know what the path forward is.
> 
> I feel that this is a very important issue that needs to be addressed in order for developers to truly adopt and embrace Shadow DOM seriously. Shadow DOM enabled web components are one most exciting and powerful features of the modern web, but sadly if developers cannot properly or easily write tests for them without going through kludgy or complex workarounds then their adoption will be crippled since the cons will outweigh the pros in many cases. More on this [here](https://github.com/SeleniumHQ/selenium-ide/issues/702).

There was an offer from Salesforce to tackle this.

This is really not something the WebDriver spec should be tackling as it should have been handled by the DOM spec and all the work around web components. We are trying to rectify the issue now.

AutomatedTester:
Closing this for now. This feature will be added but hopefully Salesforce will do it soon

letsgotomars:
Thanks so much for the update @AutomatedTester!

I have one question:

> This is really not something the WebDriver spec should be tackling as it should have been handled by the DOM spec and all the work around web components. We are trying to rectify the issue now.

First of all it's great to hear this is being actively worked on! How can I track the progress of this DOM/web components work and the aforementioned Salesforce team work? Are there relevant issues/PRs that me and others can reference?

leobalter:
Plugging some updates here.

We do have a merged PR for [Supporting Shadow DOM](https://github.com/w3c/webdriver/pull/1565), and respective [WPT support](https://github.com/web-platform-tests/wpt/pull/27132). Thanks @jimevans!

It seems that we still need browser implementations to catch up with the changes. @letsgotomars I believe the next step is looking at how we can help browser vendors now the tests have landed.

PR #1323, opened 2018-10-26
Actions: invoke all tick actions in parallel

jugglinmike:
Previously, the "dispatch tick actions" algorithm specified a
dispatching just one "tick action"--whichever action happened to be
referenced by the final iteration of the preceding loop.

Update the algorithm to dispatch every action unless interrupted by an
error.


jugglinmike:
I think we need to go a bit further, actually: serial action dispatch within
ticks doesn't seem desirable. I've pushed up another commit to dispatch those
actions in parallel.

I can't find any text in the latest revision which definitively describes an
intention for parallel action dispatch. Most actions describe singular events,
so parallelism has no effect.

However, the "pointerMove" action may comprise several distinct events over
some duration. That makes parallelism important--without it, there doesn't seem
to be any benefit to having multiple "tracks" (since all tracks could be
collapsed in to one without any effect on the final schedule).

If that's right, then we may also want to also extent the informative text to
call it out.

andreastt:
This actually reflect how it is implemented in Marionette: https://searchfox.org/mozilla-central/source/testing/marionette/action.js

There’s no intention for parallelism _inside_ an action tick: the intention is that the discrete events that happen inside pointerMove are meant to be treated as one “parallel unit”.

Put different, there is a desire for parallelism on the tick level. For each tick, perform each action primitive in parallel and wait for them all to complete before moving onto the next tick in the sequence. The exact execution order of actions in a vertical time slot or tick is therefore undefined, but obviously at some level they will occur in (some) sequence at the leisure of scheduling and the sequenitial nature of the event queue.

This PR seems to capture my mental model of this is supposed to work, so thanks for putting in the effort! 👍🏻 I don’t think there’s much we can do for testing this, apart from the tests we already have.

jugglinmike:
> I don’t think there’s much we can do for testing this, apart from the tests we already have.

I came across this while researching an infrastructure test in WPT. If this change lands, then we'll need to update that test: https://github.com/web-platform-tests/wpt/pull/13741

> This PR seems to capture my mental model of this is supposed to work, so thanks for putting in the effort!

My pleasure :)

jugglinmike:
@andreastt I've expressed the scheduling with promises because that's how I'm
used to specifying asynchronous operations. The values aren't actually exposed
to the caller, which might make Promises a confusing choice for use in this
algorithm.  I could see objection to this along the same lines as if I had
defined the algorithm in JavaScript and written, "evaluate this string." That's
hyperbolic, but do you see what I'm saying? Do you think it's a concern?

jugglinmike:
I spoke with @annevk about this on IRC briefly. They suggested using a task
queue, although they weren't aware this was in the context of WebDriver. I can
see how a task queue would be more direct, but I'm not sure this specification
has a task queue to begin with. If not, then we probably can't use Promises,
either. That may have implications for the way we've specified "Execute
Script".

jgraham:
I'm confused. The mental model that I have (and hence what the spec intended to say) is this:

* Actions are provided on-wire grouped by device and ordered by time
* Internally the input is transposed so that we get them grouped by time and ordered by device.
* Actions that have a duration are asynchronous within the action itself.

So ordering of events generated by "instantaneous" actions is well defined. If I have two pointer input devices and they both register a `pointerDown` in the same tick, the event from whichever is the first device on the wire is defined to be the first to fire, always. The confusion only comes for `pointerMove` because that is asynchronous inside the tick (assuming a non-zero duration); it schedules a series of events but doesn't block. So if you look at the event queue, the first movement of the first defined pointer is always first, but the overall action isn't blocking.

I don't think those invariants should be relaxed.

andreastt:
My understanding was initially that this PR did not relax the
principle that actions are scheduled in sequence, grouped first by
device and then by tick/time. However, I suppose _scheduling_ does
not necessarily constitute the _completeness_ of an action, and
this is where this PR may be at odds.

As you say this is not normally a problem as most actions execute
and complete immediately. There are two actions that don’t complete
immediately, but which internally executes a sequence of steps, and
these are _pause_ (wait for an arbitrary amount of time) and
_pointerMove_ (simulate multiple mousemove DOM events within the
computed overall duration of the tick/frame). Is it your assumption
that these actions must be allowed to complete before before the
next device source’s action _within_ the same tick/frame?

jgraham:
No. https://w3c.github.io/webdriver/#dfn-dispatch-a-pointermove-action step 10 talks about asynchronously waiting before starting the action and https://w3c.github.io/webdriver/#dfn-perform-a-pointer-move talks about performing a small delta of the action and then asynchronously waiting again. The code flow is expected to be something like (pseudocode, not exactly matching the spec)

```js
async function dispatchPointerMove(device, target, duration) {
  let deltaT = duration / NUM_STEPS;
  let deltaPosition = (target - device.getCurrentPointerPosition()) / NUM_STEPS;
  while i < numSteps {
    let pointerPos = device.getCurrentPointerPosition();
    device.movePointer(pointerPos + deltaPosition);
    await setTimeout(deltaT);
  }
}
```

So for pointerMove on multiple devices (e.g. multiple fingers), you would have a tick-running function that looked like
```js
async function runTick(actions) {
  await Promise.all(actions.map((action) => action));
}

await runTick([dispatchPointerMove(finger1, finger1Target, duration), dispatchPointerMove(finger2, finger2Target, duration)]);
```

In this toy implemenation approach, overall tick duration is handled in a similar way (i.e. the tick-running function that returns a promise that resolves after a timeout) and instantaneous function return promises that are already resolved.

Does that make things any clearer?

jugglinmike:
@jgraham The term "asynchronously wait" is defined in terms of the active
browsing context's event loop. I don't understand how that relates to the
WebDriver processing model, though.

I interpreted it to be a blocking wait for WebDriver. That's why I proposed an
explicit mechanism to run each "tick action" in parallel.

Your example's modeling of `dispatchPointerMove` as an async function makes it
clear that it is *not* intended to be a blocking wait for WebDriver.

That interpretation makes the second commit inappropriate. The first is
still necessary, but maybe not sufficient. "Dispatch a pointerMove action"
needs to return "success", and I *think* the effect on the WebDriver processing
model needs to be reflected in the algorithm. What do you think about removing
the second commit, keeping the first, and adding the following change to
"dispatch a pointerMove action":

```diff
diff --git a/index.html b/index.html
index f812af0..04168a0 100644
--- a/index.html
+++ b/index.html
@@ -8601,35 +8601,41 @@ is also removed.
   otherwise.
 
+ <li><p>Run the following substeps <a>in parallel</a>:
+
+ <ol>
    <li><p>If <var>duration</var> is greater than 0 and inside any
     implementation-defined bounds, <a>asynchronously wait</a> for an
     implementation defined amount of time to pass.
 
     <aside><p>This wait allows the implementation to model the overall
      pointer move as a series of small movements occurring at an
      implementation defined rate (e.g. one movement per
      vsync).</p></aside>
    </li>
 
    <li><p><a>Perform a pointer move</a> with arguments
     <var>source id</var>, <var>input state</var>, <var>duration</var>,
     <var>start x</var>, <var>start y</var>, <var>x</var>, <var>y</var>.
+ </ol>
+
+ <li><p>Return <a>success</a> with data <a>null</a>.
 </ol>
 
 <p>When required to <dfn>perform a pointer move</dfn> with
```

jgraham:
Oh, I see the confusion. The intent is that whilst you are asynchronously waiting you return control to the caller and keep running the other steps, returning to the following steps when the wait expires. But that isn't spelled out explictly. So I think a PR to clear up that confusion is what's required in this case. Maybe specifying that the steps are run `<a>in parallel,/a>` is indeed sufficient.

jugglinmike:
Cool--I've reverted the second commit and added a fixup for the first. If that
looks good to you folks, GitHub's "squash and merge" feature should produce a
commit that's suitable for `master`.

jugglinmike:
On second thought, as soon as we go "in parallel," we lose control over
execution sequence. That means that if we want to make the sequence of "first
movements within a tick" deterministic, we can't use "in parallel" from the
"dispatch a pointerMove" algorithm.

I've added another commit that addresses this by pushing the parallelism down
into the very end of "perform a pointer move." This seems like the easiest way
to preserve the invariant, but it's also a little indirect (and it introduces
more parallelism than we actually care about). That's why I've also added a
note to explain the intention.

Issue #1138, opened 2017-11-01
Add command to open a new top-level browsing context


whimboo:
Currently opening a new tab/window can only be done by first loading a web page, which opens links with the target `_blank`. But then it is not clear if it will become a new tab or a new window. It all depends on the settings of the Browser in how `_blank` is handled.

To make it easier to open a tab or window, we should introduce a new window context related command like `open` (similar to `close` for closing). It would need an argument for the type of new context, and by default it should assume a new tab to save memory resources.

Please note that formerly key strokes like `Ctrl+T` and `Ctrl+N` were working, but those are not allowed anymore from within the context of the web page. Browsers like Firefox restrict those shortcuts.

andreastt:
Thanks for filing this issue. I was surprised to find we didn’t have it filed already.

In our previous discussions there was resounding agreement that WebDriver should offer commands for creating new windows. `let newWin = window.open(…)` is not a great substitute because of security escalation issues.

yucer:
> In our previous discussions there was resounding agreement that WebDriver should offer commands for creating new windows.

@andreastt I suppose it would be good to offer also commands to handle Tabs, Bookmarks, Feeds, Downloads and all the other stuff that are common in all the browsers and do not belong to the document content.

whimboo:
The webdriver spec is about accessing and operating with web content, and not controlling the browser itself. Everything outside of content, has to be handled somewhere else, and for now is up to the driver vendor in how and when to get it implemented.

Also please let us not drift away from the original request on this issue. Thanks.

yucer:
> The webdriver spec is about accessing and operating with web content, and not controlling the browser itself.

Sorry I was confused by the the goal defined in the standard:

> WebDriver is a remote control interface that enables introspection and control of user agents. It provides a platform- and language-neutral wire protocol as a way for out-of-process programs **to remotely instruct the behavior of web browsers.**
> 
> Provided is a set of interfaces to discover and manipulate DOM elements in web documents *and to control the behavior* of a user agent. It is primarily intended to allow web authors to write tests that automate a user agent from a separate controlling process, but may also be used in such a way as to allow in-browser scripts to control a — possibly separate — browser. 

Also the [section 9](https://www.w3.org/TR/webdriver/#navigation) contain navigation commands like going [Back](https://www.w3.org/TR/webdriver/#back) and [Forward](https://www.w3.org/TR/webdriver/#forward) over the browsing history.

[...]

PR #1324, opened 2018-10-26
Add a endpoint for creating a new window.

jgraham:
This adds an endpoint at /session/<session id>/window/new that opens a
new top-level browsing context. This takes a single parameter "type"
that is used as a hint as to the kind of window to create within the
application. The return value includes the handle of the new window and
the type of the window that was created.

Fixes https://github.com/w3c/webdriver/issues/1138

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/pull/1324.html" title="Last updated on Nov 22, 2018, 5:58 PM GMT (b9160cd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1324/be19393...b9160cd.html" title="Last updated on Nov 22, 2018, 5:58 PM GMT (b9160cd)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -761,6 +761,12 @@ when it receives a particular <a>command</a>.
 
  <tr>
   <td>POST</td>
+  <td>/session/{<var>session id</var>}/window/new</td>
+  <td><a>Create Window</a></td>
+ </tr>
+
+ <tr>
+  <td>POST</td>
   <td>/session/{<var>session id</var>}/frame</td>
   <td><a>Switch To Frame</a></td>
  </tr>
@@ -3229,6 +3235,76 @@ make the tab containing the <a>browsing context</a> the selected tab.
   and compare it with the set after the action is performed.
 </aside>
 </section> <!-- /Get Window Handles -->
+
+<section>
+<h3><dfn>Create Window</dfn></h3>
+
+<table class="simple jsoncommand">
+ <tr>
+  <th>HTTP Method</th>
+  <th>URI Template</th>
+ </tr>
+ <tr>
+  <td>GET</td>
+  <td>/session/{<var>session id</var>}/window/new</td>
+ </tr>
+</table>
+
+<p>Create a new <a>top-level browsing context</a>.
+
+<p>The <a>remote end steps</a> are:
+
+<ol>
+ <li><p>If the implementation does not support creating new top-level
+  browsing contexts, return <a>error</a> with <a>error code</a>
+  <a>unsupported operation</a>.
+
+ <li><p>If the <a>current top-level browsing context</a> is <a>no longer open</a>,
+  return <a>error</a> with <a>error code</a> <a>no such window</a>.
+
+ <li><p><a>Handle any user prompts</a> and return its value if it is an <a>error</a>.
+
+ <li><p>Let <var>type hint</var> be the result of <a>getting the property</a>
+  "<code>type</code>" from the <var>parameters</var> argument.
+
+ <li><p>Create a new <a>top-level browsing context</a> by running
+  the <a>window open steps</a> with <var>url</var> set to
+  "<code>about:blank</code>", <var>target</var> set to the empty
+  string, and <var>features</var> set to "<code>noopener</code>" and
+  the user agent configured to create a new browing context. This must
+  be done without invoking the <a>focusing steps</a> for the created browsing
+  context. If <var>type hint</var> has the value "<code>tab</code>",
+  and the implementation supports multiple browsing context in the
+  same OS window, the new browsing context should share an OS window
+  with the <a>current browsing context</a>. If <var>type hint</var> is
+  "<code>window</code>", and the implementation supports multiple
+  browsing contexts in seperate OS windows, the created browsing
+  context should be in a new OS window. In all other cases the details
+  of how the browsing context is presented to the user are
+  implementation defined.
+
+ <li><p>Let <var>handle</var> be the
+  associated <a>window handle</a> of the newly created window.
+
+ <li><p>Let <var>type</var> be "<code>tab</code>" if the newly created
+  window shares an OS-level window with the <a>current browsing
+  context</a>, or "<code>window</code>" otherwise.
+
+ <li><p>Let <var>result</var> be a new JSON <a>Object</a> initialised with:
+
+ <dl>
+  <dt>"<code>handle</code>"
+  <dd>The value of <var>handle</var>.
+
+  <dt>"<code>type</code>"
+  <dd>The value of <var>type</var>.
+
+ </dl>
+
+ <li><p>Return <a>success</a> with data <var>result</var>.
+</ol>
+
+</section> <!-- /Create Window -->
 
 <section>
 <h3><dfn>Switch To Frame</dfn></h3>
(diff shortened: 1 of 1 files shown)


foolip:
Does this spec follow a testing policy for normative changes, i.e., is a corresponding PR on web-platform-tests required to merge this?

jgraham:
Yes.

index.html
```diff
@@ -3761,6 +3767,66 @@ <h3><dfn>Get Window Handles</dfn></h3>
...
+  browsing context should share an OS window with the <a>current
+  browsing context</a>, otherwise the created browsing context should be in a
+  new OS window.
+
+ <li><p>Let <var>handle</var> be the
+  associated <a>window handle</a> of the newly created window.
+
+ <li><p>Let <var>type</var> be "<code>tab</code>" if the newly created
+  window shares an OS-level window with the <a>current browsing
+  context</a>, or "<code>window</code>" otherwise.
```
andreastt: We talked about letting remote ends return what makes sense to them, but to _suggest_ returning `tab` or `window`.

christian-bromann: also should this define an error to be thrown if `type` is neither `tab` nor `window`

jgraham: Not throwing an error was intentional so that we can extend the list in the future without breaking compat. (basically new values will always look like `window` in old drivers).

christian-bromann: @jgraham in case of selector strategies we throw an `Invalid Selector` error, shouldn't we stay consistent in that sense?

index.html
```diff
@@ -1313,6 +1313,12 @@ <h3>List of Endpoints</h3>
   <td><a>Get Window Handles</a></td>
  </tr>
 
+ <tr>
+  <td>POST</td>
+  <td>/session/{<var>session id</var>}/window/new</td>
+  <td><a>Create Window</a></td>
```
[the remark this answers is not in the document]
whimboo: As @shs96c added in https://github.com/w3c/webdriver/issues/1138#issuecomment-433340296 we agreed on a `POST` to `/session/{sessionId}/window/handles`. Why do we have a different endpoint here?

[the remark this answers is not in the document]
jgraham: For two reasons:

* It matches the existing Safari implementation, but removes the vendor prefix
* I would expect a `POST` to the `[…]/window/handles` endpoint would take a handle. But in this case you can't pass a handle.

In existing cases where we have `POST` and `GET` endpoints, the `POST` typically overwrites the value that the `GET` would return. The exception is `[…]/cookie`, where the `POST` appends as we do here, but that's clearly misdesigned and should have been `POST` to `[…]/cookie/{name}`.

That said, I don't really care so if there's a consensus that we shouldn't use a different path here that's fine.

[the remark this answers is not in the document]
andreastt: Aligning with what Safari does already makes sense to me, as that is the path of least resistance making this a reality.

index.html
andreastt: @christian-bromann We discussed the need to meet vendor’s future requirements for window types that is something other than `tab` or `window`, e.g. `hololens`, `space`, or `halfspace`. We also talked about how making this infallible is in line with existing commands.

We want to provide _a hint_ to the browser describing what the most desirable type of new top-level browsing context would be, to optimise for scripts that work with minimal product-specific knowledge cross-browser. If the browser can’t open a new top-level browsing context, it obviously makes sense to error with an unsupported operation.

index.html
```diff
@@ -3761,6 +3767,66 @@ <h3><dfn>Get Window Handles</dfn></h3>
...
+ <dl>
+  <dt>"<code>handle</code>"
+  <dd>The value of <var>handle</var>.
+
+  <dt>"<code>type</code>"
+  <dd>The value of <var>type</var>.
+
+ </dl>
+
+ <li><p>Return <a>success</a> with data <a>result</a>.
```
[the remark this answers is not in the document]
jgraham: Not if it only allows opening `about:blank` since that loads synchronously. And I've made it explicit that the new window has to initially contain `about:blank`.

index.html
jgraham: I think it makes more sense to require remote ends to return specific things at the moment; we probably don't want an unconstrained extension space here.

index.html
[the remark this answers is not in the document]
whimboo: Oh, I see. Thanks for the clarification. I assume the synchronous load is equal for all browsers?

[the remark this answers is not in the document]
jgraham: It isn't possible to observe the state where `about:blank` isn't loaded in any browser. They *may* differ on whether there are events relating to the initial `about:blank` document loading or not, but that isn't visible through this API so I'm not concerned about it here.

index.html
```diff
@@ -3794,6 +3801,71 @@ <h3><dfn>Get Window Handles</dfn></h3>
...
+
+<ol>
+ <li><p>If the <a>current top-level browsing context</a> is <a>no longer open</a>,
+  return <a>error</a> with <a>error code</a> <a>no such window</a>.
+
+ <li><p><a>Handle any user prompts</a> and return its value if it is an <a>error</a>.
+
+ <li><p>If the implementation does not support creating new top-level
+  browsing contexts, return <a>error</a> with <a>error code</a>
+  <a>unsupported operation</a>.
```
whimboo: I think it would make sense to move up this step to be the first one. There is no need to check for the top-level browsing context or any user prompt, if the implementation doesn't support this method anyway.

index.html
```diff
@@ -3794,6 +3801,71 @@ <h3><dfn>Get Window Handles</dfn></h3>
...
+
+ <li><p>Let <var>type hint</var> be the result of <a>getting the property</a>
+  "<code>type</code>" from the <var>parameters</var> argument.
```
whimboo: Here I miss a check for invalid arguments, which means if it is not "window" or "tab".

jgraham: It doesn't matter. We aren't going to throw for forward-compat reasons, and the logic below doesn't depend on getting a valid value.

index.html
jgraham: Done.

index.html
```diff
@@ -3794,6 +3801,71 @@ <h3><dfn>Get Window Handles</dfn></h3>
...
+ <li><p>Create a new <a>top-level browsing context</a> by running
+  the <a>window open steps</a> with <var>url</var> set to
+  "<code>about:blank</code>", <var>target</var> set to the empty string,
+  and <var>features</var> set to "<code>noopener</code>" and the user
+  agent configured to create a new browing context. If <var>type
+  hint</var> has the value "<code>tab</code>" and the implementation
+  supports multiple browsing context in the same OS window, the new
+  browsing context should share an OS window with the <a>current
+  browsing context</a>, otherwise the created browsing context should
+  be in a new OS window.
```
whimboo: It's not clear from the content if the tab should open as a foreground tab, or background tab. The same would apply to the window, if it should be opened as front-most, or background window. Given that the context hasn't been switched at this time, the current top-level browsing context would be in a background tab/window, if the new one will be the active one.

Also I think we need to apply s/browsing context/top-level browsing context/g to this paragraph.

jgraham: I think "new browsing context" is fine in context; we've already stated that it's a top-level browsing context and it's not necessary to keep repeating that fact.

I think whether the window/tab is focused is probably an implementation detail? *In theory* it doesn't matter, but I would *expect* that it isn't until you actually switch to it. Since we don't tell you to focus it or anything that matches the simplest reading of the current text.

whimboo: Note that it might affect the currently selected browsing context. Whether it stays in foreground or gets moved into the background could affect website operations. Like mentioned at TPAC eg for Facebook JS could be throttled.

As such it might be good to open the top-level browsing context as background tab, or as a window with a lower z-level as the current selected top-level browsing context's window.

index.html
```diff
@@ -3794,6 +3801,71 @@ <h3><dfn>Get Window Handles</dfn></h3>
...
+  "<code>type</code>" from the <var>parameters</var> argument.
+
+ <li><p>Create a new <a>top-level browsing context</a> by running
+  the <a>window open steps</a> with <var>url</var> set to
+  "<code>about:blank</code>", <var>target</var> set to the empty string,
+  and <var>features</var> set to "<code>noopener</code>" and the user
+  agent configured to create a new browing context. If <var>type
+  hint</var> has the value "<code>tab</code>" and the implementation
+  supports multiple browsing context in the same OS window, the new
+  browsing context should share an OS window with the <a>current
```
whimboo: What I miss here is what should happen when the implementation supports tabs, but not windows (eg. Fennec), and the typ is set to `window`. Should it then default to `tab`? I would say so. The implementation should select the best possible option.

andreastt: That is what the spec text says.

whimboo: No, the text above only covers the case when `tab` is given as type hint, but not `window`. And specifically when the implementation cannot open a window, but could perfectly share an OS window.

Btw. is `type hint` a valid variable we can use here? Or should it just be `type`, whereby that is a reserved keyword in most languages.

jgraham: I fixed the text.

AutomatedTester:
implementation and tests in https://bugzilla.mozilla.org/show_bug.cgi?id=1504756

AutomatedTester:
@jgraham can you rebase this please

AutomatedTester:
@jgraham and when you rebase feel free to get it landed.

PR #1325, opened 2018-10-26
Don't perform interactability checks for send keys to input type=file

jgraham:
This breaks cases where authors have hidden the file input and
interact with it via js. Also add a strictFileInteractabilityCheck
capability to re-enable the previous behaviour of performing the
checks always.


index.html
```diff
@@ -2421,6 +2434,11 @@ <h3>Processing Capabilities</h3>
     indicating the session will not implicitly trust untrusted
     or self-signed TLS certificates on <a data-lt=go>navigation</a>.
 
+   <dt>"<code>strictFileInteractability</code>"
+    <dd><a>Boolean</a> initially set to <code>false</code>,
+    indicating that interactabilty checks will be applied to
+    &lt;input type=file&gt;.
```
AutomatedTester: This should probably  read like 
```suggestion
    an <a><code>input</code> element</a> whose <a><code>type</code> attribute</a> is in the <a>File Upload state</a>
```

index.html
```diff
@@ -2845,6 +2866,14 @@ <h3><dfn data-lt="creating a new session">New Session</dfn></h3>
     property</a> of <var>capabilities</var> with name
     "<code>pageLoadStrategy</code>" and value "<code>normal</code>".
 
+   <li><p>Let <var>strictFileInteractability</var> be the result of getting property
+    "<code>strictFileInteractability</code>" from <var>capabilities</var>.
+
+   <li><p>If <var>strictFileInteractability</var> is a boolean,
+    set the <a>current session</a>’s <a>strict file interactability</a> to
+    <var>strictFileInteractability</var>. Otherwise set the
+    <a>current session</a>'s <var>strict file interactability</a> to <code>false</code>.
```
AutomatedTester: When we are deserializing the capabilities it looks like we error if this is not set as a boolean so we don't need the default though it makes me wonder if we should set a default if nothing comes over the wire earlier than here

index.html
```diff
@@ -6123,22 +6152,29 @@ <h3><dfn>Element Send Keys</dfn></h3>
   of <a>trying</a> to <a>get a known connected element</a>
   with <a>url variable</a> <var>element id</var>.
 
- <li><p><a>Scroll into view</a> the <var>element</var>.
+ <li><p>Let <var>file</var> be <code>true</code> if If <var>element</var> is
```
AutomatedTester: double if

andreastt: We typically don’t put booleans in `<code>` tags. There are some cases we do, but that should be fixed in
https://github.com/w3c/webdriver/issues/1334. This remark applies throughout this patch.

index.html
```diff
@@ -6123,22 +6152,29 @@ <h3><dfn>Element Send Keys</dfn></h3>
...
+   an <a><code>input</code> element</a> whose
+    <a><code>type</code> attribute</a> is <a>File</a>, or <code>false</code> otherwise.
```
andreastt: You should use `<a><code>input</code> element</a> in the <a>file upload state</a>`, and I think you should remove the above _file_ variable.

```suggestion
  an <a><code>input</code> element</a> in the <a>file upload state</a>,
  or false otherwise.
```

index.html
jgraham: Is that clearer? This text is informative, so it doesn't have to be precise.

index.html
jgraham: I think it can be not a boolean if the property isn't set at all in which case `getting a property` will return `undefined`.

index.html
jgraham: I don't understand what you mean about removing the variable, we end up needing to use it in two places, so it seems senisble to keep that logic in a single place.

index.html
andreastt: It still uses `&lt;input type=file&gt;` when it should say `an <a><code>input</code> element</a> in the <a>file upload state</a>`.

index.html
andreastt: Leave the variable if you want.

index.html
jgraham: Right. AFAICT this text is all informative and we could remove it without changing behaviour. In light of that I claim that `<input type=file>` is easier to understand.

andreastt: Oh I see what you mean this is descriptive, resolving this conversation.

index.html
andreastt: That makes sense.

PR #1329, opened 2018-10-28
dependencies: fix tuplicate definition of type

andreastt:
"Type" is defined twice, but this is only referred to as "ECMAScript
type", so we can skip the text content definition.

Fixes the respec error.


AutomatedTester:
Thanks for doing this. It was on my todo for today

PR #1330, opened 2018-10-29
Minor wording fix

christian-bromann:
## Current Wording

In chapter 10. Command Contexts, section 10.3 Switch To Window:

> Return success with date null.

## Expected Wording

Return success with data null.


christian-bromann:
Looks like it is already fixed as part of https://github.com/w3c/webdriver/pull/1324/files#diff-eacf331f0ffc35d4b482f1d15a887d3bR3731

christian-bromann:
Thanks @andreastt

PR #1332, opened 2018-10-29
Element retrieval commands have to invoke the user prompt handler (#1086)

### LINKED ISSUE #1086: Element retrieval commands do not invoke the user prompt handler
I was just looking through this issue: https://github.com/mozilla/geckodriver/issues/920 and I got to wondering why it is that Find Element does not have a step to handle an unexpected dialog the way GetElementText (for example) does:

Handle any user prompts and return its value if it is an error. 

I don't remember discussing this, so I'm not sure if it's an omission by choice or accident...

whimboo:
Fixes #1086.

### DIFF
--- a/index.html
+++ b/index.html
@@ -4913,6 +4913,8 @@ session.execute("arguments[0].remove()", [body]);
   open</a>, return <a>error</a> with <a>error code</a> <a>no such
   window</a>.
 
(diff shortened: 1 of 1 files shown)


whimboo:
Tests are getting added via https://bugzilla.mozilla.org/show_bug.cgi?id=1503015

PR #1339, opened 2018-11-02
Validate spec with ReSpec in CI

andreastt:
I’ve implemented some automation tasks that runs PRs through the ReSpec validator and blocks them from landing if there are any warnings or errors. This will help us prevent cases such as https://github.com/w3c/webdriver/pull/1337.

And it seems to work fine: https://travis-ci.org/andreastt/webdriver/builds/449850440

For it to pass the tests we first need to land https://github.com/w3c/webdriver/pull/1336.

***
<a href="https://pr-preview.s3.amazonaws.com/andreastt/webdriver/pull/1339.html" title="Last updated on Nov 20, 2018, 4:36 PM GMT (41eb82c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1339/e65af93...andreastt:41eb82c.html" title="Last updated on Nov 20, 2018, 4:36 PM GMT (41eb82c)">Diff</a>


Makefile
```diff
@@ -0,0 +1,21 @@
+.DEFAULT_GOAL = test
+.PHONY: test validate respec
+
+VALIDATOR ?= https://validator.w3.org/nu/
+VALIDATOR_OPTS ?= -F laxtype=yes -F parser=html5 -F level=error
+
+curl = curl -sSLf
+respec = respec2html --disable-sandbox -ew
```
christian-bromann: Is `respec2html` a dependency? If so, how does it get installed and can we document it in case someone wants to run it locally?

andreastt: Fixed in https://github.com/w3c/webdriver/pull/1339/commits/44c7faaad841b299283aa5c2ba353e06a7852341.

andreastt:
There appears to be some unrelated problem specifically with the respec command-line linter that needs to be resolved first:

    😱 ReSpec error: Timed-out waiting for highlight: [object HTMLPreElement]

We don’t see this in the [published document](http://w3c.github.io/webdriver/), so I wonder if @marcoscaceres has any idea?

andreastt:
The problem seems to disappear if I mark all `<pre>` blocks with `<pre class=nohighlight>`.

It is arguably more important to run ReSpec tests on incoming PRs than it is to get source code highlighting on examples, so perhaps we should just do that temporarily until the ReSpec people have figured out the problem?

marcoscaceres:
Really sorry, I only just saw that you pinged me. The bug with highlighting should be fixed. It was a bug in Puppeteer.

index.html
```diff
@@ -1022,7 +1022,7 @@ <h3>Errors</h3>
  where <code>1234</code> is not the <a>session id</a> of a <a>current session</a>
  would return an <a>HTTP response</a> with the status 404 and a body of the form:
 
- <pre><code>{
+ <pre class=nohighlight><code>{
```
marcoscaceres: Let try this out... you shouldn’t neeed to do this.

andreastt: Did something change in the Puppeteer dependency?

I dropped the commit adding the `nohighlight` class, and it appears to pass in CI, so thanks!

marcoscaceres: > Did something change in the Puppeteer dependency?

Lastest version broke something in Web Workers... highlighting is done using a web worker.

marcoscaceres:
I got inspired by what you folks built here, so I made a command line tool:

https://github.com/marcoscaceres/respec-validator

It works pretty well locally, but I've hit a bug trying to do authentication with GitHub tokens when running on Travis.

PR #1340, opened 2018-11-03
Fix collection serialization and support collection types found in the wild

andreastt:
WebDriver’s JSON serialisation algorithm doesn’t handle all the collection types handled by implementations found in the wild. These are:

* Arguments
* FileList
* HTMLAllCollection
* HTMLFormControlsCollection
* HTMLOptionsCollection

Whilst adding these I also discovered that the algorithm didn’t really inspect the interior of array-like collections at all, so the PR also fixes that. This is effectively mirroring how geckodriver handles JSON currently. See each commit message for more detailed information, since I also fixed a few other minor bugs along the way.

Tests for this can be found in https://github.com/web-platform-tests/wpt/blob/master/webdriver/tests/execute_script/collections.py.

This was discovered whilst reviewing https://github.com/web-platform-tests/wpt/pull/13880.

### DIFF
--- a/index.html
+++ b/index.html
@@ -6376,28 +6376,13 @@ of the <a>current browsing context</a> <a>active document</a>.
  <dt>type <a>String</a>
  <dd><p><a>Success</a> with data <var>value</var>.
 
- <dt>instance of <a>element</a>
- <dd><p>If the <var>element</var> <a>is stale</a>,
-  return <a>error</a> with <a>error code</a> <a>stale element reference</a>.
-
   <p>Otherwise, return <a>success</a>
    with the <a data-lt="JSON serialization of an element">serialization to JSON</a>
    of the <a>web element</a> <var>value</var>.
 
- <dt>a <a><code>WindowProxy</code></a> object
- <dd><p>If the associated <a>browsing context</a>
-  of the <a><code>WindowProxy</code></a> object
-  in <var>value</var> has been <a>discarded</a>,
-  return <a>error</a> with <a>error code</a> <a>stale element reference</a>.
-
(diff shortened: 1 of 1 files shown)


gsnedders:
This seems a bit nonsensical?

We now have "A collection is an Object that implements the Iterable interface, and whose: […] type is HTMLAllCollection". 

HTMLAllCollection isn't an interface with an iterable declaration, so instances are still not collections.

(I haven't gone through all the types listed, so there might be others which similarly don't have iterable declarations.)

gsnedders:
Also: what's a "type"?

andreastt:
Yes, thanks for pointing those things out. There are some rough edges here. The effect we want to achieve with this PR is that those things we define as “collection types” should be expanded to an array of items, for which we invoke the serialisation algorithm again on each item.

In Marionette we implement this in JS using `[...collection]` because it works on all of the collection types, including HTMLAllCollection despite it not carrying an iterable declaration. I would love some feedback on how you define “expand all these things to arrays, as if calling `[...collection]` in JS”.

HTML uses a narrower definition of what a _collection_ is because it limits them to things deriving from NodeList if I remember correctly. In particular this excludes the Arguments object, but more importantly, there is not a centralised list of collections in HTML or DOM, and this is an attempt an enumerating those, plus adding Arguments into the mix.

“type” probably should be replace with “an instance of”.

jugglinmike:
That's useful context, @ato, because it explains why the new text references the Iterable interface. This pull request was merged, so I'll share some thoughts on this in a new issue: https://github.com/w3c/webdriver/issues/1347

PR #1343, opened 2018-11-03
Clarify conformance chapter

andreastt:
See individual commit messages.

### DIFF
--- a/index.html
+++ b/index.html
@@ -90,15 +90,10 @@ dl.subcategories { margin-left: 2em }
 <section>
 <h2>Conformance</h2>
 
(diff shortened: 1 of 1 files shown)


AutomatedTester:
@sideshowbarker is it ok to drop the RFC2119 wording?

sideshowbarker:
> @sideshowbarker is it ok to drop the RFC2119 wording?

Yeah, it’s OK. There’s no W3C requirement to explicitly state that the spec is using RFC 2119 keywords in their RFC 2119 sense. The HTML spec also doesn’t contain any such statement; instead it just never uses and RFC 2119 keywords non-normatively.

PR #1345, opened 2018-11-03
Reorganise chapters

andreastt:
This is a high-level reorganisation of the document's chapters.

Most notably, all element related commands are moved into a single
chapter ("Elements"), with subchapters for Interactability, Retrieval,
State, and Interaction.

Various other chapter names have been shortened, which gives a
simpler look to the table of contents and way of referencing the
specification.

There are no functional changes apart from editorial fixups in this
patch.

### DIFF
--- a/index.html
+++ b/index.html
@@ -674,7 +674,7 @@ rather than performance, in mind.
 </section> <!-- /Conformance -->
 
 <section>
-<h2>Design Notes</h2>
+<h2>Design</h2>
 
 <i>This section is non-normative</i>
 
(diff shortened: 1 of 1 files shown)


andreastt:
This PR may be a bit controversial (you don’t hear me saying that often!), but it has annoyed me for a long time that commands targetting elements have been scattered around different chapters with little or no overarching idea.

This will be the first in a series of PRs I intend to submit to make maintenance and editing the document easier. There are some obvious things left to do especially with the script execution and actions chapters. There are also some main chapters I think it would be good to shift around, and I think we should get rid of the _Dependencies_ chapter in favour of using the ReSpec citation capability. But all this in good time…

andreastt:
@christian-bromann Thanks for the screenshot comparison! That was a good idea.

I’ve rebased the PR now.

PR #1366, opened 2018-11-20
fix strings throughout

andreastt:
Strings should be formatted with the quotes outside the <code> tag
so that it is clear to the reader the quote marks are not part of
the string itself, like this:

	"<code>foo</code>"

This patch comes along with a few editorial changes, such as the
removal of unnecessary commas in logical groupings, some table
simplification, and indentation changes as a result of breaking up
the text.  Although overall, there were no larger functional changes
made in this patch.

***
<a href="https://pr-preview.s3.amazonaws.com/andreastt/webdriver/pull/1366.html" title="Last updated on Nov 20, 2018, 3:15 PM GMT (a69f73a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1366/6d31f89...andreastt:a69f73a.html" title="Last updated on Nov 20, 2018, 3:15 PM GMT (a69f73a)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4872,9 +4872,10 @@ but provides the best compatibility with existing users.
 </aside>
 
 <p>
-When processing text, <dfn>whitespace</dfn> is defined as
-characters from the Unicode Character Database ([[UAX44]])
-with the <a>Unicode character property</a> <code>"WSpace=Y"</code>.
+When processing text,
+<dfn>whitespace</dfn> is defined as characters from the Unicode Character Database
+with the <a>Unicode character property</a> "<code>WSpace=Y</code>" or "<code>WS</code>".
+[[!UAX44]]
 
 <p>The <a>remote end steps</a> are:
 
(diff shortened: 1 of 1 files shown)


andreastt:
Indentation was only removed where there is only one level. It is maintained where there is need for it.

christian-bromann:
@andreastt what about this:
screen shot 2018-11-20 at 15 55 36

andreastt:
I also had a conversation with @AutomatedTester about this. I went a bit overboard with the changes here, and will in the future apply restructuring changes to a section in a separate commit ahead of the semantic change.

The restructuring changes were in an effort at establishing some degree of [semantic linefeeds](https://rhodesmill.org/brandon/2012/one-sentence-per-line/) in the Actions chapter. This would in theory make subsequent diffs to these sections smaller, but had the opposite effect in this case. I should’ve realised this before I started making the changes.

PR #1404, opened 2019-03-20
Clarify how a response is constructed.

shs96c:
As currently worded, the response's value may be double
JSON-encoded, which is far from ideal. We also fail to
state that the bytes must be UTF-8 encoded.

***
<a href="https://pr-preview.s3.amazonaws.com/shs96c/webdriver-1/pull/1404.html" title="Last updated on Mar 21, 2019, 12:04 AM UTC (c41c371)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1404/4280c91...shs96c:c41c371.html" title="Last updated on Mar 21, 2019, 12:04 AM UTC (c41c371)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -582,9 +582,10 @@ when it receives a particular <a>command</a>.
    <dd>"<code>no-cache</code>"
   </dl>
 
- <li><p>Let <var>response</var>’s <a>body</a> be a JSON <a>Object</a>
-  with a key "<code>value</code>"
-  set to the <a>JSON serialization</a> of <var>data</var>.
+ <li><p>Let <var>response</var>’s <a>body</a> be the UTF-8
+  encoded <a>JSON serialization</a> of a JSON <a>Object</a> with a key
+  "<code>value</code>" set to <var>data</var>. It is valid for the
+  constructed JSON <a>Object</a> to have more keys than this.
 
  <li><p>Let <var>response bytes</var> be the byte sequence resulting
   from serializing <var>response</var> according to the rules in [[!RFC7230]].
--- a/index.html
+++ b/index.html
@@ -582,10 +582,10 @@ when it receives a particular <a>command</a>.
    <dd>"<code>no-cache</code>"
   </dl>
 
- <li><p>Let <var>response</var>’s <a>body</a> be the UTF-8
-  encoded <a>JSON serialization</a> of a JSON <a>Object</a> with a key
-  "<code>value</code>" set to <var>data</var>. It is valid for the
-  constructed JSON <a>Object</a> to have more keys than this.
+ <li><p>Let <var>response</var>’s <a>body</a> be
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -582,9 +582,10 @@ <h3>Processing model</h3>
    <dd>"<code>no-cache</code>"
   </dl>
 
- <li><p>Let <var>response</var>’s <a>body</a> be a JSON <a>Object</a>
-  with a key "<code>value</code>"
-  set to the <a>JSON serialization</a> of <var>data</var>.
+ <li><p>Let <var>response</var>’s <a>body</a> be the UTF-8
+  encoded <a>JSON serialization</a> of a JSON <a>Object</a> with a key
+  "<code>value</code>" set to <var>data</var>. It is valid for the
+  constructed JSON <a>Object</a> to have more keys than this.
```
andreastt: It goes without saying that the object can be returned with more fields than `value`, precisely because there’s no wording to the effect of preventing them. I think if we _need_ to say anything about this, it should go in a note.

Further, arbitrary fields are allowed on nested objects as well, or at least that’s how driver implementations work in practice.

index.html
```diff
@@ -582,9 +582,10 @@ <h3>Processing model</h3>
    <dd>"<code>no-cache</code>"
   </dl>
 
- <li><p>Let <var>response</var>’s <a>body</a> be a JSON <a>Object</a>
-  with a key "<code>value</code>"
-  set to the <a>JSON serialization</a> of <var>data</var>.
+ <li><p>Let <var>response</var>’s <a>body</a> be the UTF-8
+  encoded <a>JSON serialization</a> of a JSON <a>Object</a> with a key
```
andreastt: We should reference [UTF-8 encode](https://encoding.spec.whatwg.org/#utf-8-encode) from INFRA.

index.html
shs96c: I can drop that bit. The people I was talking to were interpreting the spec with the expectation that only the fields mentioned would ever be present on the returned payload, but I agree it should go without saying.

index.html
shs96c: I shall add this.

index.html
```diff
@@ -9360,6 +9360,12 @@ <h2>Index</h2>
    <!-- undefined --> <li><dfn><a href=https://www.ecma-international.org/ecma-262/5.1/#sec-4.3.9>Undefined</a></dfn>
   </ul>
 
+ <dt>Encoding
```
andreastt: Sorry, yes, I obviously meant Encoding 😅

index.html
andreastt: Yes, but then it has come up on multiple occasions, so I think pointing it out somehow would be a great addition. If you want to submit another patch for that I will happily accept it.

Issue #1334, opened 2018-10-30
Inconsistent treatment of identifiers and keywords


jugglinmike:
A few identifiers and keywords are presented with varied visual treatment:

    $ git rev-parse HEAD
    8344d2ff9b26571ad58c73e68b2d62625ce3bfab
    $ grep -Eo '(<[^>]+>)?(true|false|undefined|null)' index.html | sort | uniq -c
         88 <a>null
         37 <a>undefined
          8 <code>false
          1 <code>null
          7 <code>true
          3 <dfn>null
         25 false
          4 null
         26 true
          3 undefined
          4 <var>false
          2 <var>null

Some of the unadorned usages may be correct (e.g. when they are used for their plain-English meaning in informative text like "the UA's behavior is undefined"), but many seem to be unintentional omissions.

I would be happy to submit a patch to correct this, but I could use guidance on which treatment is preferred for each word.

Also, we've [identified](https://github.com/w3c/webdriver/issues/760) and [fixed](https://github.com/w3c/webdriver/pull/771) this once before, so linting for this may be warranted.

PR #1422, opened 2019-05-10
Normalize treatement of `null` keyword

jugglinmike:
Consistently reference the `null` keyword via its definition and using a
`<code>` tag.

This patch was generated programatically using the following command:

    perl -pe 's/(<var>|<a>)+null(<\/var>|<\/a>)+/<a><code>null<\/code><\/a>/g' index.html

***
<a href="https://pr-preview.s3.amazonaws.com/bocoup/webdriver/pull/1422.html" title="Last updated on May 11, 2019, 1:27 AM UTC (56d05c9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1422/b7c6791...bocoup:56d05c9.html" title="Last updated on May 11, 2019, 1:27 AM UTC (56d05c9)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -448,13 +448,13 @@ when it receives a particular <a>command</a>.
    <li><p>Let the <a>current session</a> be the <a>session</a>
     with <a data-lt="session id">ID</a> <var>session id</var>
     in the list of <a>active sessions</a>,
-    or <a>null</a> if there is no such matching <a>session</a>.
+    or <a><code>null</code></a> if there is no such matching <a>session</a>.
 
-   <li><p>If the <a>current session</a> is <a>null</a>
+   <li><p>If the <a>current session</a> is <a><code>null</code></a>
    <a>send an error</a> with <a>error code</a> <a>invalid session id</a>,
    then jump to step 1 in this overall algorithm.
 
-   <li><p>If the <a>current session</a> is not <a>null</a>:
+   <li><p>If the <a>current session</a> is not <a><code>null</code></a>:
 
     <ol>
       <li><p>Enqueue <var>request</var>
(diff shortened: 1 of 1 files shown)


jugglinmike:
This is in service of gh-1334.

index.html
```diff
@@ -6075,7 +6075,7 @@ <h3>Executing Script</h3>
 
 <p>If at any point during the algorithm a <a>user prompt</a> appears,
  abort all subsequent substeps of this algorithm, and return
- <a>Completion</a> { [[\Type]]: <code>normal</code>, [[\Value]]: <var>null</var>, [[\Target]]: <code>empty</code> }.
```
AutomatedTester: This doesn't look like it needed to be changed

index.html
```diff
@@ -6091,7 +6091,7 @@ <h3>Executing Script</h3>
  <li><p>If <var>body</var> is not parsable as a <a>FunctionBody</a>
   or if parsing detects an <a>early error</a>,
   return
-  <a>Completion</a> { [[\Type]]: <code>normal</code>, [[\Value]]: <var>null</var>, [[\Target]]: <code>empty</code> }.
```
AutomatedTester: or here

index.html
jugglinmike: It's my understanding that `<var>` is intended for use with algorithm variables that store values. In this case and the one that follows, `null` refers to the null value, so `<var>` seems inappropriate. Can you say more?

jugglinmike:
> it also looks like you didn't update the definition to be `<a><code>null</code></a>`.

You're right! I've fixed that. The manual edit contradicts the commit message, so I'd like to modify that when/if this patch is approved.

Issue #1439, opened 2019-09-17
Accessibility primitives: computedRole and computedLabel


cookiecrook:
WebDriver should add primitives for computedRole and computedLabel (or some equivalent names)

The engine’s computed accessibility role is exposed to web developers through developer tools (WebKit Inspector, Google DevTools, F5, etc.) but is not available in standard JavaScript contexts. Even extension developers can’t get this information. 

This serves a dual benefit of allowing web authors to perform better Accessibility testing on their sites, and web spec authors to test/compare browser implementations.

cookiecrook:
As an example that cannot be determined in JavaScript is `<div role=“switch checkbox”>` 

Both roles are valid in ARIA 1.0 (checkbox) and ARIA 1.1 (switch) so there is no way from the content side to determine whether the engine supports the newer switch role, or whether it falls back to the 1.0 role.

cookiecrook:
Currently it would be very challenging and possibly undesirable (implementation-wise) to expose these outside the testing context. There is a long-term roadmap in AOM for standardizing the Accessibility tree, but that is out of scope for the near future.

cookiecrook:
@AutomatedTester Requested I fill in a bit more info about what these are. 

**Accessibility Label**

Accessibility Label (sometimes also referred to as Accessible Name) is a short string that labels the function of the control (e.g. the string "Comment" or "Sign In" on a button). Each platform Accessibility API has a mapping for these. For example, iOS has [accessibilityLabel on UIAccessibilityElement](https://developer.apple.com/documentation/objectivec/nsobject/1615181-accessibilitylabel).

The W3C spec that defines how labels should be computed in web rendering engines is [AccName (currently v1.1)](https://www.w3.org/TR/accname-1.1/), which was split out from the ARIA spec a few years ago. The Name and Description computation section of that spec defines rules for what to do when HTML, ARIA, and other technologies combine. For example, the `title` attribution would be sufficient fallback label in some contexts, but a more explicit `aria-label` attractive would override most host language features in HTML, SVG, etc.

Currently the AccName spec is tested manually. Adding a computedLabel property to WebDriver would allow the AccName authors/maintainers to write WPT scripts that could validate the spec implementation in each browser. I expect this new capability would result in bug fixes to each engine, as well as updates/changes to the AccName spec.

cookiecrook:
**Accessibility Role**

[...]

PR #1440, opened 2019-09-18
Introduce "Accessible Name" locator strategy

jugglinmike:
[At TPAC 2019, folks working in ARIA and WebDriver discussed ways to extend WebDriver in order to help web developers learn and enforce best practices in accessibility.](https://www.w3.org/2019/09/16-aria-minutes.html) While discussing the prospect of new commands based on [the APG](https://w3c.github.io/aria-practices/), a more generic feature came up: the ability to select elements according to [accessible name](https://w3c.github.io/accname/).

Today, it's easy for web developers to write UI tests with a "whatever works" mindset. [Without a good amount of experience and forethought](https://bocoup.com/blog/accessibility-for-robots), the CSS selectors they choose to target elements can easily be broken by benign refactoring. In order to keep their tests passing, developers have to perform the additional maintenance work of updating selectors. Not only is this distracting and inefficient, but it's somewhat dangerous. Editing tests while refactoring diminishes developers' ability to verify equivalent functionality and increases the likelihood of regression.

Because an element's accessible name is directly exposed to end users, it is less subject to change than many of the implementation details upon which a CSS selector can rely. Web developers could use a locator strategy based on this to write tests that are more resilient to changes in their application code.

Like many accessibility best practices, web developers don't always use accessible names correctly (or at all). The promise of more resilient tests could help incentivize developers to make their applications more accessible in this way.

There are some concerns about exposing this capability to the DOM generally, both in terms of performance and security. @minorninth and @alice (the principle contributors to [the Accessibility Object Model](https://github.com/WICG/aom)) can probably speak more to that. It's my understanding that integration with WebDriver is much more feasible. At TPAC, @cookiecrook mentioned that this WebKit and/or Safari already make this available in some privileged context.

Could a "accessible name" locator strategy fit in the WebDriver specification? If so, is there some existing algorithm we can reference?

***
<a href="https://pr-preview.s3.amazonaws.com/bocoup/webdriver/pull/1440.html" title="Last updated on Sep 18, 2019, 6:43 PM UTC (087d4ec)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1440/29c5f74...bocoup:087d4ec.html" title="Last updated on Sep 18, 2019, 6:43 PM UTC (087d4ec)">Diff</a>

- - -
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1440)

### DIFF
--- a/index.html
+++ b/index.html
@@ -4267,6 +4267,11 @@ argument <var>reference</var>, run the following steps:
   <td><a>XPath selector</a>
   <td>"<code>xpath</code>"
  </tr>
+
+ <tr>
+  <td><a>Accessible name</a>
+  <td>"<code>accessible name</code>"
+ </tr>
 </table>
 
 <section>
@@ -4421,6 +4426,15 @@ argument <var>reference</var>, run the following steps:
 
 </ol>
 </section> <!-- /XPath -->
+
+<section>
+<h5>Accessible name</h5>
+
+<p>To find a <a>web element</a> with the <dfn>Accessible
+ Name</dfn> <a>strategy</a> return <a>success</a> with data set to ???.
+ <span class="issue">Can the accessibility tree be queried here? Can the result
+ be mapped back to a DOM element?</span>
+</section> <!-- /Accessible name -->
 </section> <!-- /Locator strategies -->
 
 <section>


andreastt:
This seems to be superseded by https://github.com/w3c/webdriver/pull/1444 by @cookiecrook. Is that correct?

andreastt:
@AutomatedTester

cookiecrook:
> This seems to be superseded by #1444 by @cookiecrook. Is that correct?

Works for me.

jugglinmike:
The API proposed in gh-1444 seems optimized for testing browsers, not web applications.

It's definitely important to be able to verify the browser's correctness. However, if we only define commands like "Get Computed Role" and "Get Computed Label," we may miss an opportunity to reach more developers (and to further promote best practices).

With those commands, developers will be empowered to extend their current tests with additional assertions about the accessibility tree. That's far better than the status quo! However, just like with ARIA practices themselves, it's up to the developer to recognize when and where these details are relevant and to add additional code.

In other words, the developer who is most likely to use those APIs is the same developer who is already well-versed in best practices.

If the accessibility primitives were instead exposed through locator strategies, then authors could write their acceptance tests in terms of the accessibility tree. Unlike application development, where ARIA patterns are commonly layered on after-the-fact (sometimes only inconsistently and oftentimes not at all), tests written on top of such APIs would innately require accessibility. That's what I had in mind when I opened this draft.

zcorpan:
Right, this PR would fix issue [Accessibility primitive: get element(s) by accessibility label](https://github.com/w3c/webdriver/issues/1441) which is different from issue [Accessibility primitives: computedRole and computedLabel](https://github.com/w3c/webdriver/issues/1439) (#1444)

PR #1444, opened 2019-09-25
Add support for calculating the accessible label and the accessible role of elements

### LINKED ISSUE #1439: Accessibility primitives: computedRole and computedLabel

AutomatedTester:
…ole of elements

Fixes #1439

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/pull/1444.html" title="Last updated on May 29, 2020, 3:15 PM UTC (580edd3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1444/f66dd0e...580edd3.html" title="Last updated on May 29, 2020, 3:15 PM UTC (580edd3)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -885,6 +885,18 @@ when it receives a particular <a>command</a>.
   <td><a>Is Element Enabled</a></td>
  </tr>
 
+<tr>
+  <td>GET</td>
+  <td>/session/{<var>session id</var>}/element/{<var>element id</var>}/computedrole</td>
+  <td><a>Get Computed Role</a></td>
+</tr>
+
+<tr>
+  <td>GET</td>
+  <td>/session/{<var>session id</var>}/element/{<var>element id</var>}/computedlabel</td>
+  <td><a>Get Computed Label</a></td>
+</tr>
+
  <tr>
   <td>POST</td>
   <td>/session/{<var>session id</var>}/element/{<var>element id</var>}/click</td>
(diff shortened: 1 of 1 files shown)


AutomatedTester:
This is to start getting feedback from accessibility teams. If they are happy I will add wpt

index.html
```diff
@@ -9205,6 +9284,24 @@ <h2>Index</h2>
 -->
 
 <dl>
+ <dt>ARIA
+ <dd>
+   <p>The following terms are defined
+     in the Accessible Rich Internet Applications (WAI-ARIA) 1.2 specification: [wai-aria-1.2]
```
andreastt: These presumably need to be normative references.

index.html
```diff
@@ -9205,6 +9284,24 @@ <h2>Index</h2>
...
+     <ul>
+       <li><dfn><a href="https://w3c.github.io/aria/#introroles">WAI-ARIA role</a></dfn>
+     </ul>
+
+  <dt>Accessible Name and Description Computation
```
andreastt: Generally we tend to list related specs under a single topic, so this be a `<dd>` under `<dt>Aria` from above.

cookiecrook:
> I don’t know ARIA so I’m relying on @cookiecrook’s judgement that these are the right things to expose.
> 
> I have a couple of questions and a few concerns:
> 
> * Are these ARIA concepts that is implementable in all browsers?

Neither of these is specific to ARIA so it would not make sense to tie them to that spec. 

If you wanted to group them, you could possibly call them "accessibility"-specific but the lines between accessibility and mainstream are pretty blurry already. Accessibility as a grouping may make sense because other accessibility concepts (some ARIA-related) may be needed later.

> * Is there vendor commitment to implement these WebDriver endpoints?

Yes for these two. Right now it's the lowest common denominator that every vendor agrees would be useful here. Each major vendor (Blink, WebKit, Gecko) has these exposed in Developer Tools contexts already, so I expect implementation will be trivial.

> * Should this go in th  WebDriver core spec, or should they be WebDriver extensions within ARIA?

Not ARIA specific, so it'd be inappropriate to host it there. 

Also, this came up at the TPAC F2F, and I confirmed they are not otherwise available in the JavaScript context. If I understood @AutomatedTester correctly (or was it another WebDriver WG member?), that means these would have be in the core WebDriver spec.

Every major vendor has these concepts central in each implementation, but there is no way to get at this from an over-the-counter JavaScript context currently... Exposing them in WebDriver would allow us to test implementations against each other.

andreastt:
Very good, thanks! That dispels most of my concerns.

What is missing then are the editorial fixups and tests for this to move forward.

index.html
```diff
@@ -5153,6 +5165,73 @@ <h4><dfn>Is Element Enabled</dfn></h4>
...
+        return <a>error</a> with <a>error code</a> <a>no such window</a>.
+
+    <li><p><a>Handle any user prompts</a>
+        and return its value if it is an <a>error</a>.
+
+    <li><p>Let <var>element</var> be the result
+        of <a>trying</a> to <a>get a known connected element</a>
+        with <a>url variable</a> <var>element id</var>.
+
+    <li><p>Let <var>role</var> be the result of computing the <a>WAI-ARIA role</a>.
```
jgraham, line 5196: Presumably this is `of <var>element</var>`.

AutomatedTester: fixed

jgraham:
Tests are https://github.com/web-platform-tests/wpt/pull/23793

PR #1445, opened 2019-09-26
add support for pinning scripts in remote endpoints

titusfortner:
This is the feature that was discussed at TPAC. What wasn't mentioned was this implementation where scripts are associated with names. The idea is that intermediary nodes (like Sauce) can provide a list of named scripts available to the user without the user needing to pin them at the beginning of their tests. If scripts were only listed with UUIDs generated by the end node, this would not be possible. 

I'm looking forward to hearing back on whether this PR works for how the rest of the group understood the feature request.

There's one spot in here that I'm not sure I've specified correctly, but the intent should be clear, so let me know what needs changing. Thanks!

***
<a href="https://pr-preview.s3.amazonaws.com/titusfortner/webdriver/pull/1445.html" title="Last updated on Oct 16, 2019, 11:59 AM UTC (accf942)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1445/3216584...titusfortner:accf942.html" title="Last updated on Oct 16, 2019, 11:59 AM UTC (accf942)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -6198,13 +6198,17 @@ a <a>remote end</a> must run the following steps:
  </tr>
  <tr>
   <td>POST</td>
-  <td>/session/{<var>session id</var>}/execute/sync/{?<var>name</var>}</td>
+  <td>/session/{<var>session id</var>}/execute/sync</td>
  </tr>
 </table>
 
 <p>The <a>remote end steps</a> are:
 
 <ol>
+ <li><p>Let <var>name</var> be the result of
+  <a>getting a property</a> named <var>name</var>
+  from the <var>parameters</var>.
+
  <li><p>Let <var>body</var> and <var>arguments</var> be the result of
   <a>trying</a> to <a>extract the script arguments from a request</a>
   with argument <var>parameters</var>.
@@ -6254,7 +6258,7 @@ a <a>remote end</a> must run the following steps:
  </tr>
  <tr>
   <td>POST</td>
-  <td>/session/{<var>session id</var>}/execute/async/{?<var>name</var>}</td>
+  <td>/session/{<var>session id</var>}/execute/async</td>
  </tr>
 </table>
 
@@ -6268,6 +6272,10 @@ The first argument provided to the function will be serialized to JSON and retur
 <p>The <a>remote end steps</a> are:
 
 <ol>
+  <li><p>Let <var>name</var> be the result of
+     <a>getting a property</a> named <var>name</var>
+     from the <var>parameters</var>.
+
   <li><p>Let <var>body</var> and <var>arguments</var> by the result of <a>trying</a> to
      <a>extract the script arguments from a request</a> with
      argument <var>parameters</var>.
@@ -6343,18 +6351,25 @@ The first argument provided to the function will be serialized to JSON and retur
   </tr>
   <tr>
    <td>POST</td>
-   <td>/session/{<var>session id</var>}/execute/pin/{<var>name</var>}</td>
+   <td>/session/{<var>session id</var>}/execute/pin</td>
   </tr>
  </table>
 
  <p>The <a>remote end steps</a> are:
 
  <ol>
+  <li><p>Let <var>name</var> be the result of
+  <a>getting a property</a> named <var>name</var>
+  from the <var>parameters</var>.
+
   <li><p>Let <var>script</var> be the result of
  <a>getting a property</a> named <var>script</var>
  from the <var>parameters</var>.
 
    <li><p>If <var>script</var> is <a>undefined</a> or is not a <a>String</a>,
+ return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
+
+  <li><p>If <var>name</var> is <a>undefined</a> or is not a <a>String</a>,
  return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
 
   <li><p>If <var>name</var> matches an <a>executable script name</a>
--- a/index.html
+++ b/index.html
@@ -5919,9 +5919,6 @@ of the <a>current browsing context</a> <a>active document</a>.
  associated with <var>name</var> in the <a>executable script table</a>.
 </ol>
 
-<section>
-<h3><dfn>Executing Script</dfn></h3>
-
 <p>
 A <dfn>collection</dfn> is an <a>Object</a>
 that implements the <a>Iterable</a> interface,
@@ -6190,7 +6187,6 @@ a <a>remote end</a> must run the following steps:
 <p class="note">The above algorithm is not associated
  with any particular element,
  and is therefore not subject to the document CSP <a>directives</a>.
-</section> <!-- /Executing Script -->
 
 <section>
 <h3><dfn>Execute Script</dfn></h3>
--- a/index.html
+++ b/index.html
@@ -1190,7 +1190,8 @@ when it receives a particular <a>command</a>.
   <td><dfn>no such script</dfn>
   <td>404
   <td><code>script timeout</code>
-  <td>No script matching the given script name was found amongst the list of known scripts of the current session.
+  <td>No script matching the given script name was found amongst the
+  <a>list of known scripts</a> of the current session.
  </tr>
 
  <tr>
@@ -5890,7 +5891,7 @@ of the <a>current browsing context</a> <a>active document</a>.
   <a href="#protocol">protocol</a>, between <a>remote</a> and
   <a>local</a> ends.
 
-  Each script has an associated <dfn>executable script name</dfn>
+  Each script may include an optional associated <dfn>executable script name</dfn>
   that uniquely identifies the script within the <a>current session</a>.
 
   Each <a>session</a> maintains a <dfn>list of known scripts</dfn>.
@@ -5908,8 +5909,8 @@ of the <a>current browsing context</a> <a>active document</a>.
   <a>list of known scripts</a>.
 
 
- To <dfn>get a known script</dfn> with
- argument <var>name</var>, run the following steps:
+ To <dfn>get a script</dfn> with argument <var>name</var>,
+ run the following steps:
  <ol>
  <li>If <var>name</var> does not match an <a>executable script name</a>
   in the <a>list of known scripts</a> for the <a>current session</a>,
@@ -6097,14 +6098,14 @@ a <a>remote end</a> must run the following steps:
   <a>getting a property</a> named <code>script</code>
   from the <var>parameters</var>.
 
- <li><p>If <var>script</var> is not a <a>String</a>,
-  return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
-
  <li><p>If <var>script</var> is set and <var>name</var> is set,
  return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
 
+ <li><p>If <var>script</var> is not a <a>String</a>,
+  return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
+
  <li><p>If <var>name</var> is set let <var>script</var> be the result
- of <a>trying</a> to <a>get a known script</a> with argument <var>name</var>
+ of <a>trying</a> to <a>get a script</a> with argument <var>name</var>
 
  <li><p>Let <var>args</var> be the result of
   <a>getting a property</a> named <code>args</code>
@@ -6128,9 +6129,6 @@ a <a>remote end</a> must run the following steps:
  <a>Completion</a> { [[\Type]]: <code>normal</code>, [[\Value]]: <a><code>null</code></a>, [[\Target]]: <code>empty</code> }.
 
 <ol>
- <li><p>Let <var>window</var> be the <a>associated window</a>
-  of the <a>current browsing context</a>’s <a>active document</a>.
-
  <li><p>Let <var>environment settings</var> be
   the <a>environment settings object</a> for <var>window</var>.
 
--- a/index.html
+++ b/index.html
@@ -6129,6 +6129,9 @@ a <a>remote end</a> must run the following steps:
  <a>Completion</a> { [[\Type]]: <code>normal</code>, [[\Value]]: <a><code>null</code></a>, [[\Target]]: <code>empty</code> }.
 
 <ol>
+ <li><p>Let <var>window</var> be the <a>associated window</a>
+  of the <a>current browsing context</a>’s <a>active document</a>.
+
  <li><p>Let <var>environment settings</var> be
   the <a>environment settings object</a> for <var>window</var>.
 
@@ -6357,10 +6360,6 @@ The first argument provided to the function will be serialized to JSON and retur
 
    <li><p>If <var>script</var> is <a>undefined</a> or is not a <a>String</a>,
  return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
-
-   <li><p>If the <a>current browsing context</a> is <a>no longer
- open</a>, return <a>error</a> with <a>error code</a> <a>no such
- window</a>.
 
   <li><p>If <var>name</var> matches an <a>executable script name</a>
  in the <a>list of known scripts</a> for the <a>current session</a>,
--- a/index.html
+++ b/index.html
@@ -5912,7 +5912,7 @@ of the <a>current browsing context</a> <a>active document</a>.
  To <dfn>get a script</dfn> with argument <var>name</var>,
  run the following steps:
  <ol>
- <li>If <var>name</var> does not match an <a>executable script name</a>
+ <li>If <var>name</var> is not equal to an <a>executable script name</a>
   in the <a>list of known scripts</a> for the <a>current session</a>,
   return <a>error</a> with <a>error code</a> <a>no such script</a>.
  <li>Let <var>script</var> be the <a>executable script</a>
--- a/index.html
+++ b/index.html
@@ -1187,6 +1187,13 @@ when it receives a particular <a>command</a>.
  </tr>
 
  <tr>
+  <td><dfn>no such script</dfn>
+  <td>404
+  <td><code>script timeout</code>
+  <td>No script matching the given script name was found amongst the list of known scripts of the current session.
+ </tr>
+
+ <tr>
   <td><dfn>script timeout error</dfn>
   <td>500
   <td><code>script timeout</code>
@@ -2110,6 +2117,10 @@ and <a href=#elements>element retrieval</a>.
 <p>
 A <a>session</a> has an associated <a>user prompt handler</a>.
 Unless stated otherwise it is in the <a>dismiss and notify state</a>.
+
+<p>A <a>session</a> has an associated <a>list of known scripts</a>.
+
+<p>A <a>session</a> has an associated <a>executable script table</a>.
 
 <p>A <a>session</a> has an associated list of <a>active input sources</a>.
 
@@ -5871,7 +5882,44 @@ of the <a>current browsing context</a> <a>active document</a>.
 </section> <!-- /Get Page Source -->
 
 <section>
-<h3>Executing Script</h3>
+ <h3>Scripts</h3>
+
+ <p>
+  An <dfn>executable script</dfn> is an abstraction used to
+  identify a script when it is transported via the
+  <a href="#protocol">protocol</a>, between <a>remote</a> and
+  <a>local</a> ends.
+
+  Each script has an associated <dfn>executable script name</dfn>
+  that uniquely identifies the script within the <a>current session</a>.
+
+  Each <a>session</a> maintains a <dfn>list of known scripts</dfn>.
+  This list is initially empty. When an <a>executable script name</a>
+  is added to the <a>list of known scripts</a>, a corresponding entry
+  is made in the <a>executable script table</a> where the key is the
+  <a>executable script name</a> and the value is the <var>script</var>.
+  When an <a>executable script name</a> is removed from the
+  <a>list of known scripts</a>, the corresponding entry in
+  the <a>executable script table</a> is also removed.
+
+ <p>Each <a>session</a> has an associated <dfn>executable script table</dfn>.
+  This is a map between <a>executable script name</a>
+  and the script, with one entry for each item in the
+  <a>list of known scripts</a>.
+
+
+ To <dfn>get a known script</dfn> with
+ argument <var>name</var>, run the following steps:
+ <ol>
+ <li>If <var>name</var> does not match an <a>executable script name</a>
+  in the <a>list of known scripts</a> for the <a>current session</a>,
+  return <a>error</a> with <a>error code</a> <a>no such script</a>.
+ <li>Let <var>script</var> be the <a>executable script</a>
+ associated with <var>name</var> in the <a>executable script table</a>.
+</ol>
+
+<section>
+<h3><dfn>Executing Script</dfn></h3>
 
 <p>
 A <dfn>collection</dfn> is an <a>Object</a>
@@ -6052,6 +6100,12 @@ a <a>remote end</a> must run the following steps:
  <li><p>If <var>script</var> is not a <a>String</a>,
   return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
 
+ <li><p>If <var>script</var> is set and <var>name</var> is set,
+ return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
+
+ <li><p>If <var>name</var> is set let <var>script</var> be the result
+ of <a>trying</a> to <a>get a known script</a> with argument <var>name</var>
+
  <li><p>Let <var>args</var> be the result of
   <a>getting a property</a> named <code>args</code>
   from the <var>parameters</var>.
@@ -6135,6 +6189,7 @@ a <a>remote end</a> must run the following steps:
 <p class="note">The above algorithm is not associated
  with any particular element,
  and is therefore not subject to the document CSP <a>directives</a>.
+</section> <!-- /Executing Script -->
 
 <section>
 <h3><dfn>Execute Script</dfn></h3>
@@ -6146,7 +6201,7 @@ a <a>remote end</a> must run the following steps:
  </tr>
  <tr>
   <td>POST</td>
-  <td>/session/{<var>session id</var>}/execute/sync</td>
+  <td>/session/{<var>session id</var>}/execute/sync/{?<var>name</var>}</td>
  </tr>
 </table>
 
@@ -6202,7 +6257,7 @@ a <a>remote end</a> must run the following steps:
  </tr>
  <tr>
   <td>POST</td>
-  <td>/session/{<var>session id</var>}/execute/async</td>
+  <td>/session/{<var>session id</var>}/execute/async/{?<var>name</var>}</td>
  </tr>
 </table>
 
@@ -6281,7 +6336,43 @@ The first argument provided to the function will be serialized to JSON and retur
       and data <var>result</var>.
 </ol>
 </section> <!-- /Execute Async Script -->
-</section> <!-- /Executing Script -->
+<section>
+ <h3><dfn>Pin Script</dfn></h3>
+
+ <table class="simple jsoncommand">
+  <tr>
+   <th>HTTP Method</th>
+   <th>URI Template</th>
+  </tr>
+  <tr>
+   <td>POST</td>
+   <td>/session/{<var>session id</var>}/execute/pin/{<var>name</var>}</td>
+  </tr>
+ </table>
+
+ <p>The <a>remote end steps</a> are:
+
+ <ol>
+  <li><p>Let <var>script</var> be the result of
+ <a>getting a property</a> named <var>script</var>
+ from the <var>parameters</var>.
+
+   <li><p>If <var>script</var> is <a>undefined</a> or is not a <a>String</a>,
+ return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
+
+   <li><p>If the <a>current browsing context</a> is <a>no longer
+ open</a>, return <a>error</a> with <a>error code</a> <a>no such
+ window</a>.
+
+  <li><p>If <var>name</var> matches an <a>executable script name</a>
+ in the <a>list of known scripts</a> for the <a>current session</a>,
+ remove the script <var>name</var> from the <a>list of known scripts</a>.
+
+ <li><p>Add the script <var>name</var> to the <a>list of known scripts</a>
+ for the <a>current session</a>.
+</ol>
+</section> <!-- /Pin Script -->
+</section> <!-- /Scripts -->
 </section> <!-- /Document -->
 
 
--- a/index.html
+++ b/index.html
@@ -2119,9 +2119,7 @@ and <a href=#elements>element retrieval</a>.
 A <a>session</a> has an associated <a>user prompt handler</a>.
 Unless stated otherwise it is in the <a>dismiss and notify state</a>.
 
-<p>A <a>session</a> has an associated <a>list of known scripts</a>.
-
-<p>A <a>session</a> has an associated <a>executable script table</a>.
+<p>A <a>session</a> has an associated <a>executable script map</a>.
 
 <p>A <a>session</a> has an associated list of <a>active input sources</a>.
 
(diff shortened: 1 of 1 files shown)


titusfortner:
I've linked my github account to my w3c account, not sure how to revalidate the failing check.

index.html
```diff
@@ -5871,7 +5882,44 @@ <h3><dfn>Get Page Source</dfn></h3>
...
+  <a>list of known scripts</a>, the corresponding entry in
+  the <a>executable script table</a> is also removed.
+
+ <p>Each <a>session</a> has an associated <dfn>executable script table</dfn>.
+  This is a map between <a>executable script name</a>
+  and the script, with one entry for each item in the
+  <a>list of known scripts</a>.
+
+
+ To <dfn>get a known script</dfn> with
```
AutomatedTester:
```suggestion
 To <dfn>get a script</dfn> with
```

index.html
```diff
@@ -6052,6 +6100,12 @@ <h3>Executing Script</h3>
  <li><p>If <var>script</var> is not a <a>String</a>,
   return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
 
+ <li><p>If <var>script</var> is set and <var>name</var> is set,
+ return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
+
+ <li><p>If <var>name</var> is set let <var>script</var> be the result
+ of <a>trying</a> to <a>get a known script</a> with argument <var>name</var>
```
AutomatedTester:
```suggestion
 of <a>trying</a> to <a>get a script</a> with argument <var>name</var>
```

index.html
```diff
@@ -6281,7 +6336,43 @@ <h3><dfn>Execute Async Script</dfn></h3>
...
+ <table class="simple jsoncommand">
+  <tr>
+   <th>HTTP Method</th>
+   <th>URI Template</th>
+  </tr>
+  <tr>
+   <td>POST</td>
+   <td>/session/{<var>session id</var>}/execute/pin/{<var>name</var>}</td>
+  </tr>
+ </table>
```
AutomatedTester: Is it worth having a getter here? We don't have a mechanism for executing code currently the only way to know a script is there is one we have placed ourselves. What happens if a user wants to use one that a Selenium SaaS sets? Do we assume they will have access to those and need to code that from the scratch?

jgraham:
Is there a corresponding wpt PR adding tests for the new feature?

titusfortner:
@jgraham  https://github.com/web-platform-tests/wpt/pull/19675

index.html
titusfortner: Not sure the value of limiting size, do browser vendors want it?

A getter might be a good idea. I'm not worried about getting a list from SaaS, that should be in documentation before the user writes the code, not something they need an open session to evaluate. A better use case would be the ability to:
```
pin_script(name, script) unless pinned_scripts.include?(name)
```

In which case maybe we would want to throw an exception for duplicated script names rather than just overwriting?

andreastt:
I have two questions:

1. We want to add extension locator strategies to WebDriver, and one way to do that would be to register a chunk of JS with the driver and associate it with a custom locator name. Such an approach would share many characteristics with script pinning, and thus introduce a dependency here.

   My preference would be _not to share_ the same infrastructure for extension locators as for script pinning because I think there are better ways (pre-registration through capabilities for example), but I wanted to make sure to raise the question that we all feel that way, and that extension selectors should not be implemented with pinned scripts?

2. Is there a use case to replace pinned scripts at runtime? If that is not a strong requirement, a simpler implementation would be to provide pinned scripts at session creation.

jgraham:
> I have two questions:
> 
>     1. We want to add extension locator strategies to WebDriver, and one way to do that would be to register a chunk of JS with the driver and associate it with a custom locator name. Such an approach would share many characteristics with script pinning, and thus introduce a dependency here.
>        My preference would be _not to share_ the same infrastructure for extension locators as for script pinning because I think there are better ways (pre-registration through capabilities for example), but I wanted to make sure to raise the question that we all feel that way, and that extension selectors should not be implemented with pinned scripts?

I think they should share the same infrastructure; they are basically the same feature but with the difference that the locator strategies must return either an element or an array of elements. I think that can work well with the approach in this PR.

>     2. Is there a use case to replace pinned scripts at runtime? If that is not a strong requirement, a simpler implementation would be to provide pinned scripts at session creation.

Only allowing pinned scripts to be uploaded at session creation seems interesting but I wonder if it's flexible enough. At session creation time you might not know that you're running e.g. react app tests which need a particular set of tests. With that in mind I wonder if it should be possible to remove one or all pinned scripts so you could pin some scripts in a fixture for a set of tests and unpin those scripts in the teardown for the tests without requiring the overhead of getting a new session between test modules.

index.html
andreastt: If we can’t see a use case for getting the script, let’s avoid the complexity altogether.

I do see value in a driver refusing to pin a script for whatever reason deemed necessary.

index.html
```diff
@@ -6281,7 +6336,43 @@ <h3><dfn>Execute Async Script</dfn></h3>
...
+  <li><p>Let <var>script</var> be the result of
+ <a>getting a property</a> named <var>script</var>
+ from the <var>parameters</var>.
+
+   <li><p>If <var>script</var> is <a>undefined</a> or is not a <a>String</a>,
+ return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
+
+   <li><p>If the <a>current browsing context</a> is <a>no longer
+ open</a>, return <a>error</a> with <a>error code</a> <a>no such
+ window</a>.
```
andreastt: I think we should drop this requirement for pinning scripts as it can be done globally.

index.html
```diff
@@ -6281,7 +6336,43 @@ <h3><dfn>Execute Async Script</dfn></h3>
...
+
+  <li><p>If <var>name</var> matches an <a>executable script name</a>
+ in the <a>list of known scripts</a> for the <a>current session</a>,
+ remove the script <var>name</var> from the <a>list of known scripts</a>.
+
+ <li><p>Add the script <var>name</var> to the <a>list of known scripts</a>
+ for the <a>current session</a>.
```
andreastt: This is fine, but not how you would implement this in actual code.

You would have a mutable `HashMap<String, String>` where the script value for the key _executable script name_ gets replaced. See also my above question about the duplication of the data structure.

index.html
```diff
@@ -6281,7 +6336,43 @@ <h3><dfn>Execute Async Script</dfn></h3>
...
+ <h3><dfn>Pin Script</dfn></h3>
+
+ <table class="simple jsoncommand">
+  <tr>
+   <th>HTTP Method</th>
+   <th>URI Template</th>
+  </tr>
+  <tr>
+   <td>POST</td>
+   <td>/session/{<var>session id</var>}/execute/pin/{<var>name</var>}</td>
```
andreastt: Is it just me, or is it a bit weird to pick one piece of data from the query parameters and the other from the request body?

andreastt:
> Only allowing pinned scripts to be uploaded at session creation seems interesting but I wonder if it's flexible enough. At session creation time you might not know that you're running e.g. react app tests which need a particular set of tests. With that in mind I wonder if it should be possible to remove one or all pinned scripts so you could pin some scripts in a fixture for a set of tests and unpin those scripts in the teardown for the tests without requiring the overhead of getting a new session between test modules.

This sounds reasonable to me.

I wonder if this means having an endpoint to _clear_ the pinned scripts might be necessary. The way this proposal works the client would need to hold on to a reference to the pinned script and I wonder if this might be a problem for long-running sessions.

titusfortner:
@twalpole this was actually your idea that I stole, chime in with your opinions, please.

index.html
```diff
@@ -5871,7 +5882,44 @@ <h3><dfn>Get Page Source</dfn></h3>
...
+  is made in the <a>executable script table</a> where the key is the
+  <a>executable script name</a> and the value is the <var>script</var>.
+  When an <a>executable script name</a> is removed from the
+  <a>list of known scripts</a>, the corresponding entry in
+  the <a>executable script table</a> is also removed.
+
+ <p>Each <a>session</a> has an associated <dfn>executable script table</dfn>.
+  This is a map between <a>executable script name</a>
+  and the script, with one entry for each item in the
+  <a>list of known scripts</a>.
```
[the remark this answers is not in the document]
titusfortner: I liberally copied from other sections. This wording was lifted from the Actions class section. It seems to make sense to me that we would both want to iterate over a list and be able to do lookups. If we can reference iterating over a list of keys from a table, that would work as well.

index.html
```diff
@@ -5871,7 +5882,44 @@ <h3><dfn>Get Page Source</dfn></h3>
...
+
+
+ To <dfn>get a known script</dfn> with
+ argument <var>name</var>, run the following steps:
+ <ol>
+ <li>If <var>name</var> does not match an <a>executable script name</a>
```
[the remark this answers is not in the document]
titusfortner: I copied that from the actions section as well, which is also a String value:
> If source’s source type does not match type return an error with error code invalid argument.

index.html
```diff
@@ -6135,6 +6189,7 @@ <h3>Executing Script</h3>
 <p class="note">The above algorithm is not associated
  with any particular element,
  and is therefore not subject to the document CSP <a>directives</a>.
+</section> <!-- /Executing Script -->
```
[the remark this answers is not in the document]
titusfortner: I added a new section "Script" as the top level with Executing Script, Execute Script, Execute Async Script, and Pin Script as the subsections. Let me know if this is not what we want.

index.html
titusfortner: I was modeling this after getting attribute value. We do that from the endpoint and not the request body.

index.html
titusfortner: Are we discussing getting a list of names or getting the actual script? I was thinking the name list, just wanted to make sure we're discussing the same thing.

index.html
andreastt: Getting an element attribute is conceptually different because it is idempotent. In this case you are modifying state, and the _name_ field in the URL template is a disguised attempt simulating query parameters (`?foo=bar`) in a POST request.

In RESTful APIs the model is usually that the POST request returns an identifier you can later use in a GET request. If we hypothetically imagine we wanted a getter for the pinned script, `GET /session/{session id}/execute/pin/{name}` would be acceptable.

I also have two other concerns:

1. Script names that include “special” characters that are not permitted in URLs must be encoded, for example `foo bar baz æøå` becomes `foo%20bar%20baz%20%C3%A6%C3%B8%C3%A5`. This is less legible and transparent than `{"name": "foo bar baz æøå", "script": …}`.

2. If we want to have a command clearing the pinned scripts the most logical endpoint would be `POST /session/{session id}/execute/pin/clear`. It also blocks future extensions of the API, as related commands would need their own top-level endpoint, e.g. `/session/{session id}/execute/specialisedPin`.

index.html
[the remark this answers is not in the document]
andreastt: I had a look at [the preview rendering of this PR](https://pr-preview.s3.amazonaws.com/titusfortner/webdriver/pull/1445.html) and it looks like this:

* Document
** Get Page Source
** Scripts
*** Executing Script
*** Execute Script
*** Execute Async Script
*** Pin Script

The introduction of the _Scripts_ section actually looks fine, but the _Executing Script_ sub-sub section seems redundant.

index.html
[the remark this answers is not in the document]
andreastt: That must be a bug, because “matching” is not defined for strings.

index.html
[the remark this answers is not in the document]
andreastt: I had a look at [the preview rendering of this PR](https://pr-preview.s3.amazonaws.com/titusfortner/webdriver/pull/1445.html) and it looks like this:

```
* Document
** Get Page Source
** Scripts
*** Executing Script
*** Execute Script
*** Execute Async Script
*** Pin Script
```

The introduction of the _Scripts_ section actually looks fine, but the _Executing Script_ sub-sub section seems redundant.

index.html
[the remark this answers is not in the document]
andreastt: [Infra](https://infra.spec.whatwg.org) appears to support (maps)[https://infra.spec.whatwg.org/#maps), so let’s use them.

They more accurately reflect how most implementations would be written (unless you’re writing one in C) and helps clarity of the prose.

index.html
jgraham: FWIW I think `POST`ing some data to a URL containing a user-chosen unique id is a reasonable pattern. For deleting it seems reasonable that sending `DELETE` to `/session/{session id}/execute/pin` would work (and that deleting a specific script would be  `DELETE` `/session/{session id}/execute/pin/{name}`).

However I don't want this to be seen as a strong opinion in either direction; the exact details of the HTTP endpoints are not something I particularly care about.

index.html
andreastt: Sorry, I thought you meant getting the full body of the scripts earlier.

Having an endpoint to get _a list_ of the pinned scripts’ names does seems reasonable to me, for the same reasons I mentioned in https://github.com/w3c/webdriver/pull/1445#issuecomment-541749149.

If there is no way of knowing which scripts are pinned in the driver, the local end must maintain these references locally. If it for whatever reason loses these references this could lead to a build-up of unused pinned scripts.

So far my thinking is we need:

- An endpoint for getting a list of the pinned scripts’ names.
- An endpoint for deleting individual pinned scripts by name.
- Optionally a shorthand method for deleting all pinned scripts.

What do you think?

index.html
andreastt: I put a proposal in https://github.com/w3c/webdriver/pull/1445#discussion_r334833023 for addressing pinned script deletion. I like the idea of `DELETE /session/{session id}/execute/pin/{name}` deleting a named pinned script and `DELETE /session/{session id}/execute/pin` clearing all. This is similar to how cookies are modelled.

I’m still not convinced using a URI template field name for the POST request is a good idea.

titusfortner:
@andreastt 
> 1. We want to add extension locator strategies to WebDriver, and one way to do that would be to register a chunk of JS with the driver and associate it with a custom locator name. Such an approach would share many characteristics with script pinning, and thus introduce a dependency here.

I don't think we need something different for element location, because we already handle the use case of `execute/sync` endpoint returning an Element ID value. 

All the client bindings need to be able to do is to allow the creation of custom locators. Like in Ruby I'd do something like:

```
# In Selenium bindings
Selenium::WebDriver::SearchContext::CUSTOM_FINDERS = []

def find_element(how, what)
  if FINDERS[how.to_sym]
    by = FINDERS[how.to_sym] 
    bridge.find_element_by by, what.to_s, ref
  elsif CUSTOM_FINDERS[how.to_sym]
    custom_proc = CUSTOM_FINDERS[how.to_sym]
    custom_proc.call(bridge, what)
   end
end
```
```
# In framework code (or user's test code)
def self.register_react_locator!
  driver.pin_script(:react_location, react_locate_script)
  Selenium::WebDriver::CUSTOM_FINDERS[:react] = Proc.new |context, args| { 
  context.execute_script(:react_location, *args)}
end
```
```
# In user's test code
MyTestFramework.register_react_locator!
driver.find_element(react_locate: 'MyComponent')
```

> 2. Is there a use case to replace pinned scripts at runtime? If that is not a strong requirement, a simpler implementation would be to provide pinned scripts at session creation.

Are you suggesting that we should only pin scripts at session creation? I'd like more flexibility than that.

index.html
```diff
@@ -5871,7 +5882,44 @@ <h3><dfn>Get Page Source</dfn></h3>
...
 
 <section>
-<h3>Executing Script</h3>
+ <h3>Scripts</h3>
+
+ <p>
+  An <dfn>executable script</dfn> is an abstraction used to
+  identify a script when it is transported via the
+  <a href="#protocol">protocol</a>, between <a>remote</a> and
+  <a>local</a> ends.
```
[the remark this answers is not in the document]
titusfortner: yeah, my thought was to do what "Elements" section does, and try to define "Script" in the context of its usage in the main section, then have subsections that talk about what you do with them.

index.html
```diff
@@ -5871,7 +5882,44 @@ <h3><dfn>Get Page Source</dfn></h3>
...
+
+  Each script has an associated <dfn>executable script name</dfn>
+  that uniquely identifies the script within the <a>current session</a>.
```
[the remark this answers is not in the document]
titusfortner: Right, this makes sense, let me think through the rewrite.

index.html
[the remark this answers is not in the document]
titusfortner: yup, this part was copied from Element definition

index.html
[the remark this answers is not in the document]
titusfortner: So, I was trying to get across the idea that the script itself is the same whether it is pinned or not. Rather than defining a separate "pinned script," it's just a matter of how you obtain it (either from pinned or from the parameter value). I'll change this to `Each script may include an optional associated executable script name` for now and we can address the bigger organizational consideration after I tidy things.

index.html
```diff
@@ -6052,6 +6100,12 @@ <h3>Executing Script</h3>
  <li><p>If <var>script</var> is not a <a>String</a>,
   return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
 
+ <li><p>If <var>script</var> is set and <var>name</var> is set,
+ return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
```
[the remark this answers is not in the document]
titusfortner: does it matter if they are both returning an invalid argument? I'll swap the order; let me know if that's not what you are suggesting.

index.html
titusfortner: the way I'm looking at it, a pinned script represents a state of a script object not its own object. In which case I would think David's suggestion is the correct fix.

index.html
titusfortner: I modeled this after the wording in the Actions section where there was both a list and a table where the item in the list was the key value, isn't it essentially the same thing? How is that part implemented?

index.html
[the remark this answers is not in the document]
andreastt: I guess it does not matter because the difference is not observable.

index.html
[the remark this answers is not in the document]
titusfortner: Ok, I fixed it and made #1451

index.html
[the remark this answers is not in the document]
andreastt: How can a script be uniquely identified if it does not have an associated named?

index.html
andreastt: I touch on this in https://github.com/w3c/webdriver/pull/1445#discussion_r334827587.

The effect is the same, but implementing a C-style hash map just seems a bit tedious and error prone.

index.html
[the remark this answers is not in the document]
titusfortner: Ok, I removed it as an extra section. All of the `Executing Script` info is inside `Scripts`

titusfortner:
> In my original comment I was actually suggesting we might need something different for custom locator strategies.

> it would be more flexible and less dependent on a particular client implementation if the element retrieval commands would accept custom strategies

Yes, I agree the bindings need this. I think @shs96c said he had ideas related to this, but I don't know what he was considering. I like the lambda / closure approach I outlined above (which I think does what you are suggesting @andreastt), but I'm not sure how "Ruby" of an approach that is relative to what other languages can do.

index.html
titusfortner: Oh, I agree with Andreas here, I think being able to namespace the script name with a colon is important, which can't be done as an endpoint. I'll change it.

index.html
titusfortner: Based on comment elsewhere (that I missed when I posted this), I'll take a look at https://infra.spec.whatwg.org/#maps but I am still curious how it is different from Actions.

index.html
titusfortner: TL/DR: Is it ok if the initial implementation does not have these extra commands and we add it if people request the feature? I think if we encourage SaaS & Frameworks to namespace their scripts, these shouldn't be necessary.

If we keep current PR approach of merely overwriting name collisions, then delete is definitely not necessary, and even list might not be necessary. Overwriting the script is the most forgiving thing, and I see a lot of accidental duplicated wire calls that users send to Sauce. Even though pinning a command multiple times would be inefficient, I'm not sure it makes sense for the spec to prevent something just to protect a user from being inefficient.

But even if we do change it to return an error, the use case for delete & list would only be if the user were not in complete control over their own code (e.g. the initialization code is controlled by another team and they force everyone to use a specific implementation of "something" that another team doesn't want to use). I just don't see this happening.

I'm trying to think through scenarios for this feature. And realize vast majority of people probably won't need or use this, and if they do decide to use this to tweak performance, they are more likely to figure out the better ways to do it.

1. Allow SaaS/Grid to handle it, so no pinning is necessary, delete isn't necessary, and list shouldn't be necessary because they'd be writing code based off of provided documentation or SaaS API endpoint, not trying to figure out what is available from the driver in the middle of a session. Hmm, maybe if Grid implements atoms with this, then it would make sense to get a list because our documentation... lags
```
driver.execute_script('sauce:visible', element)
driver.execute_script('se:visible', element)
```

2. Allow open source test library (or user framework code) to wrap it and manage state so it's completely encapsulated from user
```
class Scripts
def pin(script_name)
  js = file.read File.expand_path("../scripts/#{script_name}.js", __FILE__)
  driver.pin_script("watir:#{script_name}", js)
  pinned_scripts << script_name
end
end

class Element
def visible?
  scripts.pin(:visible) unless scripts.pinned_scripts.include?(:visible)
  execute_script('watir:visible', element)
end
end
```

3. Have user/framework pin ALLTHETHINGS up front, regardless of whether they'll use them. (potentially ouch)

4. Have the user manage things as needed in their code. This is the only time I can think of when you'd need to be able to get a list, and again I'm not sure why you'd want to delete things.
```
driver.pin_script('visible', script) unless driver.pinned_scripts.include?('visible')
```

If the whole point of this feature is to reduce latency and bandwidth for remote interactions, this is going to add an extra wire call every time, which isn't ideal, but maybe would provide the flexibility for teams that can't figure out how to do it in a better way, but again a user is only going to do this themselves if they figure out it improves performance, otherwise they'll stop.

index.html
titusfortner: > I do see value in a driver refusing to pin a script for whatever reason deemed necessary

Is this the same as the overwrite/error discussion, or is this just a failsafe for a driver to punt on things if it wants?

titusfortner:
So to summarize the various things that are still outstanding in the comments above.

I'm going to:
1. Change the names to not be part of the endpoints so that we can use colons
2. Change the data structure to use `Map` instead of `list` & `table`

Outstanding questions/decisions:
1. Do we want list / delete / delete all endpoints (my vote is no)
2. Do references to "script" need to be changed to "pinned script" for clarity? (I don't think so, but let me know)

Are there other outstanding issues to be addressed from the above that I'm glossing over?

index.html
titusfortner: I think this is all I need to do to fix this. I think I already addressed the things that were overcomplicated the last time I tried this approach... 

https://github.com/w3c/webdriver/pull/1445/commits/4a79a5dbe0ebc079915320005cf3c4b769fb2650

twalpole:
@titusfortner After a quick perusal of the comments here, for my use case I have no need for getting a list of the names, or for deleting scripts. Being able to overwrite an existing script would probably be useful though.  I definitely wouldn't want a limit on the script size since the whole point of the feature is to replace large script transfers with much smaller calls.  I'll try and read everything in more depth later today and may have some more useful comments.

index.html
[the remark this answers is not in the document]
titusfortner: It only needs a name if it is pinned, otherwise it is just processed from the provided arguments

index.html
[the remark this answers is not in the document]
titusfortner: addressed in https://github.com/w3c/webdriver/pull/1445#issuecomment-542264084

index.html
titusfortner: addressed in https://github.com/w3c/webdriver/pull/1445#issuecomment-542264084

index.html
andreastt: We are trying to make a wholesome API that doesn’t just cater to SaaS users of WebDriver. I think it would be an unfortunate piece of design if WebDriver didn’t allow the pinned scripts to be queried.

An intermediary node (SaaS) or a local end (client) might know which scripts it has already pinned and take whatever action it feels compelled to make to prevent/allow overwriting scripts, but these are fundamentally not concerns of the specification. The specification goes to great length to _not specify_ the concerns of the local end, meaning they are free to implement the API as they wish. But it is of course useful and indeed necessary to take into consideration the use cases local ends have.

The problem I have with a setter-only approach is the assumption that local ends are always stateful, or that they in all cases have the correct picture of the remote end’s state. Let me quickly give two examples:

1. A stateless WebDriver client might not have the ability to store states across invocations if all it knows about on the local system is the WebDriver session ID. This means unless there’s an API that allows querying the list of pinned scripts, it is impossible for it know if it’s modifying/overwriting an existing script.

2. Even with a stateful client that locally maintains references to the pinned scripts, nothing prevents an intermediary node or another client from making additional requests without the original client’s knowledge. In other words, accumulating and trusting state built up on the local end is not _atomic_ in the sense that other endpoints might’ve manipulated the pinned scripts.

Because of the particular nature of the script pinning API and the way it accumulates data in memory, I think it is important to have primitives that allow the pinned scripts to be queried and cleared. My proposal would be to have:

* POST `/session/{session id}/execute/pin` with a JSON body `{name: "…", script: "…"}` that would overwrite any script by the same name. (Optionally you might consider returning `{overwritten: true/false}`, but I don’t particularly see a use case for it and it could always be introduced later.)
* GET `/session/{session id}/execute/pin` returning a JSON Array of scripts by their names, i.e. `["One", "Two", …]`.
* DELETE `/session/{session id}/execute/pin/{?name}` that deletes a single script by name if `name` is given, or all pinned scripts.

This mirrors accurately how you might design other RESTful APIs for your web app or whatever.

Of course we can contain this PR to the first API and follow up on the others later, but I feel it is important to have a clear picture in mind of where we are heading.

index.html
andreastt: > > I do see value in a driver refusing to pin a script for whatever reason deemed necessary
> 
> Is this the same as the overwrite/error discussion, or is this just a failsafe for a driver to punt on things if it wants?

Yes, this is specifically about the fact that something might happen on the driver side that might prevent it from storing the script. An example would be the driver running out of memory.

I thought we were all in agreement that another script pinned under the same name should overwrite the last one. At least I don’t think _I_ suggested that.

index.html
```diff
@@ -2111,6 +2119,8 @@ <h2>Sessions</h2>
 A <a>session</a> has an associated <a>user prompt handler</a>.
 Unless stated otherwise it is in the <a>dismiss and notify state</a>.
 
+<p>A <a>session</a> has an associated <a>executable script map</a>.
```
andreastt: This needs to be something like:

> A _session_ has an associated _map_ of _executable scripts_.

Then you need to add [map](https://infra.spec.whatwg.org/#maps) to the [Index](https://w3c.github.io/webdriver/#index) under the _Infra_ entry.

index.html
```diff
@@ -5871,7 +5881,38 @@ <h3><dfn>Get Page Source</dfn></h3>
...
+  <a href="#protocol">protocol</a>, between <a>remote</a> and
+  <a>local</a> ends.
+
+  Each script may include an optional associated <dfn>executable script name</dfn>
+  that uniquely identifies the script within the <a>current session</a>.
+
+  Each <a>session</a> maintains an <a>ordered map</a> <dfn>executable script map</dfn>,
+  with each entry having a key of an <a>executable script name</a>
+  and a value of an <a>exeuctable script</a>.
+  This map is initially empty.
```
andreastt: This is already defined above under :2122.

Could you drop this paragraph and move the entry descriptions up?

(I think probably it would make more sense fo the various bit of session state to live closer to where they are used, but for the moment we should put it upstairs in the session chapter.)

index.html
```diff
@@ -5871,7 +5881,38 @@ <h3><dfn>Get Page Source</dfn></h3>
...
+
+  To add an <a>executable script</a> to the session,
+  for <a>executable script map</a>, key <a>executable script name</a>, and value <var>script</var>,
+  set map[key] to value.
```
andreastt: This is not a normative definition because it is never called anywhere. We only ever add something to this map when under the _Pin Script_ command, so it would be sufficient to make this entire paragraph a step under this command.

index.html
```diff
@@ -5871,7 +5881,38 @@ <h3><dfn>Get Page Source</dfn></h3>
...
+
+  To remove an <a>executable script</a> from the session,
+  for <a>executable script map</a>, key <a>executable script name</a>, and value <var>script</var>,
+  remove map[key].
```
andreastt: Similarly.

index.html
```diff
@@ -5871,7 +5881,38 @@ <h3><dfn>Get Page Source</dfn></h3>
...
+
+  To <dfn>get a script</dfn> with argument <var>name</var>, run the following steps:
```
andreastt: Missing `<p>`.

index.html
```diff
@@ -5871,7 +5881,38 @@ <h3><dfn>Get Page Source</dfn></h3>
...
+ <ol>
+  <li>If <var>name</var> is not equal to a key in the <a>executable script map</a>
+   for the <a>current session</a>, return <a>error</a> with <a>error code</a>
+   <a>no such script</a>.
+  <li>Let <var>script</var> be the value in the <a>executable script map</a>
+   associated with the key <var>name</var>.
```
andreastt: This should probably be:

```
<li>Return <a>executable script map</a>[<var>name</var>].
```

Or in a slightly longer form:

```
<li>Return the restult of <a>getting the value of an entry</a> on the <a>executable script map</a> given <var>name</var>.
```

index.html
```diff
@@ -6049,9 +6090,15 @@ <h3>Executing Script</h3>
...
   from the <var>parameters</var>.
 
+ <li><p>If <var>script</var> is set and <var>name</var> is set,
+ return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
+
  <li><p>If <var>script</var> is not a <a>String</a>,
   return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
 
+ <li><p>If <var>name</var> is set let <var>script</var> be the result
+ of <a>trying</a> to <a>get a script</a> with argument <var>name</var>
```
andreastt: This is nitpicky, but I would prefer the slightly more English:

s/with argument/by/

index.html
```diff
@@ -6153,6 +6200,10 @@ <h3><dfn>Execute Script</dfn></h3>
 <p>The <a>remote end steps</a> are:
 
 <ol>
+ <li><p>Let <var>name</var> be the result of
+  <a>getting a property</a> named <var>name</var>
+  from the <var>parameters</var>.
+
  <li><p>Let <var>body</var> and <var>arguments</var> be the result of
   <a>trying</a> to <a>extract the script arguments from a request</a>
   with argument <var>parameters</var>.
```
andreastt: And somewhere in there (and below, under Execute Async Script) we need to replace the _body_ with the pinned script, right?

index.html
```diff
@@ -6281,7 +6336,45 @@ <h3><dfn>Execute Async Script</dfn></h3>
...
+ <li><p>If <var>script</var> is <a>undefined</a> or is not a <a>String</a>,
+ return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
+
+ <li><p>If <var>name</var> is <a>undefined</a> or is not a <a>String</a>,
+ return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
+
+ <li><p>If <var>name</var> is equal to a key in the <a>executable script map</a>
+ remove the <a>executable script</a> from the session
+
+ <li><p>add the <a>executable script</a> to the session.
```
andreastt: To be more in line with how [Infra suggests using maps](https://infra.spec.whatwg.org/#maps), both of these steps can be replaced with:

```
<li><p>Set <a>executable script map</a>[<var>name</var>] to <var>script</var>.
```

index.html
```diff
@@ -6281,7 +6336,45 @@ <h3><dfn>Execute Async Script</dfn></h3>
...
+
```
andreastt: Missing successful return here.

andreastt:
@titusfortner 

I think there is agreement on the general approach, but there are a number of outstanding issues that prevents this PR from landing.

Do you have time to move this forward?

I would like to wrap up all outstanding PRs before I proceed with https://github.com/w3c/webdriver/issues/1462.

titusfortner:
I'm... not certain I will have time this week. Maybe the weekend or early next week?

christian-bromann:
This can be eventually replaced by the [Bootstrap Scripts](https://github.com/w3c/webdriver/blob/master/webdriver-bidi/bootstrap-scripts.md) idea that was suggested for the BiDi proposal. This would solve the same use case of pinning scripts to a certain execution context (e.g. on a frame before page load) to attach functionality to the script context.

shs96c:
I'm not sure that's true. A bootstrap script is run in every new instance of a context. The pinnable scripts are a shorthand for sending a single script to be executed right now across the wire: that is, the pinned scripts aren't available in every context, and so don't interfere with things like loading times or the performance of the browser.

The two could probably play nicely together: either choosing to update a bootstrap script using a pinned script, or calling something that needs to be on every page that's injected using a bootstrap script via a pinned script.

whimboo:
Recently we landed the spec prose for [Preload scripts in WebDriver BiDi](https://w3c.github.io/webdriver-bidi/#preload-scripts). As such I'm fairly sure that we no longer want to add this feature to WebDriver classic. If we close as wontfix we should also close https://github.com/web-platform-tests/wpt/pull/19675.

@jgraham do you have any objections?

jgraham:
I agree that the main feature here is covered by preload scripts. The additional feature request at TPAC was the ability to use these as part of a locator strategy, but in theory middleware could already make that work (by intercepting the locator strategy and converting it into script execution).

christian-bromann:
This is awesome, thanks folks!

PR #1451, opened 2019-10-15
matching is not defined for strings

titusfortner:
@andreastt based on your comment in #1445 I went through and looked for all of the times "matching" was used to refer to Strings. Let me know if any of these shouldn't be changed. Also, I'm unsure about https://github.com/w3c/webdriver/blob/master/index.html#L7133 but I think it is ok as is.

***
<a href="https://pr-preview.s3.amazonaws.com/titusfortner/webdriver/pull/1451.html" title="Last updated on Oct 15, 2019, 7:38 PM UTC (7066cca)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1451/3216584...titusfortner:7066cca.html" title="Last updated on Oct 15, 2019, 7:38 PM UTC (7066cca)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -630,7 +630,7 @@ when it receives a particular <a>command</a>.
 
  <li><p>Remove each entry in <var>endpoints</var>
   for which the <i>method</i> column
-  is not an exact case-sensitive match for <var>method</var>.
+  is not equal to <var>method</var>.
 
  <li><p>If there are no entries in <var>endpoints</var>,
   return <a>error</a> with <a>error code</a> <a>unknown method</a>.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -623,14 +623,14 @@ <h3>Routing requests</h3>
  <li><p>Remove each entry from <var>endpoints</var>
   for which the concatenation of the <a>URL prefix</a>
   and the entry’s <a>URI template</a>
-  does not match <var>URL</var>’s <a>path</a>.
+  is not equal to <var>URL</var>’s <a>path</a>.
```
andreastt: Ah, well in this particular case it borrows from https://tools.ietf.org/html/rfc6570, which _does_ define matching.

Issue #1093, opened 2017-09-13
Export PDF


eduardohmg:
Spec for printing PDF using browsers' native feature would be useful.

eduardohmg:
Instead of waiting, I can do this by sending an http request to Chrome, like:

 `/session/:sessionId/screenshot`
for screenshot.

Can you tell me what request I have to use in order to get a PDF file on Chrome?

jugglinmike:
A standard way to "print to PDF" would enable a number of use cases:

- web users who would like to capture a full page for use in a PDF reader
- web authors who are targeting print media
- web implementers who want to verify correctness of print-specific web platform features (e.g. [CSS Fragmentation](https://drafts.csswg.org/css-break-3/#intro))

@eduardohmg if you had other uses in mind, please let us know!

For the first two use cases, there are tools like [wkhtmltopdf](https://wkhtmltopdf.org/), and [some browsers expose command-line arguments to perform the task](https://superuser.com/questions/592974/how-to-print-to-save-as-pdf-from-a-command-line-with-chrome-or-chromium). However, because these solutions simply capture the page after it is done loading, they don't offer the kind of control that a WebDriver command would. With a WebDriver command, users could print from any state.

Chrome DevTools Protocol version 1.3 (currently designated as "Stable RC") implements [a `Page.printToPDF` method](https://chromedevtools.github.io/devtools-protocol/1-3/Page#method-printToPDF). This demonstrates that the premise is feasible and it suggests interest beyond Eduardo and me.

@AutomatedTester @andreastt can you speak to how difficult this would be to implement in Firefox? If standardized, would implementation be likely to receive priority?

@gsnedders @jgraham WPT could use this to define reftest-style tests, only comparing PDFs instead of PNGs. Does that sound realistic to you? Do you think we would need a standard, or could we get by with using proprietary binary arguments?

@fantasai Do you think we could use this feature to fully automate tests? Would it be still be useful even if the tests still had to be interpreted by a human?

gsnedders:
@jugglinmike Yes, such a thing would be very much useful for print tests; I had some vague plan to implement roughly this in Marionette for https://github.com/web-platform-tests/wpt/issues/7760. Last I knew there was a lack of interest on the CDP side, which made it impossible to do in ChromeDriver, but obviously that has changed.

jgraham:
> @gsnedders @jgraham WPT could use this to define reftest-style tests, only comparing PDFs instead of PNGs. Does that sound realistic to you?

[...]

PR #1468, opened 2019-12-09
Add support for outputting a paginated document as PDF

jgraham:
This is apparently something that people use automation for rather often, so there should be a cross-browser solution for it.

The featureset is based on what's common in browsers and is close to what CDP offers, without the control over headers and footers, which I propose to defer until we have some experience implementing this more basic featureset.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/pull/1468.html" title="Last updated on Dec 11, 2019, 11:55 AM UTC (16fd8ee)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1468/f462670...16fd8ee.html" title="Last updated on Dec 11, 2019, 11:55 AM UTC (16fd8ee)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -998,6 +998,12 @@ when it receives a particular <a>command</a>.
   <td>/session/{<var>session id</var>}/element/{<var>element id</var>}/screenshot</td>
   <td><a>Take Element Screenshot</a></td>
  </tr>
+
+ <tr>
+  <td>POST</td>
+  <td>/session/{<var>session id</var>}/print</td>
+  <td><a>Print Page</a></td>
+ </tr>
 </table>
 </section> <!-- /Endpoints -->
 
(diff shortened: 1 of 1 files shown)


jgraham:
There may be some kind of WG decision required to add this feature, but I wanted to make the draft spec available asap.

index.html
```diff
@@ -9025,6 +9039,267 @@ <h3><dfn>Take Element Screenshot</dfn></h3>
...
+are specified in inches.
+
+<p>When required to <dfn>parse a page range</dfn> with
+arguments <var>pageRanges</var> and <var>totalPages</var>, an
+implementation must:
+
+<ol>
+  <li>Let <var>pages</var> be an empty <a>Set</a>
+
+  <li>For each <var>range</var> in <var>pageRanges</var>, run the
```
whimboo: Which type do we expect for `pageRanges`? Is it an array or a comma-separated string as used in all the print dialogs?

index.html
```diff
@@ -9025,6 +9039,267 @@ <h3><dfn>Take Element Screenshot</dfn></h3>
...
+    <ol>
+      <li>If <var>range</var> is not either a <a>Number</a> or
+      a <a>String</a>, return <a>error</a> with <a>error
+      code</a> <a>invalid argument</a>.
+
+      <li><p>If <var>range</var> is a <a>Number</a>:
+        <ol>
+          <li>If <var>range</var> is not an integer or is less than 0,
+      return <a>error</a> with <a>error code</a> <a>invalid
+      argument</a>
```
whimboo: We also have a max limit, which we might never reach, but would it make sense to include a check here?

index.html
```diff
@@ -9025,6 +9039,267 @@ <h3><dfn>Take Element Screenshot</dfn></h3>
...
+
+ <li><p>Let <var>pageHeight</var> be the result of <a>getting a
+  property with default</a> named <code>width</code> and with a
+  default of <code>11</code> from <var>page</var>.
+
+ <li><p>If either of <var>pageWidth</var> or <var>pageHeight</var> is
+ not an <a>Number</a>, or is less then 0, return <a>error</a>
+ with <a>error code</a> <a>invalid argument</a>.
+
+ <li><p>Let <var>margin</var> be the result of <a>getting a property
```
whimboo: Is that a left-over given that we have individual parameters for each side?

index.html
```diff
@@ -9025,6 +9039,267 @@ <h3><dfn>Take Element Screenshot</dfn></h3>
...
+ <li><p>Let <var>marginLeft</var> be the result of <a>getting a
+  property with default</a> named <code>left</code> and with a
+  default of <code>0.39</code> from <var>margin</var>.
+
+ <li><p>Let <var>marginRight</var> be the result of <a>getting a
+  property with default</a> named <code>right</code> and with a
+  default of <code>0.39</code> from <var>margin</var>.
+
+ <li><p>If any
+ of <var>martinTop</var>, <var>marginBottom</var>, <var>marginLeft</var>,
```
whimboo: `marginTop` please.

index.html
```diff
@@ -9025,6 +9039,267 @@ <h3><dfn>Take Element Screenshot</dfn></h3>
...
+ or <var>marginRight</var> is not an <a>Number</a>, or is less then 0,
```
whimboo: `not a Number`

index.html
```diff
@@ -9025,6 +9039,267 @@ <h3><dfn>Take Element Screenshot</dfn></h3>
...
+ the <a>current browsing context</a> as a PDF, with the following
+ paper settings:
+   <table>
+     <tr>
+       <th>Property
+       <th>Value
+     <tr>
+       <td>Width in inches
+       <td><var>pageWidth</var> if <var>orientation</var> is
+         "<code>portrait</code>" otherwise <var>pageHeight</var>
```
whimboo: I don't think that we have to make a decision based on the `orientation` parameter. The printing engine does it on its own.

index.html
jgraham: What's the max limit?

index.html
jgraham: There aren't individual parameters for each side; the format here is `{margin: {top, bottom, left, right}}`. They're just unpacked into `marginTop` etc. in the algorithm.

index.html
jgraham: Right, sure I don't think this would appear in code we write. But I think you need to specify what the effect of the parameter is in some way.

JohnChen0:
> I think I’d like input from other browser vendors if this is something they’d consider implementing before proceeding with a technical review.

Currently, Chrome DevTools supports print to PDF only in headless mode. If this feature is added to WebDriver spec, ChromeDriver will only be able to support it with headless Chrome initially. Adding support in normal Chrome will depend on if/when resources become available.

Chrome DevTools doesn't currently support `shrinkToFit`, so we may need to ignore that parameter.

jgraham:
> Chrome DevTools doesn't currently support `shrinkToFit`, so we may need to ignore that parameter.

I think `shrinkToFit` is equivalent to `not preferCSSPageSize` from CDP's `Page.printToPDF`?

JohnChen0:
> I think `shrinkToFit` is equivalent to `not preferCSSPageSize` from CDP's `Page.printToPDF`?

Thanks for the clarification.

index.html
```diff
@@ -9025,6 +9039,267 @@ <h3><dfn>Take Element Screenshot</dfn></h3>
 </section> <!-- /Take Element Screenshot -->
 </section> <!-- /Screen capture -->
 
+<section>
+<h2>Print</h2>
+
+<p>The print functions are a mechanism to render the document to a
```
shs96c: Does this have anything to do with print media CSS? Or is the HTML rendering used?

index.html
```diff
@@ -9025,6 +9039,267 @@ <h3><dfn>Take Element Screenshot</dfn></h3>
...
+ paginated format. It is returned to the <a>local end</a> as a Base64
+ encoded string containing a PDF representation of the paginated
+ document.
+
+<p class=note>
+For historical reasons, all dimensions relating to the printed output
+are specified in inches.
```
shs96c: Not even using the metric system? 😱

index.html
```diff
@@ -9025,6 +9039,267 @@ <h3><dfn>Take Element Screenshot</dfn></h3>
...
+
+<p>When required to <dfn>parse a page range</dfn> with
```
shs96c: How does the local end tell the valid values for the page range? I can't see that in this diff, but maybe I've missed it.

index.html
```diff
@@ -9025,6 +9039,267 @@ <h3><dfn>Take Element Screenshot</dfn></h3>
...
+arguments <var>pageRanges</var> and <var>totalPages</var>, an
+implementation must:
+
+<ol>
+  <li>Let <var>pages</var> be an empty <a>Set</a>
```
shs96c: Surely a list? Order matters, and I can imagine identical pages being present in the output.

index.html
```diff
@@ -9025,6 +9039,267 @@ <h3><dfn>Take Element Screenshot</dfn></h3>
...
+
+  <li>For each <var>range</var> in <var>pageRanges</var>, run the
+  following steps:
+    <ol>
+      <li>If <var>range</var> is not either a <a>Number</a> or
```
shs96c: Presumably the default range is "every page"?

index.html
```diff
@@ -9025,6 +9039,267 @@ <h3><dfn>Take Element Screenshot</dfn></h3>
...
+  and trailing <a>whitespace</a> characters from <var>input</var>.
+
+  <li>If <var>stripped</var> has zero length, return <a>error</a> with
+  status <a>invalid argument</a>.
+
+  <li>If <var>stripped</var> contains any characters outside the
+  range <code>U+0030</code> - <code>U+0039</code> (i.e. 0 - 9)
+  inclusive, return <a>error</a> with status <a>invalid argument</a>.
+
+  <li>Let <var>output</var> be the result of calling <a>parseInt</a>
```
shs96c: Trying to call? What happens if `parseInt` is not successful?

index.html
```diff
@@ -9025,6 +9039,267 @@ <h3><dfn>Take Element Screenshot</dfn></h3>
...
+ <li><p>If <var>background</var> is not a <a>Boolean</a>
+ return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
+
+ <li><p>Let <var>page</var> be the result of <a>getting a property
+  with default</a> named <code>page</code> and with a default of an
+  empty <a>Object</a> from the <var>parameters</var> argument.
+
+ <li><p>Let <var>pageWidth</var> be the result of <a>getting a
+  property with default</a> named <code>width</code> and with a
+  default of <code>8.5</code> from <var>page</var>.
```
shs96c: So this is equivalent to US letter size? Any chance we can have the equivalents in either millimetres or centimetres? It seems unfortunate to write a spec in the 21st century that relies on imperial measurements.

jgraham: It makes my skin crawl. But US letter size is defined in inches, and there's precedent for using inches, so I went with that. Happy to change this to use `cm` if there's a general feeling that it's acceptable.

index.html
jgraham: The previous steps ensure we have a string containing only characters in the range 0-9. I think in that case `parseInt` has to succeed.

index.html
jgraham: Yes, but that's handled by the caller.

index.html
jgraham: The way this is used is just to select every page that's in the set, not to change the order. Do print subsystems actually allow you to do things like `3,4,1-2` to change the output order? If they do maybe we ned to change the wording here.

index.html
jgraham: That isn't covered; I don't know what a solution would look like. In practice I'm guessing this range stuff is relatively uncommon and people mostly just want the full output.

index.html
jgraham: Awful, isn't it.

index.html
jgraham: I would assume that print stylesheets are applied in the same way they would for actual printing, maybe that needs to be called out explicitly.

index.html
jgraham: I tested and at least Firefox works the way I assumed; you get the pages in the default order but only if they appear in the specified ranges.

index.html
jgraham: OK, I changed this to use `cm` on the basis that although we can't, in general, have nice things, there's a basic level of human decency which we have to maintain.

jugglinmike:
This resolves gh-1093.

index.html
eduardohmg: It's in the line [9233](https://github.com/w3c/webdriver/blob/16fd8ee87ddac27cad9000eae5129a405496dc0e/index.html#L9233). That's it?

index.html
```diff
@@ -9025,6 +9039,263 @@ <h3><dfn>Take Element Screenshot</dfn></h3>
...
+  property with default</a> named <code>pageRanges</code> from
+  the <var>parameters</var> argument with default of an
+  empty <var>Array</var>.
+
+ <li><p>If <var>pageRanges</var> is not an <a>Array</a>
+ return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
+
+ <li><p>When the user agent is next to <a>run the animation frame
+ callbacks</a>, let <var>pdfData</var> be the result of <a>trying</a>
+ to take UA-specific steps to generate a paginated representation of
```
gsnedders: It seems to be implied below that `pdfData` is an array of PDFs, one per page, given removing pages; this should be spelt out here. (I mean a PDF itself can be paginated, so what does "a paginated representation" mean? Or have I misunderstood what the return value of this command is? Either way it should be clarified.

jgraham: There isn't supposed to be an arry of PDFs, just a single PDF. The point here was supposed to be that the PDF should use the appropriate CSS &c. for a paginated representation rather than something else .

gsnedders: Can we clarify that it's a single multi-page PDF?

PR #1471, opened 2019-12-11
Define non-typeable form controls more exhaustively

andreastt:
The definition used for which form control elements not to use the
actions API for is rather sketchy because it includes inline examples
in a match guard.

This change attempts to formalise the definition of non-typeable
form controls.  It also includes a new example section to guide
implementors, since the definition is rather weak (since we can't
tell UAs how to represent anything.)

***
<a href="https://pr-preview.s3.amazonaws.com/andreastt/webdriver/pull/1471.html" title="Last updated on Dec 11, 2019, 2:52 PM UTC (739d00d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1471/f462670...andreastt:739d00d.html" title="Last updated on Dec 11, 2019, 2:52 PM UTC (739d00d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -5476,6 +5476,32 @@ and then sends the provided keys to the <a>element</a>.
 In case the <a>element</a> is not <a>keyboard-interactable</a>,
 an <a>element not interactable</a> <a>error</a> is returned.
 
+<p>
+A <dfn>non-typeable form control</dfn>
+is an <a><code>input</code> element</a>
+whose <a data-lt="input type state"><code>type</code> attribute</a> state
+causes the primary input mechanism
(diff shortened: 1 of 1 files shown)


andrewnicols:
Hi @andreastt ,

It's not clear from these change which category the `date` and `local-datetime` input types fall into. I assume that you mean to indicate that in these cases the primary input mechanism is _not_ the keyboard. 
Whilst this update is an improvement, I don't think that it's clear to say that the decorations added for a `date` or `local-datetime` input type mean that the primary input mechanism changes.

filled
entry

PR #1493, opened 2020-03-27
Remove scroll behavior so it uses default.

### LINKED ISSUE #1449: "instant" is not valid scrollIntoView behavior
In chapter [12. Elements](https://w3c.github.io/webdriver/#dfn-scrolls-into-view#x12-elements):
> "behavior"
> "instant"

However, per https://drafts.csswg.org/cssom-view/#enumdef-scrollbehavior ScrollBehavior only has values "auto" and "smooth".

AutomatedTester:
CSSOM-View has removed the behaviour from ScrollIntoView dictionary.
This change updates to remove it from our scroll and means that browsers
should default to `auto` as described in CSSOM-View. Fixes #1449

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/pull/1493.html" title="Last updated on Mar 27, 2020, 10:49 AM UTC (0e22ed1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1493/4265381...0e22ed1.html" title="Last updated on Mar 27, 2020, 10:49 AM UTC (0e22ed1)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3995,9 +3995,6 @@ argument <var>reference</var>, run the following steps:
   the following <a><code>ScrollIntoViewOptions</code></a>:
 
   <dl>
-   <dt>"<code>behavior</code>"
-   <dd>"<code>instant</code>"
-
    <dt><a>Logical scroll position "<code>block</code>"</a>
    <dd>"<code>end</code>"
 


christian-bromann:
@shs96c I've seen cases where scrolling an element into view didn't work because another element got animated on top of it so I think this would be very useful. However is it actually possible to define that moment without running into edge cases where there is an endless animation loop in which case this action never finishes.

AutomatedTester:
> Do we need prose to say that we should wait for the browser animation to finish first?

No, I don't think we do. If we do we can do down the route of having to watch for mutations on the DOM that could cause reflow/repaint and there feels like there would be a lot of dragons down there.

If you want it please raise a separate issue

andreastt:
This PR aligns the specification with what browsers are doing in reality since `instant` was removed, so thanks to @AutomatedTester for fixing it.

Two further DOM synchronisation issues were brought up here:

1. The swoosh-into-view animation that may, depending on the browser configuration, happen with `auto` takes time and `scrollIntoView` is not atomic.

2. Another element may be animated so it obscures the target coordinates after scrolling, which may cause the interaction to fail.

Because `instant` is not a universal feature, one option is to tell browsers to run `scrollIntoView` without animations. It is possible to support in Gecko by adding a preference in [`PresShell::ScrollContentIntoView`](https://searchfox.org/mozilla-central/source/layout/base/PresShell.cpp#3407) and I imagine a similar tactic is possible in Chromium. This means WebDriver gets the effect of `instant` _without_ actually standardising the feature.

The second issue is also real, but we won’t run into an endless loop because the intractability check is before interaction. This means the client can distinguish between the _interaction failing_ and the _scroll into view failing_:

> 5. Scroll into view the element’s container.
>
> 6. If element’s container is still not in view, return error with error code element not interactable.
>
> 7. If element’s container is obscured by another element, return error with error code element click intercepted.

I think it’s worth filing an issue to track the synchronisation problem with animations. Perhaps a better solution than the one I outlined can be found.

Issue #1294, opened 2018-08-23
User prompts for "beforeunload" event should not automatically be dismissed


kovalev007:
In my opinion Section 18 contains not correct logic.

- User prompts that are spawned from beforeunload event handlers, are dismissed implicitly upon navigation or close window, regardless of the defined user prompt handler.

Example:
If we have onbeforeunload event and user try close window:
In chrome 68.0.3440.106 user should interact with promt before window will closed.
In firefox 61 user should interact with promt before window will closed.
In firefox 62beta user should interact with promt before window will closed.
In firefox 63nightly user should interact with promt before window will closed.

But in automated test with geckodriver we have different behavior.
geckodriver 0.21.0 silently set dom.disable_beforeunload to true. (in firefox 61, 62, 63)
and when test close window promt not showing and window closed.

It is different behaviour.

kovalev007:
Additional info:
https://github.com/mozilla/geckodriver/issues/617#issuecomment-415288372
https://github.com/mozilla/geckodriver/issues/617#issuecomment-415296231
https://github.com/mozilla/geckodriver/issues/617#issuecomment-415386466

Jarob22:
Hi - this burned us today whilst trying to write scripts to validate the appearance of a prompt on window close - is there some way we can get this addressed, please?

zeit200:
[https://github.com/mozilla/geckodriver/issues/1563](url)
Hi, wanted to add that this behaviour blocks verifying of custom obeforeunload logic in automated tests

Snipx:
Dear team, are there are news on this one? I agree that the specification currently blocks the use case of testing the user prompts on page reload/exit, and some of the implementing parties claim that they need to be compliant with the spec so I think the first move should be done by the spec community

whimboo:
Note that there is ongoing work for extending WebDriver classic for WebDriver BiDi purposes on PR #1791.

ioquatix:
I'm a bit confused about this issue.

I had the following test which no longer works in the latest chromedriver:

```
<dd><form action="cancel" method="post" onsubmit="return confirm('Are you sure you want to cancel the subscription?');"><button type="submit">Cancel Subscription</button></form></dd>
```

Test code:

```
click_button("Cancel Subscription")
accept_alert

expect(session).to have_element(css: ".subscription .status").and(
	have_attributes(text: be =~ /canceled/)
)
```

[...]

PR #1518, opened 2020-06-01
Fix various minor mishaps with <dfn> and <a>

foolip:
Discovered while converting to Bikeshed but also affects ReSpec:
- data-lt=go and data-lt="input type state" don't work since
  there's no matching <dfn>.
- data-lt=perform-actions doesn't make sense, the "Perform Actions"
  section isn't the intended target.
- Duplicate "Resettable element" definition
- Broken canvas width/height links

***
<a href="https://pr-preview.s3.amazonaws.com/foolip/webdriver/pull/1518.html" title="Last updated on Jun 1, 2020, 12:15 PM UTC (1f3cdb4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1518/162e738...foolip:1f3cdb4.html" title="Last updated on Jun 1, 2020, 12:15 PM UTC (1f3cdb4)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1461,7 +1461,7 @@ with a "<code>moz:</code>" prefix:
   <td>"<code>acceptInsecureCerts</code>"
   <td>boolean
   <td>Indicates whether untrusted and self-signed TLS certificates
-   are implicitly trusted on <a data-lt=go>navigation</a>
+   are implicitly trusted on <a>navigation</a>
    for the duration of the <a>session</a>.
  </tr>
 
@@ -1913,7 +1913,7 @@ with a "<code>moz:</code>" prefix:
    <dt>"<code>acceptInsecureCerts</code>"
    <dd><a>Boolean</a> initially set to false,
     indicating the session will not implicitly trust untrusted
-    or self-signed TLS certificates on <a data-lt=go>navigation</a>.
+    or self-signed TLS certificates on <a>navigation</a>.
 
    <dt>"<code>strictFileInteractability</code>"
     <dd><a>Boolean</a> initially set to false,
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1461,7 +1461,7 @@ <h2>Capabilities</h2>
   <td>"<code>acceptInsecureCerts</code>"
   <td>boolean
   <td>Indicates whether untrusted and self-signed TLS certificates
-   are implicitly trusted on <a data-lt=go>navigation</a>
```
jtnord, line 1464: just noting this was incorrect.
Go was renamed in #1040 so this should be to `navigate to` 
this now goes to `navigate` which is a completely different definition

foolip: Good catch, @jtnord! It looks like there are actually a bunch of incorrect links for "navigation" leading to https://w3c.github.io/webdriver/#dfn-navigating, I'll file an issue.

jgraham: I think this was wrong before and is now correct. Self-signed certs should be trusted irrespective of whether the navigation was via the `navigate` command or any other method (e.g. synthesizing a click on a link).

jtnord: Whilst this may now be the correct definition, it was not intended to be and this change was an accident.  I should have really commented on [before-unload prompts -line8619](https://github.com/w3c/webdriver/pull/1518/files#diff-0eb547304658805aad788d320f10bf1f292797b5e6d745a3bf617584da017051L8619) and commented on the first change I found instead.  That said I still believe these accidental changes should be reverted, and *then* you can have the discussion about if it was correct before or not and make that change.

whimboo: @jgraham based on our findings here  #1574 got closed. So shall we also resolve this particular conversation too? We should rather continue the discussion about `beforeunload` prompts via #1294.

PR #1536, opened 2020-07-16
Proposing a get endpoint for Take Screenshot that with fullpage option

k7z45:
Due to the high demand from
https://bugs.chromium.org/p/chromedriver/issues/detail?id=294,
come up with initial proposal for the spec

***
<a href="https://pr-preview.s3.amazonaws.com/k7z45/webdriver/pull/1536.html" title="Last updated on Nov 30, 2020, 6:56 PM UTC (1a802aa)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1536/536adfe...k7z45:1a802aa.html" title="Last updated on Nov 30, 2020, 6:56 PM UTC (1a802aa)">Diff</a>

- - -
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1536)

### DIFF
--- a/index.html
+++ b/index.html
@@ -1017,9 +1017,9 @@ when it receives a particular <a>command</a>.
  </tr>
 
  <tr>
-  <td>POST</td>
-  <td>/session/{<var>session id</var>}/screenshot</td>
-  <td><a>Take Screenshot with option</a></td>
+  <td>GET</td>
+  <td>/session/{<var>session id</var>}/screenshot/full</td>
+  <td><a>Take Fullpage Screenshot</a></td>
  </tr>
 
  <tr>
@@ -9041,7 +9041,7 @@ sets the text field of a <a>window.<code>prompt</code></a>
  and a <a>command</a> <a>Take Element Screenshot</a>
  for doing the same with the visible region of an <a>element</a>’s
  <a>bounding rectangle</a> after it has been <a>scrolled into view</a>.
- It also provides a <a>Take Screenshot with option</a> <a>command</a>
+ It also provides a <a>Take Fullpage Screenshot</a> <a>command</a>
  to optionally capture the <a>fullpage region</a>.
 
 <p>An <a>element</a>’s <dfn data-lt="fullpage region">fullpage region</dfn>
@@ -9291,7 +9291,7 @@ that evaluates to false, the <a>element</a> will not be <a>scrolled into view</a
 </section> <!-- /Take Element Screenshot -->
 
 <section>
-<h3><dfn>Take Screenshot with option</dfn></h3>
+<h3><dfn>Take Fullpage Screenshot</dfn></h3>
 
 <table class="simple jsoncommand">
  <tr>
@@ -9299,8 +9299,8 @@ that evaluates to false, the <a>element</a> will not be <a>scrolled into view</a
   <th>URI Template</th>
  </tr>
  <tr>
-  <td>POST</td>
-  <td>/session/{<var>session id</var>}/screenshot</td>
+  <td>GET</td>
+  <td>/session/{<var>session id</var>}/screenshot/full</td>
  </tr>
 </table>
 
@@ -9310,11 +9310,7 @@ that evaluates to false, the <a>element</a> will not be <a>scrolled into view</a
  <li><p>If the <a>current top-level browsing context</a> is <a>no longer open</a>,
   return <a>error</a> with <a>error code</a> <a>no such window</a>.
 
- <li><p>Let <var>fullpage</var> be the result of <a>getting a
-    property with default</a> named <code>fullpage</code> and with
-    default <code>true</code> from the <var>parameters</var> argument.
-
-    <li><p>If <var>fullpage</var> is true, when the user agent is next to <a>run the animation frame callbacks</a>:
+    <li><p>When the user agent is next to <a>run the animation frame callbacks</a>:
   <ol>
 
    <li><p>Let <var>content rect</var> be the <a>current top-level browsing context</a>’s
@@ -9334,7 +9330,6 @@ that evaluates to false, the <a>element</a> will not be <a>scrolled into view</a
   </ol>
 
  <li><p>Return <a>success</a> with data <var>encoded string</var>.
- <li><p>Otherwise, run the steps for <a>Take Screenshot</a>
 </ol>
 
 </section> <!-- /Take Screenshot with option -->
--- a/index.html
+++ b/index.html
@@ -1012,14 +1012,14 @@ when it receives a particular <a>command</a>.
 
  <tr>
   <td>GET</td>
-  <td>/session/{<var>session id</var>}/element/{<var>element id</var>}/screenshot</td>
-  <td><a>Take Element Screenshot</a></td>
+  <td>/session/{<var>session id</var>}/screenshot/full</td>
+  <td><a>Take Full Screenshot</a></td>
  </tr>
 
  <tr>
   <td>GET</td>
-  <td>/session/{<var>session id</var>}/screenshot/full</td>
-  <td><a>Take Full Page Screenshot</a></td>
+  <td>/session/{<var>session id</var>}/element/{<var>element id</var>}/screenshot</td>
+  <td><a>Take Element Screenshot</a></td>
  </tr>
 
  <tr>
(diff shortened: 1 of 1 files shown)


jgraham:
geckodriver supports this, but it's a new endpoint rather than a parameter (`…/screenshot/full`). Since the existing endpoint is a `GET` and we don't have other cases that use parameters on the GET, that design would make more sense to me.

k7z45:
Thanks for the comment, made the change accordingly.

index.html
```diff
@@ -9035,6 +9041,74 @@ <h2>Screen capture</h2>
...
+ <li><p>Let <var>height</var> be the value
+  returned by
+  <p><a>max</a>(document.body.scrollHeight, document.documentElement.scrollHeight,
+                document.body.offsetHeight, document.documentElement.offsetHeight,
+                document.body.clientHeight, document.documentElement.clientHeight)
+
+  <li><p>Return {<var>x</var>, <var>y</var>, <var>width</var>, <var>height</var>}.
+</ol>
+
+<p>In order to <dfn>draw a bounding box from the fullpage region</dfn>,
```
jgraham, line 9071: Could this just reuse the "draw a bounding box from the framebuffer" steps? It seems to be mostly duplicated.

index.html
```diff
@@ -9035,6 +9041,74 @@ <h2>Screen capture</h2>
...
+ and height,
+ it can be calculated this way:
+<ol>
+ <li><p>Let <var>x</var> be 0
+
+ <li><p>Let <var>y</var> be 0
+
+ <li><p>Let <var>width</var> be the value
+  returned by
+  <p><a>max</a>(document.body.scrollWidth, document.documentElement.scrollWidth,
```
jgraham, line 9058: Are there really no defined relations that enforce an ordering between any of these properties?

jgraham (changes requested):
Note that merging this also requires some web-platform-tests to be submitted/approved for the feature.

index.html
k7z45: I am not sure what you mean regarding ordering of the properties. I took this "heuristics" from https://javascript.info/size-and-scroll-window. In reality, ChromeDriver implementation will probably utilize CDP https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-getLayoutMetrics and use the contentSize width and height directly.

k7z45:
Will work on web-platform-tests  once I have "verbal" approval for this.

sideshowbarker:
Hi @k7z45 — in order to accept this PR we need to make our tooling recognize that you’re already a participant in the Browser Testing and Tools Working Group.

You can do that by using https://www.w3.org/users/myprofile to log into your W3C account — and from there, follow the **Connected Accounts** link in the sidebar to connect your W3C account with your GitHub account.

If you don’t remember your W3C username or password, you can use https://www.w3.org/accounts/recover to recover them.

k7z45:
Thanks Michael! Just did that.

sideshowbarker:
Thanks @k7z45 — https://labs.w3.org/repo-manager/pr/id/w3c/webdriver/1536 shows this is now green for the IPR check

k7z45:
Hi @jgraham ,
Pending review web-platform-tests: https://github.com/web-platform-tests/wpt/pull/26652 for the feature.

index.html
```diff
@@ -1016,6 +1016,12 @@ <h3>Endpoints</h3>
   <td><a>Take Element Screenshot</a></td>
  </tr>
 
+ <tr>
+  <td>GET</td>
+  <td>/session/{<var>session id</var>}/screenshot/full</td>
+  <td><a>Take Full Page Screenshot</a></td>
```
whimboo, line 1022: We always capture the page, so maybe this should just be `Take Full Screenshot`?

index.html
```diff
@@ -9035,6 +9041,8 @@ <h2>Screen capture</h2>
  and a <a>command</a> <a>Take Element Screenshot</a>
  for doing the same with the visible region of an <a>element</a>’s
  <a>bounding rectangle</a> after it has been <a>scrolled into view</a>.
+ It also provides a <a>Take Full Page Screenshot</a> <a>command</a>
```
whimboo, line 9044: Please move this part up so it's listed before `Take Element Screenshot`, and continues the sentence.

index.html
```diff
@@ -9082,6 +9090,45 @@ <h2>Screen capture</h2>
  <li><p>Return <a>success</a> with <var>canvas</var>.
 </ol>
 
+<p>In order to <dfn>draw a scrollable region from the framebuffer</dfn>,
```
whimboo, line 9093: What exactly is that part about? Is it for the document element, or a child node? For child nodes we have `Take Element Screenshot`. Otherwise check you do below we already have at the top of the former list.

I feel we don't have to repeat all the steps again.

index.html
```diff
@@ -9215,6 +9262,47 @@ <h3><dfn>Take Element Screenshot</dfn></h3>
  <li><p>Return <a>success</a> with data <var>encoded string</var>.
 </ol>
 </section> <!-- /Take Element Screenshot -->
+
+<section>
+<h3><dfn>Take Full Page Screenshot</dfn></h3>
```
whimboo, line 9267: Please move before `Take Element Screenshot` because it's closer related to `Take Screenshot`.

k7z45: Done

index.html
k7z45: Done.

index.html
k7z45: Done.

index.html
k7z45: This is for describing taking the screenshot for the full page instead of only the visible viewport.
It's for document.scrollingElement.
I did not use "draw a bounding box from the framebuffer" because I think step 2 and step 3
do not satisfy the requirement for full page screenshot.

"2. Let paint width be the initial viewport’s width – min(rectangle x coordinate, rectangle x coordinate + rectangle width dimension).
3. Let paint height be the initial viewport’s height – min(rectangle y coordinate, rectangle y coordinate + rectangle height dimension)."

Notice the - denotes the minus sign, so I think  for Take Screenshot and Take Element Screenshot, the width and height will not be greater than the visible viewport's width and height but the full page screenshot takes the document.scrollingElement.scrollWidth and scrollHeight which can be greater. So I removed step 2 and 3 and uses the scrollWidth and Height directly for paint width and height.

k7z45: @whimboo Could you take a look, please?

jugglinmike:
I'm interested in [bringing this functionality to WebDriver BiDi](https://github.com/w3c/webdriver-bidi/issues/384). Whether it belongs here in WebDriver "classic" is not a topic I can comment on, but I'm interested in learning if the normative text proposed here is viable for either spec.

Specifically, I haven't found any implementer's perspective on @jgraham's [2021 comment](https://github.com/w3c/webdriver-bidi/issues/120#issuecomment-882399249):

> One challenge is that the `fullPage` option may not be trivial to specify; I think we skipped it for WebDriver Classic but in practice implementations have it, although I don't know they all work the same The typical kind of problem is "what if you haven't already rendered the whole page, but just some tiles. Then you might need to simulate scrolling to render the whole page, but what about position:fixed headers, or content that changes duirng scroll". So really you want "render the entire current layout tree as if the viewport length was unbounded", but that might not actually be a thing that's supported in engines.

@chrishtr [you and I investigated a Chrome rendering bug a number of years ago](https://github.com/jugglinmike/chrome-screenshot-race/issues/1); is there any chance you could weigh in here or perhaps refer me to someone more appropriate?

@juliandescottes could you direct me to a subject-matter expert at Mozilla?

whimboo:
For Chrome there is https://bugs.chromium.org/p/chromium/issues/detail?id=770769&desc=2 which at least for headless seems to still limit the maximum height to 16384px.

[For Firefox we currently have the limit set ](https://searchfox.org/mozilla-central/rev/09f7e00da9eaef55c80620c17e8575ddd4532451/remote/shared/Capture.sys.mjs#15-16) to 32767px for both height and width, and a maximum texture size of 472907776.

As of now we do not have problems in capturing full page screenshots for pages exceeding the visible viewport.

whimboo:
Note that we will discuss the full screenshot feature for BiDi during the TPAC sessions on Thursday or Friday this week.

PR #1541, opened 2020-08-29
Modify: SameSite now defaults to Lax

Harsha509:
## Previous Behaviour

SameSite defaults to `None` and it can be set either `Strict` or `Lax`

## New Behaviour

From the data points provided in https://web.dev/samesite-cookies-explained/#samesitenone-must-be-secure and in https://support.google.com/chrome/a/answer/7679408#84, sameSite now deefaults to Lax and also when set to `None`  must also be marked as `Secure`.

***
<a href="https://pr-preview.s3.amazonaws.com/Harsha509/webdriver/pull/1541.html" title="Last updated on Aug 29, 2020, 2:34 PM UTC (43fe4b4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1541/f0c7bdf...Harsha509:43fe4b4.html" title="Last updated on Aug 29, 2020, 2:34 PM UTC (43fe4b4)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -6577,8 +6577,11 @@ The first argument provided to the function will be serialized to JSON and retur
   <td>"<code>SameSite</code>"
   <td>✓
   <td>Whether the cookie applies to a SameSite policy.
-    Defaults to None if omitted when <a>adding a cookie</a>.
-    Can be set to either <a><code>Lax</code></a> or <a><code>Strict</code></a>.
+    Defaults to Lax if omitted when <a>adding a cookie</a>.
+    Can be set to either <a><code>None</code></a> or <a><code>Strict</code></a>.
+
+  cookie with SameSite=None must also specify secure.
+
  </tr>
 </table>
 
--- a/index.html
+++ b/index.html
@@ -176,7 +176,7 @@ produces the largest integer, closest to positive infinity,
 that is not larger than <var>value</var>.
 
 <p>
-A <dfn data-lt=uuid>Universally Unique IDentifier (UUID)</dfn>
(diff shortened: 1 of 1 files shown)


whimboo:
So what about backward compatibility with older versions of the browser that don't set `Lax` by default? This change in behavior by browsers is fairly new.

Also once decided the appropriate WebDriver cookie tests from the [web-platform-tests repository](https://github.com/web-platform-tests/wpt/tree/master/webdriver/tests) would have to be updated with that change.

Harsha509:
> So what about backward compatibility with older versions of the browser that don't set `Lax` by default? This change in behavior by browsers is fairly new.
> 
> Also once decided the appropriate WebDriver cookie tests from the [web-platform-tests repository](https://github.com/web-platform-tests/wpt/tree/master/webdriver/tests) would have to be updated with that change.

Agreed!

Any suggestions on how to modify in doc, because the behaviour `sameSite=None` is specific for few versions of browser( as i know it is introduced in chrome 80 - 83 and from 84 it is `Lax`) and we need to have both behaviours to support backward compatibility in WebDriver.

Any suggestions from PLC team please !

whimboo:
Maybe @jgraham can give better input than me in how to handle that. Maybe we have to wait until all browsers have it enabled by default, and only then switch? Or finding preferences/settings for those browsers that work with `lax` by default, and turn it off?

Harsha509:
ok @whimboo !.

Thank you for the inputs.

Will wait until all browsers have it enabled. Will close this pr for now!

Thanks,
Harsha

PR #1543, opened 2020-09-08
Store the top-level and parent browsing contexts explicitly.

jgraham:
Previously the parent and top-level browsing contexts were computed
properties based on the current browsing context. This fails in the
case where the frame containing the current browsing context is
removed from the DOM tree since the browsing context is then discarded
meaning that trying to compute its parents will fail. In practice we
expect operations working with the top-level browsing context to
continue working in that case (and indeed implementations are
mostly/entirely not following the current spec), so this makes the
top-level and parent browsing contexts explictly stored properties of
the session that are updated at the point where we change browsing
contexts.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/pull/1543.html" title="Last updated on Sep 8, 2020, 1:10 PM UTC (43f6e94)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1543/f0c7bdf...43f6e94.html" title="Last updated on Sep 8, 2020, 1:10 PM UTC (43f6e94)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3465,20 +3465,21 @@ make the tab containing the <a>browsing context</a> the selected tab.
   or an <a>Object</a> that <a>represents a web element</a>,
   return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
 
- <li><p>If the <a>current browsing context</a> is <a>no longer open</a>,
-  return <a>error</a> with <a>error code</a> <a>no such window</a>.
-
- <li><p><a>Handle any user prompts</a>
-  and return its value if it is an <a>error</a>.
-
  <li><p>Run the substeps of the first matching condition:
 
   <dl class=switch>
    <dt><var>id</var> is <a><code>null</code></a>
    <dd>
     <ol>
+      <li><p>If the <a>current top-level browsing context</a> is <a>no
+            longer open</a>, return <a>error</a> with <a>error code</a> <a>no
+            such window</a>.
+
+      <li><p><a>Handle any user prompts</a>
+          and return its value if it is an <a>error</a>.
+
      <li><p><a>Set the current browsing context</a> with
-      the <a>current top-level browsing context</a>.
+         the <a>current top-level browsing context</a>.
     </ol>
 
    <dt><var>id</var> is a <code>Number</code> object
@@ -3486,6 +3487,13 @@ make the tab containing the <a>browsing context</a> the selected tab.
     <ol>
      <li><p>If <var>id</var> is less than 0 or greater than 2<sup>16</sup> – 1,
       return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
+
+     <li><p>If the <a>current browsing context</a> is <a>no
+           longer open</a>, return <a>error</a> with <a>error code</a> <a>no
+           such window</a>.
+
+     <li><p><a>Handle any user prompts</a>
+         and return its value if it is an <a>error</a>.
 
      <li><p>Let <var>window</var> be the <a>associated window</a>
       of the <a>current browsing context</a>’s <a>active document</a>.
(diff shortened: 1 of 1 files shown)


whimboo:
Please note that appropriate WebDriver tests will be added/updated via https://bugzilla.mozilla.org/show_bug.cgi?id=1493108 once that PR has been approved.

index.html
```diff
@@ -3539,10 +3561,9 @@ <h3><dfn>Switch To Parent Frame</dfn></h3>
 
  <li><p><a>Handle any user prompts</a> and return its value if it is an <a>error</a>.
 
- <li><p>If the <a>current browsing context</a> is not equal to
-  the <a>current top-level browsing context</a>,
-  set the <a>current browsing context</a> to
-  the <a>parent browsing context</a> of the <a>current browsing context</a>.
+ <li><p>If the <a>current session</a>'s <a>current parent browsing
+ context</a> is not <a>null</a>, <a>set the current browsing
+ context</a> with the <a>current parent browsing context</a>.
```
whimboo, line 3566: Please note that the first check for `current browsing context` of the remote steps still cause an early abort of the command, which leads to not being able to switch to the parent context, if the current one is gone.

index.html
```diff
@@ -3455,7 +3477,7 @@ <h3><dfn>Switch To Frame</dfn></h3>
    <dt><var>id</var> is <a><code>null</code></a>
    <dd>
     <ol>
-     <li><p>Set the <a>current browsing context</a> to
+     <li><p><a>Set the current browsing context</a> with
```
whimboo, line 3480: Please note that the check for the `current browsing context` above is still faulty and wouldn't allow switching to the top-level browsing context at all if the current one is gone, but returns early.

index.html
```diff
@@ -3465,6 +3488,13 @@ <h3><dfn>Switch To Frame</dfn></h3>
      <li><p>If <var>id</var> is less than 0 or greater than 2<sup>16</sup> – 1,
       return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
 
+     <li><p>If the <a>current browsing context</a> is <a>no
+           longer open</a>, return <a>error</a> with <a>error code</a> <a>no
+           such window</a>.
```
whimboo, line 3493: This should be a `no such frame` error, right? Or should it only be the used when the current browsing context is not the top-level one?

jgraham: I think `no such frame` is only for the case where the specified frame can't be found, and `no such window` is always used for the case where the current browsing context can't be found.

index.html
```diff
@@ -2160,14 +2160,14 @@ <h2>Sessions</h2>
...
-<p>A <a>session</a> has an associated
- <dfn>current top-level browsing context</dfn>, which is the <a>current browsing
- context</a> if it itself is a <a>top-level browsing context</a>, and otherwise
- is the <a>top-level browsing context</a> of the <a>current browsing
- context</a>.
+<p>A <a>session</a> has an associated <dfn>current browsing
+ context</dfn>, which is the <a>browsing context</a> against
+ which <a>commands</a> will run, an associated <dfn>current parent
+ browsing context</dfn>, which is set to the parent of the <a>current
+ browsing contexts</a> when changing browsing contexts, and an
```
mjzffr, line 2167: An extra "s" to remove?

> current parent browsing context, which is set to the parent of the current browsing context**s** when changing browsing contexts,

jgraham: Oops, thanks.

index.html
whimboo: Ok, that makes sense. Otherwise also all the other commands would have to be changed too. And for those it doesn't make sense.

whimboo (approved):
From my perspective it looks fine, but it would be good to also get a review from @AutomatedTester and/or @shs96c.

AutomatedTester (approved):

PR #1552, opened 2020-10-06
Async session creation

christian-bromann:
As a cloud provider starting a WebDriver session can take a while for arbitrary reasons. Long session creation times are problematic for users as request libraries tend to time out after a while. Currently workarounds are used where you would respond with a 302 redirect, redirecting to the same resource. Having a way to asynchronously create a session would help to get rid of such hacky workarounds.

This patch was [original authored](https://gist.github.com/shs96c/108f5313eae54b94658ee018e37926d2) by @shs96c during TPAC 2018. I would love to finally pick this up and bring it into the spec.

At a high level these would be the step to create an asynchronous session:

1. Send a request with same payload as for synchronous session requests to `[POST] /session/async`
   Response payload would be as follows:
   ```json
   {
       "sessionCreationJobId": "6fcd0780e995bd50fedd330cd49d0531",
       "status": "queueing",
       "timeout": 600000,
       "requestedTime": 1601990778990,
       "events": [],
       "data": {}
   }
   ```
1. Regularly check for the status of the session creation job by making requests to `[GET] /session/async/6fcd0780e995bd50fedd330cd49d0531`
   Response payload once job finished loading would be
   ```json
   {
       "sessionCreationJobId": "6fcd0780e995bd50fedd330cd49d0531",
       "status": "created",
       "timeout": 600000,
       "requestedTime": 1601990778990,
       "endedTime": 1601990878990,
       "events": [{ eventId: "eventId-1" }, { eventId: "eventId-2" }, { eventId: "eventId-3" }, { eventId: "eventId-4" }, ...],
       "data": {}
   }
   ```
1. If there is interest introspecting events happen while creating the job the user can query for an event id via `[GET] /session/async/6fcd0780e995bd50fedd330cd49d0531/eventId-1`
   Response payload for this request would contain
   ```json
   {
       "eventId": "event-1",
       "time": 1601990778990,
       "type": "progress",
       "message": "...",
       "data": "...",
       "stacktrace": "..."
   }
   ```

Notes from Simon when he drafted this with my comments:

> Should session creation progress be represented using “events” or “logs”?

Given that session creation progress is always connected to a point in time I would favor `events` as representation for it.

> Should session creation jobs have multiple timeouts, for example overall job timeout and wait timeout?

One timeout for the maximal amount of time it should take to create a job is fine. Overall job timeout is set as session timeout.

> Should we require representation of session creation attempts? (Compare [AWS Batch Describe Jobs](https://docs.aws.amazon.com/en_pv/batch/latest/APIReference/API_DescribeJobs.html))

IMO attempts can be easier represented in the bindings implementation rather than having to spec this into the protocol.

> Should we use a common suffix like <code>Time</code> for names of properties representing times? (e.g. the [Appium event timings API](http://appium.io/docs/en/advanced-concepts/event-timings/) uses the suffix <code>Time</code> but represents times using “JS timestamps”, presumably milliseconds as provided by [`Date.now`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/now), the [AWS Device Farm Get Job endpoint](https://docs.aws.amazon.com/en_pv/devicefarm/latest/APIReference/API_GetJob.html) uses no suffix, for example "<code>started</code>", the [SauceLabs Get Tunnel endpoint](https://wiki.saucelabs.com/display/DOCS/Tunnel+Methods#TunnelMethods-GetTunnel) uses no suffix, for example "<code>last_connected</code>". If we do use a suffix, should we specify the units, for example <code>endedTimeMillis</code>?

I propose how it is currently spec'ed with `endedTime` and `requestedTime`.

## Questions:
- @shs96c it is not quite clear for me how the user receives the `sessionId` after the user has fetched a created async job, what am I missing?

***
<a href="https://pr-preview.s3.amazonaws.com/christian-bromann/webdriver/pull/1552.html" title="Last updated on Oct 28, 2020, 3:47 PM UTC (afd7aee)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1552/99f7ced...christian-bromann:afd7aee.html" title="Last updated on Oct 28, 2020, 3:47 PM UTC (afd7aee)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1341,6 +1341,14 @@ when it receives a particular <a>command</a>.
  </tr>
 
  <tr>
+  <td><dfn>unknown session</dfn>
+  <td>404
+  <td><code>unknown session</code>
+  <td>The requested <a>sessionId</a> did not match 
+    with any <a>sessionId</a> known by an intermediary node.
+ </tr>
+
+ <tr>
   <td><dfn>unsupported operation</dfn>
   <td>500
   <td><code>unsupported operation</code>
@@ -2231,17 +2239,8 @@ with a "<code>moz:</code>" prefix:
 </tr>
 
 <tr>
-<td><dfn>Capabilities request</dfn>
-<td>Empty JSON <a>Object</a>
-<td>"<code>capabilitiesRequest</code>"
-<td>
-<td>
-<td>The capabilities request.
-</tr>
-
-<tr>
 <td><dfn>Session creation timeout</dfn>
-<td>600,000
+<td>
 <td>"<code>timeout</code>"
 <td>
 <td>✓
@@ -2720,11 +2719,12 @@ If the creation fails, a <a>session not created</a> <a>error</a> is returned.
 
       <dt>"<code>capabilities</code>"
       <dd>The capabilities of the <a>current session</a>.
-
-      <dt>"<code>capabilitiesRequest</code>"
-      <dd>The capabilities request for the <a>current session</a>.
     </dl>
   </dl>
+
+  <li><p>If the <a>session id</a> is not matching with any
+    of the <var>session id</var> in the list of <a>active sessions</a>,
+    return <a>error</a> with <a>error code</a> <a>unknown session</a>.
 
   <li><p>Return success with data <var>body</var>.
 </section> <!-- /Get Session -->
@@ -2746,23 +2746,20 @@ If the creation fails, a <a>session not created</a> <a>error</a> is returned.
 <p>The <a>remote end</a> steps are:
 
 <ol>
- <li><p>Let <var>session creation requested time</var> be the current time
-  in milliseconds since the Epoch as an integer.
-
  <li><p>Let <var>timeout</var> be the result of <a>getting the
   property</a> named <var>timeout</var> from <var>parameters</var>.
 
   <ol>
    <li><p>If <a>maximum timeout for async session creation</a> is an <a>integer</a>,
     and <var>timeout</var> is undefined,
-    let <var>timeout</var> be <a>min</a>(600,000, <a>maximum timeout for async session creation</a>).
+    let <var>timeout</var> be <a>min</a>(an implementation-defined maximum timeout, <a>maximum timeout for async session creation</a>).
     <li><p>If <a>maximum timeout for async session creation</a> is <a><code>null</code></a>,
     and <var>timeout</var> is undefined,
-    let <var>timeout</var> be 600,000</a>.
+    let <var>timeout</var> be an implementation-defined maximum timeout</a>.
    <li><p>If <var>timeout</var> is <a><code>null</code></a>,
     let <var>timeout</var> be <a>maximum timeout for async session creation</a>.
    <li><p>If <a>maximum timeout for async session creation</a> is an <a>integer</a>,
-    and less than <var>timeout</var>,
+    and less than <var>maximum timeout for async session creation</var>,
     return <a>error</a> with <a>error code</a> <a>invalid timeout</a>.
    <li><p>If <a>timeout</a> is not <a><code>null</code></a> or an <a>integer</a>,
     return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
@@ -2807,9 +2804,6 @@ If the creation fails, a <a>session not created</a> <a>error</a> is returned.
 
     <dt>"<code>events</code>"
     <dd>An empty JSON <a>List</a>.
-
-     <dt>"<code>data</code>"
-     <dd>A JSON <a>Object</a> with implementation-specific data.
    </dl>
 
   <li>Add the <var>session creation job</var> to the list of
@@ -2913,7 +2907,7 @@ If the creation fails, a <a>session not created</a> <a>error</a> is returned.
 
  <li><p>If <var>session creation job</var> has a <a data-lt="session creation status">status</a> of
   "<code>cancelled</code>" or "<code>failed</code>",
-  return <a>error</a> with <a>error code</a> <a>invalid session creation job id</a>.
+  return success with data <a><code>null</code></a>.
 
  <li><p>If <var>session creation job</var> has a <a data-lt="session creation status">status</a> of
   "<code>created</code>":
@@ -2922,7 +2916,7 @@ If the creation fails, a <a>session not created</a> <a>error</a> is returned.
    <li>Set the <a data-lt="session creation status">status</a> to "<code>cancelled</code>".
 
    <li>If <var>session creation job</var>’s job <a>session ID</a> references an <a>active session</a>,
-   schedule a background job to <a>try</a> to <a>close the session</a>.
+   <a>try</a> to <a>close the session</a>
   </ol>
 
  <li><p>If <var>session creation job</var> has a <a data-lt="session creation status">status</a> of
@@ -2932,15 +2926,12 @@ If the creation fails, a <a>session not created</a> <a>error</a> is returned.
   <ol>
    <li>Set <var>session creation job</var>’s <a data-lt="session creation status">status</a> to "<code>cancelled</code>".
    <li>Set <var>session creation job</var>’s <a data-lt="session creation ended time">ended time</a> to the current <a>millisecond timestamp</a>.
-   <li>Schedule a background job to cancel the session creation.
+   <li><a>Try</a> to <a>close the session</a>.
   </ol>
 
  <li><p>Return success with data <a><code>null</code></a>.
 </ol>
 
-<div class="issue" title="Cancel session creation">
- <p>Define how to cancel session creation.</p>
-</div>
 </section> <!-- /Cancel New Session -->
 
 <section>
--- a/index.html
+++ b/index.html
@@ -194,6 +194,10 @@ as described by The Open Group Base Specifications Issue 7
 An <dfn>integer</dfn> is a <a>Number</a> that is unchanged
 under the <a>ToInteger</a> operation.
 
+<p>A <dfn>millisecond timestamp</dfn> is an <a>Integer</a> that
+approximates the number of milliseconds that have elapsed since the
+<a>Unix Epoch</a>.
+
 <p>
 The <dfn>initial value</dfn> of an ECMAScript property
 is the value defined by the platform for that property,
@@ -681,6 +685,30 @@ when it receives a particular <a>command</a>.
  </tr>
 
  <tr>
+  <td>GET</td>
+  <td>/session/{session id}</td>
+  <td><a>Get Session</a></td>
+ </tr>
+ 
+ <tr>
+  <td>POST</td>
+  <td>/session/async</td>
+  <td><a>Async New Session</a></td>
+ </tr>
+ 
+ <tr>
+  <td>GET</td>
+  <td>/session/async/{<var>session creation job id</var>}{/<var>event</var>}
+  <td><a>Async New Session Status</a></td>
+ </tr>
+ 
+ <tr>
+  <td>DELETE</td>
+  <td>/session/async/{<var>session creation job id</var>}
+  <td><a>Cancel New Session</a></td>
+ </tr>
+
+ <tr>
   <td>DELETE</td>
   <td>/session/{<var>session id</var>}</td>
   <td><a>Delete Session</a></td>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -194,6 +194,10 @@ <h2>Terminology</h2>
 An <dfn>integer</dfn> is a <a>Number</a> that is unchanged
 under the <a>ToInteger</a> operation.
 
+<p>A <dfn>millisecond timestamp</dfn> is an <a>Integer</a> that
```
shs96c, line 197: `Date.prototype.toJSON` delegates to [Date.prototype.toISOString](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString) I suggest we lean into that, rather than epoch millis

index.html
```diff
@@ -1144,13 +1172,29 @@ <h3>Errors</h3>
...
   <td><code>invalid selector</code>
   <td>Argument was an invalid selector.
  </tr>
 
+ <tr>
+  <td><dfn>invalid session creation job id</dfn>
+  <td>404
+  <td><code>invalid session creation job id</code>
+  <td>Occurs if the given <a>session creation job id</a> is not in the list of <a>session creation jobs</a>,
+   or is in the list but has a <a data-lt="session creation status">status</a> of
```
shs96c, line 1194: While I can see the reasoning for this, it might be useful to be able to get the `cancelled` or `failed` event for inspection.

index.html
```diff
@@ -2139,8 +2190,203 @@ <h2>Sessions</h2>
...
+ there is no timeout beyond implementation limits.
+</tr>
+
+<tr>
+<td><dfn>Session creation requested time</dfn>
+<td>
+<td>"<code>requestedTime</code>"
+<td>
+<td>
+<td>When the session creation was requested as a <a>millisecond timestamp</a>.
```
shs96c, line 2261: As a `Date` would be more appropriate. Implementations would be free to store this as a millisecond timestamp.

index.html
```diff
@@ -2139,8 +2190,203 @@ <h2>Sessions</h2>
...
+</tr>
+
+<tr>
+<td><dfn>Session creation ended time</dfn>
+<td>
+<td>"<code>endedTime</code>"
+<td>✓
+<td>
+<td>When the job ended in success or failure as a <a>millisecond timestamp</a>.
```
shs96c, line 2270: As above.

index.html
```diff
@@ -2139,8 +2190,203 @@ <h2>Sessions</h2>
...
+<td>A string representation of a <a>UUID</a> used to uniquely identify this event.
+</tr>
+
+<tr>
+<td><dfn>Event time</dfn>
+<td>
+<td>"<code>time</code>"
+<td>
+<td>
+<td>Event time as a <a>millisecond timestamp</a>.
```
shs96c, line 2328: `Date`

index.html
```diff
@@ -2449,6 +2695,254 @@ <h3 id=new-session><dfn data-lt="new sessions|creating a new session">New Sessio
...
+ <li><p>Let <var>body</var> be a JSON <a>Object</a> initialised with:
+  <dl>
+    <dl>
+      <dt>"<code>sessionId</code>"
+      <dd>The <a>session id</a> of the <a>current session</a>.
+
+      <dt>"<code>capabilities</code>"
+      <dd>The capabilities of the <a>current session</a>.
+
+      <dt>"<code>capabilitiesRequest</code>"
```
shs96c, line 2724: Drop `capabilitiesRequest`

The shape of this object should match the shape of the object returned by `New Session`. The local end is aware of the capabilities that were requested, and (more importantly) this might be absolutely gigantic if someone has include one or more browser profiles.

index.html
```diff
@@ -2449,6 +2695,254 @@ <h3 id=new-session><dfn data-lt="new sessions|creating a new session">New Sessio
...
+      <dd>The capabilities request for the <a>current session</a>.
+    </dl>
+  </dl>
+
+  <li><p>Return success with data <var>body</var>.
```
shs96c, line 2729: You need to add a response for what to do if the `session id` is unknown.

index.html
```diff
@@ -2449,6 +2695,254 @@ <h3 id=new-session><dfn data-lt="new sessions|creating a new session">New Sessio
...
+ <tr>
+  <td>POST</td>
+  <td>/session/async</td>
+ </tr>
+</table>
+
+<p>The <a>remote end</a> steps are:
+
+<ol>
+ <li><p>Let <var>session creation requested time</var> be the current time
```
shs96c, line 2749: If we use dates instead of millisecond timestamps, you can drop the next line.

index.html
```diff
@@ -2449,6 +2695,254 @@ <h3 id=new-session><dfn data-lt="new sessions|creating a new session">New Sessio
...
+  in milliseconds since the Epoch as an integer.
+
+ <li><p>Let <var>timeout</var> be the result of <a>getting the
+  property</a> named <var>timeout</var> from <var>parameters</var>.
+
+  <ol>
+   <li><p>If <a>maximum timeout for async session creation</a> is an <a>integer</a>,
+    and <var>timeout</var> is undefined,
+    let <var>timeout</var> be <a>min</a>(600,000, <a>maximum timeout for async session creation</a>).
```
shs96c, line 2758: We may want implementations to be able to set a lower session creation timeout. While a minute is a sensible default for Sauce, on a local machine with a local driver, I'd expect async session creation to be significantly faster.

index.html
```diff
@@ -2449,6 +2695,254 @@ <h3 id=new-session><dfn data-lt="new sessions|creating a new session">New Sessio
...
+    <li><p>If <a>maximum timeout for async session creation</a> is <a><code>null</code></a>,
+    and <var>timeout</var> is undefined,
+    let <var>timeout</var> be 600,000</a>.
+   <li><p>If <var>timeout</var> is <a><code>null</code></a>,
+    let <var>timeout</var> be <a>maximum timeout for async session creation</a>.
+   <li><p>If <a>maximum timeout for async session creation</a> is an <a>integer</a>,
+    and less than <var>timeout</var>,
```
shs96c, line 2765: It'll be better to use `maximum timeout for async session creation`: presumably this occurs when the local end sets some ridiculous timeout and the remote end has something sensible in place.

index.html
```diff
@@ -2449,6 +2695,254 @@ <h3 id=new-session><dfn data-lt="new sessions|creating a new session">New Sessio
...
+    <dd><var>timeout</var>’s value.
+
+    <dt>"<code>requestedTime</code>"
+    <dd><var>session creation requested time</var>’s value.
+
+    <dt>"<code>events</code>"
+    <dd>An empty JSON <a>List</a>.
+
+     <dt>"<code>data</code>"
+     <dd>A JSON <a>Object</a> with implementation-specific data.
```
shs96c, line 2812: It's not clear what this `data` is meant to be used for. A note explaining this would be useful.

index.html
```diff
@@ -2449,6 +2695,254 @@ <h3 id=new-session><dfn data-lt="new sessions|creating a new session">New Sessio
...
+
+      <ol>
+       <li><p>If <var>found</var> is true, then break.
+
+       <li><p>Let <var>event id</var> be the value of <a>getting a property</a> named <var>eventId</var> from <var>event object</var>.
+
+       <li><p>If <var>event</var> equals <var>event id</var>,
+        set <var>found</var> to true.
+
+       <li><p>Prepend <var>event object</var> to <var>filtered events</var>.
```
shs96c, line 2866: We only ever collect a single event from this loop. Why do we maintain the list?

index.html
```diff
@@ -2449,6 +2695,254 @@ <h3 id=new-session><dfn data-lt="new sessions|creating a new session">New Sessio
...
+      </ol>
+
+      <li><p>If <var>found</var> is false,
+       return <a>error</a> with <a>error code</a> <a>invalid event id</a>.
```
shs96c, line 2870: I think it's legit to just send the empty list as it just means that there have been no events, right?

index.html
```diff
@@ -2449,6 +2695,254 @@ <h3 id=new-session><dfn data-lt="new sessions|creating a new session">New Sessio
...
+<ol>
+ <li><p>Let <var>session creation job</var> be the job in <a>session creation jobs</a>
+  with a <a>session creation job id</a> matching <var>session creation job id</var>.
+
+ <li><p>If <var>session creation job</var> is <a><code>null</code></a>,
+  return <a>error</a> with <a>error code</a> <a>invalid session creation job id</a>.
+
+ <li><p>If <var>session creation job</var> has a <a data-lt="session creation status">status</a> of
+  "<code>cancelled</code>" or "<code>failed</code>",
+  return <a>error</a> with <a>error code</a> <a>invalid session creation job id</a>.
```
shs96c, line 2916: Seems like the user-observable response would be the same as having deleted the job, no? If that's the case, should we consider returning "success with data null"?

index.html
```diff
@@ -2449,6 +2695,254 @@ <h3 id=new-session><dfn data-lt="new sessions|creating a new session">New Sessio
...
+
+ <li><p>If <var>session creation job</var> has a <a data-lt="session creation status">status</a> of
+  "<code>created</code>":
+
+  <ol>
+   <li>Set the <a data-lt="session creation status">status</a> to "<code>cancelled</code>".
+
+   <li>If <var>session creation job</var>’s job <a>session ID</a> references an <a>active session</a>,
+   schedule a background job to <a>try</a> to <a>close the session</a>.
```
shs96c, line 2925: "schedule a background job" is an implementation-specific detail. I can imagine a local driver doing this serially.

index.html
```diff
@@ -2449,6 +2695,254 @@ <h3 id=new-session><dfn data-lt="new sessions|creating a new session">New Sessio
...
+  </ol>
+
+ <li><p>If <var>session creation job</var> has a <a data-lt="session creation status">status</a> of
+  "<code>queuing</code>" or "<code>creating</code>"
+  and its <a>session ID</a> references an <a>active session</a>:
+
+  <ol>
+   <li>Set <var>session creation job</var>’s <a data-lt="session creation status">status</a> to "<code>cancelled</code>".
+   <li>Set <var>session creation job</var>’s <a data-lt="session creation ended time">ended time</a> to the current <a>millisecond timestamp</a>.
+   <li>Schedule a background job to cancel the session creation.
```
shs96c, line 2935: Again, avoid "background job". We also don't seem to remove the `session creation job` from the queue....

shs96c (changes requested):
This looks useful, and a good start. Thank you for picking this up.

index.html
```diff
@@ -2139,8 +2190,203 @@ <h2>Sessions</h2>
...
+<td>Time-ordered list of session creation events.
+</tr>
+
+<tr>
+<td><dfn>Session creation data</dfn>
+<td></a>
+<td>"<code>data</code>"
+<td>✓
+<td>
+<td>Object containing implementation-defined data associated with this job.
```
christian-bromann, line 2297: We should add a not that this will be not further specified as it is implementation specific.

AutomatedTester:
I can't see a link to tests so adding requires test label

jgraham:
I'd like to propose we start a new spec for this. The concern is that this doesn't seem like something browser vendors are likely to implement directly but something we could standardise with the implementation target being intermediary nodes. I know that's overhead, but unless there's a compelling reason that we actually need this in browser drivers I think it makes sense to seperate out so it's clear who the feature is aimed at.

shs96c:
I'm fine with us speccing this out for intermediate nodes alone, but it should be specced out.

christian-bromann:
@shs96c I addressed the comments and currently have only 2 remaining issue that I need input for:

1. How is the WebDriver use suppose to receive the session id from once the session creation job moved to `status` `created`?
2. In your original draft you defined an endpoint to fetch the session creation job (`/session/async/{session creation job id}`) as well as specific events via `/session/async/{session creation job id}/{event}` where `event` is an event id that a user can receive from the events property of the session creation job. According to the spec it would only find one event that is matching the event id provided by the url path, you commented:

    > I think it's legit to just send the empty list as it just means that there have been no events, right?

    which assumes it is able to fetch multiple events. I think the reasoning behind not allowing to just fetch all events is the same we have with logs however I don't think implementers will have as many job creation events as logs on a page. How about instead of fetching a specific event, to just have `/session/async/{session creation job id}/events` that fetches all events happening during job creation.

What do you all think?

christian-bromann:
Closing in favor of #1561

PR #1561, opened 2020-11-05
Add Get Session and Async New Session

christian-bromann:
This is the light version of #1552 that keeps implementation effort for browser vendors low while enabling intermediary nodes the ability to create async sessions in a lightweight way. There are two new endpoints being added in this patch:

1. __Get Session__
   Returns the capabilities of an active session. This is also very useful for endpoint nodes
   in the scenario where a clients wants to attach to an active session receive the capabilities
   it has attached to. This information is not necessarily given for client implementations.

2. __Async New Session__
   Endpoint nodes run the same session creation steps as in New session but return only the
   session id. Intermediary nodes keep a list of "session creation jobs" that maps "session
   creation job ids" to session ids of active sessions.

Now the use case is a user can now always make an async session by making a request to `[POST] /session/async` which returns a session creation id:

```json
{
  "sessionCreationId": "<uuid>"
}
```

With this id the "Get Session" endpoint can be requested, which returns with `unknown session` as long as the intermediary node has not yet initiated the session at the endpoint node. Once that happen a response with the following data will be returned:

```json
{
  "sessionId": "<uuid of active session>"
  "capabilities: { ... }
}
```

The client can then continue making WebDriver request with the session id it received from this response.

I am happy to start working on the web platform tests once there is a general agreement on this approach.

***
<a href="https://pr-preview.s3.amazonaws.com/webdriver/webdriver/pull/1561.html" title="Last updated on Nov 5, 2020, 3:49 PM UTC (d30fb4c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1561/5ab304e...webdriver:d30fb4c.html" title="Last updated on Nov 5, 2020, 3:49 PM UTC (d30fb4c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -681,9 +681,21 @@ when it receives a particular <a>command</a>.
  </tr>
 
  <tr>
+  <td>GET</td>
+  <td>/session/{session id}</td>
+  <td><a>Get Session</a></td>
+ </tr>
+
+ <tr>
   <td>DELETE</td>
   <td>/session/{<var>session id</var>}</td>
   <td><a>Delete Session</a></td>
+ </tr>
+
+ <tr>
+  <td>POST</td>
+  <td>/session/async</td>
+  <td><a>Async New Session</a></td>
  </tr>
 
  <tr>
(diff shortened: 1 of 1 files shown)


christian-bromann:
any comments @shs96c @AutomatedTester ?

Added WPT tests in web-platform-tests/wpt#26490

AutomatedTester (approved):

PR #1582, opened 2021-04-19
Editorial: migrate from "context object" to "this"

foolip:
The term "context object" is being removed from DOM:
https://github.com/whatwg/dom/pull/973

This also changes a reference to <a> elements to the string "a", since
querySelectorAll's argument is a string. This may have unintended side
effects of matching <svg:a> and "a" elements in other namespaces.

***
<a href="https://pr-preview.s3.amazonaws.com/foolip/webdriver/pull/1582.html" title="Last updated on Apr 19, 2021, 12:57 PM UTC (25e9da1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1582/35df53a...foolip:25e9da1.html" title="Last updated on Apr 19, 2021, 12:57 PM UTC (25e9da1)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4140,7 +4140,7 @@ argument <var>reference</var>, run the following steps:
 
 <p>An <a>element</a> <dfn>is stale</dfn>
  if its <a>node document</a> is not the <a>active document</a>
- or if its <a>context object</a> is not <a>connected</a>.
+ or if it is not <a>connected</a>.
 
 <p>To <dfn data-lt="scrolls into view|scrolled into view">scroll into view</dfn>
  an <var><a>element</a></var>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -4554,10 +4554,9 @@ <h5>Link text</h5>
  the following steps need to be completed:
 
 <ol>
- <li><p>Let <var>elements</var> be the result of
-  calling <a>querySelectorAll</a>, with argument <a><code>a</code>
-  elements</a>, with the <a>context object</a> equal to the
-  <var>start node</var>. If this throws an exception,
+ <li><p>Let <var>elements</var> be the result of calling
+  <a>querySelectorAll</a> with <var>start node</var> as <a>this</a>
+  and "<code>a</code>" as the argument. If this throws an exception,
```
foolip, line 4559: This and the following similar change could be breaking what is intended by the spec. Unfortunately I can't find any tests in WPT involving SVG, so it would be some work to test if current WebDriver implementations would filter out non-HTMLAnchorElement values as the original text might suggest (but not correctly define).

Leaving this comment to see if others have thoughts before I dig more.

jgraham: I would expect it would only consider elements that represent links; getting some `foo:a` elements in a made up namespace. However `svg:a` is also a link, so that seems like it perhaps ought to be considered? idk what existing implementations do here.

foolip: I tried to find out, and it seems like ChromeDriver implements this using a large snippet of JS from Selenium. I think this is what ends up being used:
https://github.com/SeleniumHQ/selenium/blob/64447d4b03f6986337d1ca8d8b6476653570bcc1/javascript/atoms/locators/link_text.js#L76

So that boils down to `root.querySelectorAll("a")` and I can't see anything that tries to filter to `HTMLAnchorElement` and `SVGAElement`. So I guess what I wrote here ends up matching at least chromedriver better than the original text.

I guess this should be tested though...

jgraham: https://searchfox.org/mozilla-central/source/testing/marionette/element.js#452 looks like it boils down to the same thing. I'm prepared to say this is stilly an unintentional, but probably not worth changing at this point.

jgraham (approved):

PR #1585, opened 2021-05-07
CI: Use GitHub Actions with w3c/spec-prod

sideshowbarker:
For CI, drop Travis and switch to using GitHub Actions with https://github.com/w3c/spec-prod/


jgraham (approved):
I don't know a lot about this, but it looks plausible enough to try.

PR #1608, opened 2021-07-14
Refactor session handling for better integration with BiDi

jgraham:
This is a large change, but it's almost entirely refactoring.

The main aim here is to split the session concept so that there's a base "session" which holds the session id and state that's appropriate to both WebDriver classic and BiDi. Then, we define a "HTTP Session" as one that has the HTTP Flag set and use that to compose-in any state that's only used in the classic HTTP-based protocol.

In addition, we rework the way state is passed around the spec so that we are using an explicit session object, rather than making use of a global "current session". This makes things clearer if we allow the possibility of multiple non-HTTP sessions.

There is also a bit of cleanup in the way we pass other variables around, and in the capabilites processing.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/pull/1608.html" title="Last updated on Sep 16, 2021, 1:58 PM UTC (8717fbe)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1608/d8556ca...8717fbe.html" title="Last updated on Sep 16, 2021, 1:58 PM UTC (8717fbe)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/pull/1608.html" title="Last updated on Feb 26, 2024, 4:09 PM UTC (b70ad7e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1608/a04dbf7...b70ad7e.html" title="Last updated on Feb 26, 2024, 4:09 PM UTC (b70ad7e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -274,9 +274,10 @@ It is acknowledged that this is complementary to the Evil Bit [[RFC3514]].
   above the level of the wire protocol.
 
  <dt><dfn data-lt="remote ends|remote">Remote end</dfn>
- <dd>The remote end hosts the server side of the <a href=#protocol>protocol</a>.
-  Defining the behavior of a remote end in response to the WebDriver protocol
-  forms the largest part of this specification.
+ <dd>The remote end hosts the server side of
+  the <a href=#protocol>protocol</a>.  Defining the behavior of
+  a <a>remote end</a> in response to the WebDriver protocol forms the
+  largest part of this specification.
 </dl>
 
 <p>For <a>remote ends</a> the standard defines two broad conformance
@@ -306,19 +307,17 @@ It is acknowledged that this is complementary to the Evil Bit [[RFC3514]].
   of the wire protocol.
 
 <p>
-The <dfn>readiness state</dfn> of a <a>remote end</a>
-indicates whether it is free to accept new connections.
-It must be false if the maximum <a>active sessions</a>
-is equal to the length of the list of <a>active sessions</a>,
-or if the node is an <a>intermediary node</a>
-and is known to be in a state in which attempting to create <a>new sessions</a> would fail.
-In all other cases it must be true.
+The <dfn>readiness state</dfn> of a <a>remote end</a> indicates
+whether it is free to accept new connections.  It must be false if the
+implementation is an <a>endpoint node</a> and the list of <a>active
+HTTP sessions</a> is not empty, or otherwise if the <a>remote end</a> is known to
+be in a state in which attempting to create <a>new sessions</a> would
+fail. In all other cases it must be true.
 
-<p class=example>If the <a>intermediary node</a>
- is a multiplexer that manages
- multiple <a data-lt="endpoint node">endpoint nodes</a>,
- this might indicate its ability to purvey more <a>sessions</a>,
- for example if it has hit its maximum capacity.
+<p class=example>If the <a>intermediary node</a> is a multiplexer that
+ manages multiple <a data-lt="endpoint node">endpoint nodes</a>, this
+ might indicate its ability to purvey more <a>sessions</a>, for
+ example if it has hit its maximum capacity.
 </section> <!-- /Nodes -->
 
 <section>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2175,28 +2154,33 @@ <h2>Sessions</h2>
...
+ <dfn>active sessions</dfn>, which is a list of all <a>session</a>s
+ that are currently started.
+
+<p>A <a>remote end</a> has an associated list of
+ <dfn>active HTTP sessions</dfn>, which is a list of all <a>HTTP
+ session</a>s that are currently started.
+
+<p class=note>The limitation of a single HTTP session for <a>endpoint
+node</a>s means that the first entry in the list of <a>active HTTP
+sessions</a> will be the only entry.
```
whimboo, line 2191: Given that we allow a single active HTTP session only do we actually have to carry a list of HTTP sessions?

index.html
```diff
@@ -2205,47 +2189,56 @@ <h2>Sessions</h2>
...
-<p>A <a>session</a> has an associated <dfn>request queue</dfn> which is a
- [=queue=] of [=requests=] that are currently awaiting
- processing.
+<p>An <a>endpoint node</a> has an associated <dfn>accept insecure
+ TLS</dfn> flag that indicates whether untrusted or self-signed TLS
+ certificates are treated as trusted. The default value of the flag is
+ false if the endpoint doesn't support accepting insecure TLS
+ connections, or unset otherwise.</p>
+
+<p>An <a>endpoint node</a> has an associated <dfn>has proxy
```
whimboo, line 2237: s/has proxy configuration/proxy configuration/

index.html
```diff
@@ -2205,47 +2189,56 @@ <h2>Sessions</h2>
...
 
 <p>When asked to <dfn>close the session</dfn>,
  a <a>remote end</a> must take the following steps:
 
 <ol>
+
+ <li><p>If <var>session</var>'s <a>HTTP flag</a> is set,
+ remove <var>session</var> from <a>active HTTP sessions</a>.
+
+ <li><p>Remove <var>session</var> from <a>active sessions</a>.
```
whimboo, line 2339: I assume that the BiDi related action here would be to check if a BiDi session for `session` is active when one get removed from `active sessions`, and appropriately also removes it from `active BiDi sessions`.

What happens when BiDi closes the session? Do we have to add an action here for the HTTP spec like `If a session gets removed from active sessions, which has the `HTTP` flag set, remove the session from `active HTTP sessions`?

index.html
```diff
@@ -2358,9 +2349,11 @@ <h3 id=new-session><dfn data-lt="new sessions|creating a new session">New Sessio
 variables</var> and <var>parameters</var> are:
 
 <ol>
-  <li><p>If the <a>maximum active sessions</a> is equal to
-  the length of the list of <a>active sessions</a>,
-  return <a>error</a> with <a>error code</a> <a>session not created</a>.
+  <li><p>If the implementation is an <a>endpoint node</a>, and the
+  list of <a>active HTTP sessions</a> is not empty, or otherwise if
+  the implementation is unable to start an additional session,
```
whimboo, line 2454: Perfect addition given that there can be a lot of other issues during session creation.

index.html
```diff
@@ -2205,47 +2189,56 @@ <h2>Sessions</h2>
...
+across all sessions.</p>
 
-<p>A <a>session</a> has an associated <dfn>request queue</dfn> which is a
- [=queue=] of [=requests=] that are currently awaiting
- processing.
+<p>An <a>endpoint node</a> has an associated <dfn>accept insecure
+ TLS</dfn> flag that indicates whether untrusted or self-signed TLS
+ certificates are treated as trusted. The default value of the flag is
+ false if the endpoint doesn't support accepting insecure TLS
+ connections, or unset otherwise.</p>
```
whimboo, line 2235: So when do we actually allow to set this global state? I assume the first created session will have to install the handler, and when the last session gets closed it will be removed?

Same question applies for other global states.

index.html
```diff
@@ -2229,7 +2238,96 @@ <h3>Global State</h3>
...
+     "<code>proxy</code>" from <var>capabilities</var> and run the
+     substeps of the first matching statement:
+
+       <dl class=switch>
+         <dt><var>proxy</var> is a <a>proxy configuration</a> object</dt>
+         <dd><p>Take implementation-defined steps to set the user agent proxy
+             using the extracted <var>proxy</var> configuration. If the
+             defined proxy cannot be configured return <a>error</a> with
+             <a>error code</a> <a>session not created</a>. Otherwise
+             set the <a>has proxy configuration</a> flag to true.</dd>
```
whimboo, line 2263: What if the flag is already `true`? Shouldn't we return early? I don't think that we should override the proxy settings as already set by a formerly created session.

index.html
```diff
@@ -2229,7 +2238,96 @@ <h3>Global State</h3>
...
+         <dd><a>Set a property</a> of <var>capabilities</var> with name
+           "<code>proxy</code>" and a value that is a new JSON <a>Object</a>.</dd>
+       </dl>
+   </li>
+
+  <li><p>If <var>capabilites</var> has a property named
+  "<code>acceptInsecureCerts</code>", set the <a>remote
+  end</a>'s <a>accept insecure TLS</a> flag to the result
+  of <a>getting a property</a> named
+  "<code>acceptInsecureCerts</code>" from <var>capabilities</var>.
```
whimboo, line 2275: As I read above we already configure the proxy, and here we only set the flag but don't install the handler. Shouldn't the proxy configuration happen later similar to when we set the insecure cert handler?

index.html
```diff
@@ -2371,99 +2469,30 @@ <h3 id=new-session><dfn data-lt="new sessions|creating a new session">New Sessio
...
 
  <li><p>Let <var>body</var> be a JSON <a>Object</a> initialised with:
 
   <dl>
    <dt>"<code>sessionId</code>"
-   <dd><var>session id</var>
+   <dd><var>session</var>'s <a>session ID</a>.
 
    <dt>"<code>capabilities</code>"
    <dd><var>capabilities</var>
```
whimboo, line 2493: `<var>session</var>'s capabilities`?

index.html
```diff
@@ -2371,99 +2469,30 @@ <h3 id=new-session><dfn data-lt="new sessions|creating a new session">New Sessio
...
-   <li><p>Apply changes to the user agent for any implementation-defined capabilities
-    selected during the <a>capabilities processing</a> step.
-  </ol>
-
- <li><p>Set the <a>webdriver-active flag</a> to true.</p>
-
- <li><p><a>Set the current top-level browsing context</a>
-   for <var>session</var> with the <a>top-level browsing context</a>
+ <li><p>Set the <a>current top-level browsing context</a>
+   of <var>session</var> to the <a>top-level browsing context</a>
```
whimboo, line 2497: should we add: `of the currently selected tab`? Otherwise it would be unclear in case the browser starts up with multiple tabs already open.

index.html
```diff
@@ -2346,9 +2346,9 @@ <h3>Global State</h3>
      <li><p>If the list of <a>active sessions</a> is empty, set
      the <a>webdriver-active flag</a> to false.
 
-     <li><p>An <a>endpoint node</a> must <a>close</a> any <a>top-level
-      browsing contexts</a> associated with the <a>session</a>,
-      without <a>prompting to unload</a>.
+     <li><p>Optionally, <a>close</a> any <a>top-level browsing contexts</a>
```
whimboo, line 2350: Should we clarify what `optionally` means? Is it the case when there are still entries listed in `sessions`?

index.html
```diff
@@ -6995,19 +7129,21 @@ <h3><dfn>Get Named Cookie</dfn></h3>
...
- <li><p><a>Handle any user prompts</a>, and return its value if it is an <a>error</a>.
+ <li><p><a>Try</a> to <a>handle any user prompts</a>
+  with <var>session</var>.
 
- <li><p>If the <a>url variable</a> <var>name</var>
-  is equal to a <a>cookie</a>’s <a>cookie name</a>
-  amongst <a>all associated cookies</a>
-  of the <a>current browsing context</a>’s <a>active document</a>,
-  return <a>success</a> with the <a>serialized cookie</a> as data.
+ <li><p>If the <var>URL variables</var>' name is equal to
```
shs96c, line 7142: Missing `s` after the apostrophe

shs96c (dismissed):
I think we'll need to be particularly careful to make sure that it's possible to reformulate this spec on top of Bidi, and I've called out a few places where I'm not sure that would be possible with this PR.

index.html
```diff
@@ -2200,57 +2199,161 @@ <h2>Sessions</h2>
...
+            "<code>pageLoadStrategy</code>" and value "<code>normal</code>".</p>
+
+        <li><p>Let <var>strictFileInteractability</var> be the result of
+            getting property "<code>strictFileInteractability</code>"
+            from <var>capabilities</var>. If <var>strictFileInteractability</var>
+            is a boolean, set <a>session</a>’s <a>strict file interactability</a> to
+            <var>strictFileInteractability</var>. Otherwise set
+            <a>session</a>’s <a>strict file interactability</a> to false.</p></li>
+
+        <li><p>If <var>capabilities</var> has a property with the key "<code>timeouts</code>":
```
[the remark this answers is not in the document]
jgraham: So far we don't have any use case for timeouts in BiDi (the client can always implement timeouts when the protocol is non-blocking, as long as no cleanup is required), and I'm keen to avoid timeouts as global state in BiDi.

index.html
```diff
@@ -2494,11 +2535,12 @@ <h3><dfn>Delete Session</dfn></h3>
...
 </table>
 
-<p>The <a>remote end steps</a> are:
+<p>The <a>remote end steps</a>, given <var>session</var>, <var>URL
+variables</var> and <var>parameters</var> are:
 
 <ol>
- <li><p>If the <a>current session</a> is an <a>active session</a>,
-  <a>try</a> to <a>close the session</a>.
+ <li><p>If <var>session</var> is an <a>active HTTP session</a>,
```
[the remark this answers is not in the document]
jgraham: The point is that if you have two sessions, one BiDi-only and one HTTP+BiDi you shouldn't be able to close the BiDi-only session using an HTTP call. But a HTTP+BiDi session is an HTTP session, so you can close that one with a HTTP call.

jgraham:
I've done a round of fixup here.

One thing that apparently isn't very obvious from the text is how I'm thinking about the session flags. The basic idea was to have a single kind of session object, but let it compose together the HTTP session and BiDi session properties, so that we can have all of the combinations HTTP-only, BiDi-only, HTTP+BiDi, and extend to anything new we want in the future. So the idea was that the flags are used to specify which kind of session a given session is, and knowing that you can tell what properties it must have e.g. if a session is a HTTP session it must have a timeouts property that you can access, but a BiDi session doesn't. The idea wasn't to group together _all_ the oolean properties of a session into a single flags object.

index.html
```diff
@@ -1781,8 +1786,8 @@ <h3>Processing capabilities</h3>
   to an indexed property in <var>all first match capabilities</var>:
   <ol>
    <li><p>Let <var>validated capabilities</var> be the result
-    of <a>trying</a> to <a>validate capabilities</a> with argument
-    <var>first match capabilities</var>.
+    of <a>trying</a> to <a>validate capabilities</a> with arguments
+    <var>first match capabilities</var> and <var>flags</var>.
```
lutien, line 1790: We pass here `flags` to `validate capabilities` algorithm, but it looks like it's not used there

lutien: On a second thought, maybe we actually want to update the validation algorithm to throw an invalid argument error if let's say classic capabilities are passed, but flags don't contain `http`?

jgraham: I've updated it so we only process those capabilities for the `http` case; that should cause the session to be rejected if they are passed to BiDi.

whimboo:
Hi @shs96c. Would you mind following up on some of the replies from @jgraham? Not sure if you think that those issues are still relevant or can be closed.  It would be nice to eventually get this massive PR merged. Thanks.

OrKoN:
it looks like there are no remaining unanswered questions and this PR needs a rebase?

OrKoN (approved):
LGTM (@sadym-chromium PTAL)

index.html
```diff
@@ -631,10 +629,10 @@ <h3>Routing requests</h3>
...
   containing each row in the <a>table of endpoints</a>.
 
- <li><p>Remove each entry from <var>endpoints</var>
-  for which the concatenation of the <a>URL prefix</a>
-  and the entry’s <a>URI template</a>
-  does not match <var>URL</var>’s <a>path</a>.
+ <li><p>Remove each entry from <var>endpoints</var> for which the
+  concatenation of the <a>URL prefix</a> and the entry’s <a>URI
+  template</a> does not have a valid expansion equal to
+  to <var>URL</var>’s <a>path</a>.
```
sadym-chromium, line 635: typo 
```suggestion
  template</a> does not have a valid expansion equal to
  <var>URL</var>’s <a>path</a>.
```

sadym-chromium (approved):

Issue #1578, opened 2021-03-12
Specification should explicitely allow the validation of the Content-Type of incoming requests


randomstuff:
The specification does not specify the `Content-Type` used for WebDriver requests. Even if the content of the request body is JSON, it appears to be valid (according to the specification) to send the request using `Content-Type: application/x-www-form-urlencoded` (for example).

Relevant snippets:

> If request’s method is POST:  Let parse result be the result of *parsing as JSON* with request’s body as the argument.

Where “parsing as JSON” is defined as:

> The result of JSON deserialization with text is defined as the result of calling JSON.[[Parse]](text). 

No mention that a suitable value for the `Content-Type` should be checked. This is only mandated for responses. This appears to be an involuntary omission in the specification.

Enforcing the value of the request `Content-Type` header is a simple way to prevent [CSRF attacks](https://bugzilla.mozilla.org/show_bug.cgi?id=1648964) against the WebDriver service. However, [it has been argued](https://bugs.chromium.org/p/chromium/issues/detail?id=1100097#c5) that this is a violation of the specification.

Would it be possible to at least suggest the usage of a suitable of `application/json` for requests and allow the servers to enforce this?

For reference, Geckodriver now [rejects](https://phabricator.services.mozilla.com/D81533) `application/x-www-form-urlencoded`, `multipart/form-data` and `text/plain` as a CSRF mitigation.

whimboo:
@AutomatedTester can you please have a look? Thanks.

PR #1634, opened 2021-12-03
Add some security checks when handling a HTTP connection.

jgraham:
Check the Host and Origin headers for the incoming connection to
verify the connection is allowed.

The language is intended to allow the specific behaviour to be largely
implementation defined, whilst recommending a default
behaviour that prevents CSRF-type attacks (reject host headers that
aren't an IP address or the server hostname, reject any requests with
an origin header).

Hopefully adding this text will ensure that implementations consider
the security issues accepting a connection, even though it's not
possible to give precise requirements that apply to all
implementations.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/pull/1634.html" title="Last updated on Dec 3, 2021, 4:36 PM UTC (68f0fb1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1634/1917e89...68f0fb1.html" title="Last updated on Dec 3, 2021, 4:36 PM UTC (68f0fb1)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/pull/1634.html" title="Last updated on Mar 18, 2022, 3:40 PM UTC (a26a826)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1634/cfd3d52...a26a826.html" title="Last updated on Mar 18, 2022, 3:40 PM UTC (a26a826)">Diff</a>

- - -
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1634)

### DIFF
--- a/index.html
+++ b/index.html
@@ -465,11 +465,14 @@ when it receives a particular <a>command</a>.
    <p>Then <a>send an error</a> with <a>error code</a> <a>unknown
   error</a>, and jump to step 1.</p>
 
-  <p class=note>Rejecting connections with unexpected values in the
-  <a>Host header</a> prevents DNS rebinding attacks. Implementations can opt
-  to provide more stringent controls where appropriate, for example
-  only accepting connections when the <var>host</var> value
-  corresponds to a loopback interface [[RFC5735]].</p></li>
+  <p class="note">Rejecting connections with unexpected values in the
+   <a>Host header</a> prevents DNS rebinding attacks. Implementations
+   can opt to provide more stringent controls where appropriate, for
+   example only accepting connections when the <var>host</var> value
+   corresponds to a loopback interface [[RFC5735]]. Further guidance
+   for implementors is given in the <a href="#security">security</a>
+   section.</p>
+  </li>
 
   <li><p>If <var>origin</var> is not undefined and is not identical to
   an <a>Origin header</a> value that the implementation has been
@@ -10494,18 +10497,18 @@ ensuring both privacy and preventing state from bleeding through to the next ses
  and that WebDriver remains disabled
  in publicly consumed versions of the user agent.
 
-<p>To prevent arbitrary machines on the network
- from connecting and creating <a>sessions</a>,
- it is suggested that only connections from
- loopback devices are allowed by default.
-
-<p>The <a>remote end</a> can include
- a configuration option to limit
- the accepted IP range allowed to connect and make requests.
- The default setting for this might be
- to limit connections to the IPv4 localhost
- CIDR range <code>127.0.0.0/8</code>
- and the IPv6 localhost address <code>::1</code>. [[RFC4632]]
+<p>To prevent arbitrary machines on the network from connecting and
+  creating <a>sessions</a>, it is suggested that only connections from
+  loopback devices are allowed by default. However, testing setups
+  commonly put the <a>remote end</a> and <a>local end</a> on different
+  network hosts. Users deploying such a setup are encouraged to
+  restrict access to the remote end to the greatest extent possible,
+  either by restricting network connections to trusted hosts (e.g. in
+  the case of a lab setting, or the remote end running in a containers
+  on the same bridged network), or by routing all connections through
+  an <a>intermediary node</a> that provides authorization and
+  authentication. <a>Remote end</a> implementors are encouraged to
+  provide minimal, opt-in, configuration to support these scenarios.
 
 <p>It is also suggested that user agents
  make an effort to visually distinguish
--- a/index.html
+++ b/index.html
@@ -434,14 +434,15 @@ when it receives a particular <a>command</a>.
 
   <li><p>If <var>request</var> has a <a>Host header</a>,
   let <var>host</var> be the value of that header. Otherwise
-  let <var>host</var> be null.</p></li>
+  let <var>host</var> be undefined.</p></li>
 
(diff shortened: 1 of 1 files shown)


jgraham:
The BiDi version of this is https://github.com/w3c/webdriver-bidi/pull/155

whimboo:
CC @sadym-chromium, @bwalderman

index.html
```diff
@@ -428,8 +428,52 @@ <h3>Processing model</h3>
...
+  jump to step 1.
+
+  <li><p><p>If <var>request</var> has a <a>Host header</a>,
+  let <var>host</var> be the value of that header. Otherwise
+  let <var>host</var> be null.</p></li>
+
+  <li><p><p>If <var>request</var> has an <a>Origin header</a>,
+  let <var>origin</var> be the value of that header. Otherwise
+  let <var>origin</var> be null.</p></li>
+
```
sadym-chromium, line 442: When WebDriver server is run in a Docker container, the `Host header` is always different from the `host of the HTTP server`. Current spec includes this specific to "**_or to another host the implementation has been configured to allow._**", but maybe it would make sense to extract it into the specific step, so that the need of such a configuration will be more explicit?

```suggestion

 <li><p>Let <var>implementation allowed hosts rule</var> be the configured by the implementation rule determines if the given host allowed or not. Otherwise let <var>implementation allowed hosts rule</var> be the rule matching nothing.</p></li>
  
```

index.html
```diff
@@ -428,8 +428,52 @@ <h3>Processing model</h3>
...
+  <li><p>If any of the following conditions hold:
+   <ul>
+    <li><p><var>host</var> doesn't match the <code>Host</code>
+      grammar [[RFC7230]]</p></li>
+    <li><p>The result of <a>host parsing</a> the <code>uri-host</code>
+        part of <var>host</var> is not an <a>IP address</a>,
+        a <a>domain</a> identical to the
+        <a>host</a> of the HTTP server or to another <a>host</a> the
+        implementation has been configured to allow.</p></li>
```
sadym-chromium, line 451:
```suggestion
        <a>host</a> of the HTTP server or 
        does not match the <var>implementation allowed hosts rule</var>.</p></li>
```

index.html
```diff
@@ -428,8 +428,52 @@ <h3>Processing model</h3>
...
+  corresponds to a loopback interface [[RFC5735]].</p></li>
+
+  <li><p>If <var>origin</var> is not null, and is not identical to an
+  <a>Origin header</a> value that the implementation has been
+  configured to allow, then stop running these steps and act as if the
+  requested service is not available.</p>
+
+  <p class=note>Rejecting connections with unexpected values in
+  the <a>Origin header</a> is necessary to prevent untrusted websites
+  from establishing a WebDriver session.</p></li>
```
sadym-chromium, line 476: Or maybe add note about Docker and some other edge cases here?

whimboo:
@jgraham I assume the summary of this issue should say `WebDriver` and not `WebSocket` right? 

Beside that please also have a look at issues #1578 and #1643. Both are related and I hope that we could fix them with your PR. Thanks.

index.html
jgraham: I think it's hard to be very specific here because the details of what's required depend a lot on the setup you want to support. For example it's possible to support the docker setup by making the host of the HTTP server the same as the host of the container. But it's also possible to support it by e.g. binding to an IP address, in which case the server just has to know it should accept requests with the specific host header that the conatiner will have. So I've tried to write some general guidance in the security section which suggests you probably want to provide some configuration options to support this kind of scenario, but without saying too much about exactly what they need to be.

jgraham:
I've updated this a bit; I think it now covers all the points in the linked issues apart from:
* I didn't mention `AF_LOCAL` sockets. It definitely seems reasonable to support local-only communication as well as `AF_INET*` sockets, but I'm not sure in practice if they're usable in current clients ("local ends"). 
* I didn't mention HTTP Auth, although I did mention some kind of authorization being encouraged in cases where the "remote end" is really a publically accessible service. In practice I don't know what we could suggest which would add meaingful additional security and be compatible with common use cases like running tests in a CI system where the loal and remote ends are both on the same machine.

In any case I think if there's more to do for those points we should consider them in their own issues / PRs and limit this one to just the header validation requirements for accepting incomping requests.

bwalderman (approved):

whimboo:
@sadym-chromium in case you have some available time soon maybe you could have a look at this PR again and if your comments have been addressed or clarified? Thanks.

index.html
```diff
@@ -428,8 +428,72 @@ <h3>Processing model</h3>
...
+   <ul>
+    <li><p><var>host</var> is undefined.</p></li>
+    <li><p><var>host</var> doesn't match the <code>Host</code>
+      grammar [[RFC7230]]</p></li>
+    <li><p>The result of <a>host parsing</a> the <code>uri-host</code>
+        part of <var>host</var> is not an <a>IP address</a>,
+        a <a>domain</a> identical to the
+        <a>host</a> of the HTTP server or to another <a>host</a> the
+        implementation has been configured to allow.</p></li>
+    <li><p>The <code>port</code> part of <var>host</var> is present
```
randomstuff, line 457: I don't believe verifying the port of the Host header adds any significant security (especially with respect to DNS rebinding attacks) (?). It is probably not necessary to require this verification.

PR #1639, opened 2022-02-02
Editorial: export definitions needed for Permissions API

marcoscaceres:
The Permissions API is looking to add a Web Driver extension. However, the spec is currently missing these exports for us to do the integration.

Blocks: https://github.com/w3c/permissions/pull/346

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/pull/1639.html" title="Last updated on Feb 14, 2022, 5:58 AM UTC (22bd73d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1639/1917e89...22bd73d.html" title="Last updated on Feb 14, 2022, 5:58 AM UTC (22bd73d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -343,7 +343,7 @@ In all other cases it must be true.
 
 <p>Where algorithms that return values are fallible,
  they are written in terms of returning either
- <dfn>success</dfn> or <dfn data-lt=errors>error</dfn>.
+ <dfn class="export" data-local-lt="success">WebDriver success</dfn> (success) or <dfn data-local-lt="error" class="export">WebDriver error</dfn> (error).
  A <a>success</a> value has an associated <var>data</var> field
  which encapsulates the value returned,
  whereas an <a>error</a> response has an associated <a>error code</a>.
@@ -400,7 +400,7 @@ and therefore each command produces a single <a>HTTP response</a>.
 <p>
 In response to a <a>command</a>,
 a <a>remote end</a> will run a series of actions
-known as <dfn>remote end steps</dfn>.
+known as <dfn class="export">remote end steps</dfn>.
 These provide the sequences of actions that a <a>remote end</a> takes
 when it receives a particular <a>command</a>.
 </section>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1344,12 +1344,12 @@ <h3>Extensions</h3>
...
 
 <p>Commands defined in this way
- are called <dfn data-lt="extension command">extension commands</dfn>
+ are called <dfn class="export" data-local-lt="extension commands">WebDriver extension commands</dfn>
  and behave no differently than other <a>commands</a>;
  each has a dedicated HTTP endpoint and a set of <a>remote end steps</a>.
 
 <p>Each <a>extension command</a> has an associated
- <dfn>extension command URI Template</dfn>
+ <dfn class="export">extension command URI Template</dfn>
```
marcoscaceres, line 1352:
```suggestion
 <dfn class="export" data-local-lt="extension command URI Template">WebDriver extension command URI Template</dfn>
```

marcoscaceres:
It might also be good to get into from this group on https://github.com/w3c/permissions/pull/346

Just to double check it's been done right.

miketaylr:
Mind taking a look @AutomatedTester @jgraham?

jgraham (commented):
I'm nto s huge fan of making all these names "WebDriver [foo]". I think it should be possible to refer to them with a prefix at the point of use when required, but generally I'm not aware of a pattern of prefixing the canonical name of concepts with the name of the spec they come from.

AutomatedTester:
I agree with @jgraham here, is there a way we can do this without prefixing it all with `WebDriver`? Or are we expecting things to have the spec prefixed before all `export` items?

marcoscaceres:
Yeah, I also didn’t like it :( the Permissions spec was already doing that, so I just ran with it. Having said that, the problem is that some things we need are overly generic (e.g., “success” and “error”). But I can explicitly scope to this spec and alias them in the Permissions spec instead.

marcoscaceres:
(I’ll send an updated PR tomorrow with just the “exports”)

marcoscaceres:
Ok, hopefully all good now 🙏 ... note, I deliberately removed the `data-lt` as they are not needed (and are no good for exports, they would create duplicate entries in the definitions database).

PR #1646, opened 2022-03-15
Factor global state out of the actions algorithm

jgraham:
The main point of this change is to factor global state out of the
actions processing and dispatch algorithms, so that they can also be
used from BiDi where we don't have the same notion of a current
session, or a specific browsing context.

Along the way it turned out there were a few existing spec-internal
bugs in the way we were using the actions algorithms (e.g. calling
things with the wrong arguments), so this turned into a larger
refactor.

The biggest change is that we now don't make a distinction between an
input source and the state of the input source, which means we don't
have to take care to use the state object instead of the source
itself.

In addition, the list of active input sources is removed in favour of
just having a single input state map with one entry per active source.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2227,11 +2227,8 @@ and <a href=#elements>element retrieval</a>.
 A <a>session</a> has an associated <a>user prompt handler</a>.
 Unless stated otherwise it is in the <a>dismiss and notify state</a>.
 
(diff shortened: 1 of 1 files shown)


sadym-chromium:
looking into

AutomatedTester (approved):

PR #1653, opened 2022-04-13
Add the Element Origin concept

jgraham:
This is really designed for BiDi, allowing actions to specify the
origin relative to an element in the form:

 {
  type: "element"
  sharedId: <id>
 }

rather than sending in a WebDriver-style element reference. For
simplicit this allows the new construct to work irrespective of
whether actions are being used in a HTTP session or a BiDi
session (but a BiDi session is not expected to support the HTTP-style
web element reference identifiers).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/pull/1653.html" title="Last updated on Dec 22, 2022, 4:35 PM UTC (24e4604)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1653/6c3b67a...24e4604.html" title="Last updated on Dec 22, 2022, 4:35 PM UTC (24e4604)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -7773,7 +7773,7 @@ from a machine utilisation perspective.
 <p>An <dfn>element origin</dfn> is a Map
 which [=map/exist|contains=] "<code>type</code>" and
 "<code>sharedId</code>", and for which the value of the entry with key
-"<code>type</code>" is "<code>element</code>".
+"<code>type</code>" is "<code>node</code>".
 
 <p>To <dfn>get coordinates relative to an origin</dfn>
 given <var>source</var>, <var>x offset</var>, <var>y offset</var>,
--- a/index.html
+++ b/index.html
@@ -7844,8 +7844,6 @@ options</var>:
 <ol class="algorithm">
  <li><p>Run the substeps of the first matching value
   of <var>origin</var>
-  <p>Assert: one of these conditions matches
-
   <dl>
    <dt>"<code>viewport</code>"
    <dd>
@@ -7872,23 +7870,24 @@ options</var>:
      <var>start y</var> + <var>y  offset</var>.
     </ol>
    </dd>
+
    <dt>Otherwise</dt>
    <dd>
     <ol>
      <li><p>Let <var>element</var> be the result of <a>trying</a> to
-      run <var>actions options</var>' <a>get element origin</a>
-      steps with <var>origin</var> and <var>browsing context</var>.
+     run <var>actions options</var>' <a>get element origin</a>
+     steps with <var>origin</var> and <var>browsing context</var>.
 
-      <li><p>If <var>element</var> is null, return <a>error</a> with <a>error
-       code</a> <a>no such element</a>.
+     <li><p>If <var>element</var> is null, return <a>error</a> with <a>error
+     code</a> <a>no such element</a>.
 
      <li><p>Let <var>x element</var> and <var>y element</var> be the
-      result of calculating the <a>in-view center point</a>
-      of <var>element</var>.
+     result of calculating the <a>in-view center point</a>
+     of <var>element</var>.
 
      <li><p>Let <var>x</var> equal <var>x element</var> +
-      <var>x offset</var>, and <var>y</var> equal
-      <var>y element</var> + <var>y offset</var>.
+     <var>x offset</var>, and <var>y</var> equal
+     <var>y element</var> + <var>y offset</var>.
     </ol>
    </dd>
   </dl>
@@ -8647,7 +8646,7 @@ context</var>, and <var>actions options</var>:
 
  <li><p>Let <var>actions result</var> be the result of <a>dispatch
  actions inner</a> with <var>input state</var>, <var>actions by
- tick</var>, and <var>browsing context</var>
+ tick</var>, <var>browsing context</var>, and <var>actions options</var>.
 
  <li><p>Dequeue <var>input state</var>'s <a>actions queue</a>.
      <p>Assert: this returns <var>token</var>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -7714,6 +7714,80 @@ <h3>Processing actions</h3>
  input JSON, such that the actions to be performed in a single <a>tick</a>
  are grouped together.
 
+<p>An <dfn>element origin</dfn> is a Map
+which [=map/exist|contains=] "<code>type</code>" and
+"<code>sharedId</code>", and for which the value of the entry with key
+"<code>type</code>" is "<code>element</code>".
```
whimboo, line 7720: In the BiDi spec we have `type=node` when serializing DOM nodes. So I assume that we should have that here as well?

jgraham: Yeah, although given this should really be an Element and not something else, I wonder if we should actually give Element specific serialization that differs from Node. I'll file an issue.

index.html
```diff
@@ -7714,6 +7714,80 @@ <h3>Processing actions</h3>
...
+   <dt>An <a>element origin</a>, or an object that <a>represents a web element</a></dt>
+   <dd>
+    <ol>
+     <li>
+      <p>If <var>origin</var> is an <a>element origin</a>:
+
+      <ol>
+       <li><p>Let <var>element</var> be the item in the <a>current browsing
+          context</a>’s <a>list of known elements</a> for which the <a>web
+          element reference</a> is equal to <var>origin</var>["<code>sharedId</code>"],
```
whimboo, line 7766: Nothing in the current WebDriver classic spec actually uses `sharedId`. So if only classic is used any retrieved and serialized node will not be added to the `list of known elements` containing a `sharedId` property.

I assume that this PR is blocked by some other PR. You might want to mention that in the initial comment and probably we should add a new label to indicate that? Right now dependencies are not visible.

index.html
```diff
@@ -7714,6 +7714,80 @@ <h3>Processing actions</h3>
...
+   </dd>
+   <dt>An <a>element origin</a>, or an object that <a>represents a web element</a></dt>
+   <dd>
+    <ol>
+     <li>
+      <p>If <var>origin</var> is an <a>element origin</a>:
+
+      <ol>
+       <li><p>Let <var>element</var> be the item in the <a>current browsing
+          context</a>’s <a>list of known elements</a> for which the <a>web
```
whimboo, line 7765: I think we should update that part to make use of the window scope (depending on which will land earlier). See PR #1667 for the upcoming changes.

index.html
jgraham: The intent is that by getting the `sharedId` property from the object you're effectively extracting the web element reference identifier. Other parts of the spec shouldn't care that it came from a BiDi style `{sharedId: <id>}` object vs a class-style webelement reference object.

index.html
jgraham: This isn't perfectly sorted, but I don't think it should block this.

jgraham:
@foolip @sadym-chromium could one of you review this please?

index.html
```diff
@@ -7714,6 +7714,80 @@ <h3>Processing actions</h3>
...
+
+     <li><p>Let <var>start y</var> be equal to the <code>y</code>
+     property of <var>source</var>.
+
+     <li><p>Let <var>x</var> equal <var>start x</var> +
+     <var>x offset</var> and <var>y</var> equal
+     <var>start y</var> + <var>y  offset</var>.
+    </ol>
+   </dd>
+   <dt>An <a>element origin</a>, or an object that <a>represents a web element</a></dt>
```
foolip, line 7757: I'm a bit unclear about what exactly this check is. Following the definition of represents a web element, would this match for the object `{ "element-6066-11e4-a52e-4f735466cecf": "whatever" }`?

It would be good to spell out what the element origin type test looks like, in particular if we tolerate additional keys in the map beyond "type" and "sharedId".

sadym-chromium (commented):
Still some investigation for processing `sharedId` in Classic form our side is needed.

index.html
```diff
@@ -7714,6 +7714,80 @@ <h3>Processing actions</h3>
...
+   <dd>
+    <ol>
+     <li>
+      <p>If <var>origin</var> is an <a>element origin</a>:
+
+      <ol>
+       <li><p>Let <var>element</var> be the item in the <a>current browsing
+          context</a>’s <a>list of known elements</a> for which the <a>web
+          element reference</a> is equal to <var>origin</var>["<code>sharedId</code>"],
+          if such an element exists. Otherwise return <a>error</a> with <a>error
```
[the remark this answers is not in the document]
whimboo: I think that this relies on https://github.com/w3c/webdriver-bidi/issues/236 where we want to add a `stale` like field to the serialized data.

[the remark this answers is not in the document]
sadym-chromium: makes sense

sadym-chromium:
So the suggested way is attractive, but I guess we should not extend the Classic protocol with BiDi format. To avoid that, we can pass a `deserialization algorithm`, or more hacky boolean `elementIdField`.

foolip:
Passing in an algorithm to deserialize seems fine to me. @jgraham WDYT?

jgraham:
Sure, if we don't want to just allow the new format in classic then passing in the deserialization steps is a pretty clean approach.

sadym-chromium:
> Sure, if we don't want to just allow the new format in classic then passing in the deserialization steps is a pretty clean approach.

`is element origin` and `get element origin` solves it. LGTM

sadym-chromium (approved):

whimboo (changes requested):
@jgraham the checks are still failing given some wrong references which would need an update:

https://github.com/w3c/webdriver/actions/runs/3732469018/jobs/6331952282#step:3:893

whimboo (approved):
This looks all fine to me as well. Please check the 3 comments before merging. Thanks!

whimboo:
@jgraham mind updating this PR so that we can get it merged?

Issue #1662, opened 2022-06-09
Merge `get a known element` into `get a known connected element` because they are mostly identical


whimboo:
I just noticed that [`get a known element`](https://w3c.github.io/webdriver/#dfn-get-a-known-element) and [`get a known connected element`](https://w3c.github.io/webdriver/#dfn-get-a-known-connected-element) are mostly identical despite the fact that the latter checks for `connected` state and the former for only `is stale`. `connected` also includes a check for the active document.

Given that most commands except 4 of them use `get a known connected element` I would suggest to merge `get a known element` into it. This would also fix the case that we never actually reach step 2 from `get a known connected element` given that it calls `get known element` in step 1 and which fails with `stale element reference` in its own step 2.

Not sure why this divergence exists and maybe this was an oversight when reviewing a previous PR. But it would be good to get it fixed.

I would like to get this fixed before starting with a PR on issue #1594.

whimboo:
@shs96c you added this extra staleness logic to `get a known element` via https://github.com/w3c/webdriver/pull/687. Do you remember why that was necessary? 

Are you ok with my proposal from above so that we can actually merge both?

jgraham:
They are in fact exactly identical; `get a known connected element` first calls `get a known element`. If that call returns an error, which it does if the element either doesn't exist or is stale, we return that error. If the element exists and is not stale then `get a known connected element` then checks the element is connected. Since the definition of `is stale` is a superset of the definition of `is connected` this check is a no-op and therefore `get a known connected element` is identical to `get a known element`.

PR #1663, opened 2022-06-09
Merge "get a known element" into "get a known connected element"

whimboo:
Fix for issue #1662. @shs96c please let me know what you think. Thanks.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/pull/1663.html" title="Last updated on Jun 14, 2022, 11:23 AM UTC (272ca52)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1663/a99633b...272ca52.html" title="Last updated on Jun 14, 2022, 11:23 AM UTC (272ca52)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3558,7 +3558,7 @@ make the tab containing the <a>browsing context</a> the selected tab.
          and return its value if it is an <a>error</a>.
 
      <li><p>Let <var>element</var> be the result
-      of <a>trying</a> to <a>get a known element</a>
+      of <a>trying</a> to <a>get a known connected element</a>
       by <a>web element reference</a> <var>id</var>.
 
      <li><p>If <var>element</var> <a>is stale</a>,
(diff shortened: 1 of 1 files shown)


whimboo:
Updates for the wdspec tests will be done via https://bugzilla.mozilla.org/show_bug.cgi?id=1772484.

jgraham (approved):
I think this is a no-op change because previously `get a known connected element` was first calling `get a known element` and then pointlessly doing the connectedness checks again.

If there were cases where we wanted things to work without the connectedness check then we should propose a change so that `is stale` isn't called from `get a known element`, but we just check for the right `node document`.

whimboo:
@shs96c if you could find a minute to rubberstamp this PR it would be appreciated. As @jgraham mentioned above it's basically a no-op clean-up change. Thanks.

AutomatedTester (approved):

Issue #1594, opened 2021-05-27
Make `list of known elements` per Window, not per browsing context 


whimboo:
As discussed with @jgraham the support for `stale element reference` in WebDriver HTTP should be considered to be removed. There is basically no strong reason why `no such element` cannot handle such situations eg. after a navigation. For WebDriver BiDi it looks like not be used at all.

We are aware that the removal of this error type is backward incompatible. As such we would like to get a discussion started to know:

* if the removal as suggested is acceptable for everyone
* how often this error type might actually be used in the wild
* The strategy to get it removed

Maybe it can be discussed in the next monthly meeting.

CC'ing @AutomatedTester, @shs96c, @JohnChen0, @burg, @christian-bromann

jgraham:
Some background: We current have to produce the `Stale Element Reference` error if an element is in a different document, or if it's not connected. But that's annoying to implement. It means that you can't just have a per-window weak map between element ids and elements, because the spec requires you to figure out that the reference is valid for a DOM node that's in another document, or for a DOM node that's been removed from the DOM (and might only be kept alive by the List of Known Elements in WebDriver. 

Particuarly with heavily multiprocess architectures in modern browsers, the stale element reference distinction alone requires an implementation to keep at least a list of all known element references in the main process. Without this implementations could be significantly simplified. It seems unlikely that users are strongly depending on the distinction between `no such element` and `stale element reference`, so it seems reasonable to just always return `no such element`. If we can't change HTTP WebDriver here, I at least plan to make BiDi not have this distinction and force HTTP-based WebDriver to do the bookkeeping on which element references have ben used.

AutomatedTester:
We don't need to keep a map of all element references, just ones that have been previously searched for in the past. This is why it is described as `get known elements`. Historically, Marionette used to have this at a frame level which should work with multiprocess architectures, assuming that each frame gets a new process.

How are we we planning on handling the situation where someone has a reference to a known element and it is no longer attached to the DOM? Are we going to say the element you found a while back is now `NoSuchElement`?

jgraham:
Hmm, so I re-read the relevant spec bits, and I think maybe we'd misinterpreted the scope of the staleness checks. As long as `is stale` is always called *after* you've already failed with `no such element` for elements from a different browsing context then it's probably true that the cache can be per-process. But the spec here is a real mess; as far as I can tell it's always wrong to check for staleness except in "get a known element", and I'm not seeing an observable difference between "get a known element" and "get a known connected element".

[...]

PR #1667, opened 2022-06-14
Associate known elements with a window rather than browsing context

jgraham:
This makes the lifecycle of this element list clearer; instead of
having to keep element references alive as long as the browsing
context, you only need to keep it alive as long as the underlying
window. This also makes the behaviour more consistent; before a
navigation might create a new browsing context depending on e.g. COOP
headers, or might not. But it will always create a new Window object.

Practically the difference to users is only in the kind of error you
get in various situations. In particular if you get an element
reference E from page A then navigate A's browsing context to B, and
then try to use E, you will now get "no such element", rather than
"stale element reference". If you traverse history back to A, and it
was in the bfcache, E will be usable again. If it was not in the
bfcache, A is reloaded, and so again using E will give "no such
element". This is exactly consistent with the behaviour if you switch
to another frame or window and try to use E.

You still get a "stale element reference" error in the case that no
navigation happened but the element referenced by E was removed from
the DOM.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/pull/1667.html" title="Last updated on Jun 14, 2022, 9:46 AM UTC (8983981)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1667/a99633b...8983981.html" title="Last updated on Jun 14, 2022, 9:46 AM UTC (8983981)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4063,20 +4063,18 @@ corresponding to the <a>current top-level browsing context</a>.
 <p>An ECMAScript <a>Object</a> <dfn>represents a web element</dfn>
  if it has a <a>web element identifier</a> <a>own property</a>.
 
-<p>Each <a>browsing context</a> has an associated <dfn>list of
- known elements</dfn>.
- When the <a>browsing context</a> is <a>discarded</a>,
- the <a>list of known elements</a> is discarded along with it.
+<p>Each <a>Window</a> has an associated <dfn>list of known
+ elements</dfn>.
 
 <p>To <dfn>get a known element</dfn> with
  argument <var>reference</var>, run the following steps:
 
 <ol>
  <li>Let <var>element</var> be the item in the <a>current browsing
- context</a>’s <a>list of known elements</a> for which the <a>web
- element reference</a> is equal to <var>reference</var>, if such an
- element exists. Otherwise return <a>error</a> with <a>error
- code</a> <a>no such element</a>.
+ context</a>’s <a>active window</a>'s <a>list of known
+ elements</a> for which the <a>web element reference</a> is equal
+ to <var>reference</var>, if such an element exists. Otherwise
+ return <a>error</a> with <a>error code</a> <a>no such element</a>.
  <li>If <var>element</var> <a>is stale</a>, return
  <a>error</a> with <a>error code</a>
  <a>stale element reference</a>.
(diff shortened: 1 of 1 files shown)


whimboo:
This PR is going to fix issue #1594.

whimboo:
FYI I'm going to add `no such element` tests for all the WebDriver classic commands over on https://bugzilla.mozilla.org/show_bug.cgi?id=1774182. Right now we basically have only a few tests checking that step.

Also with https://bugzilla.mozilla.org/show_bug.cgi?id=1772484 I've already added `stale element reference` error tests for all the classic commands which should be merged soon ([see this PR on the wpt repository](https://github.com/web-platform-tests/wpt/pull/34465)).

index.html
```diff
@@ -4063,20 +4063,18 @@ <h2>Elements</h2>
 <p>An ECMAScript <a>Object</a> <dfn>represents a web element</dfn>
  if it has a <a>web element identifier</a> <a>own property</a>.
 
-<p>Each <a>browsing context</a> has an associated <dfn>list of
- known elements</dfn>.
- When the <a>browsing context</a> is <a>discarded</a>,
- the <a>list of known elements</a> is discarded along with it.
```
whimboo, line 4069: I wonder if we should keep this sentence but updated to match the appropriate window.

index.html
```diff
@@ -4373,20 +4372,18 @@ <h3 id=shadow-root>Shadow Roots</h3>
  <p>An ECMAScript <a>Object</a> <dfn>represents a shadow root</dfn>
   if it has a <a>shadow root identifier</a> <a>own property</a>.
  
- <p>Each <a>browsing context</a> has an associated <dfn>list of
+ <p>Each <a>Window</a> has an associated <dfn>list of
```
whimboo, line 4375: nit: is there a reason to have a capital letter here?

index.html
```diff
@@ -4373,20 +4372,18 @@ <h3 id=shadow-root>Shadow Roots</h3>
...
   known shadow roots</dfn>.
-  When the <a>browsing context</a> is <a>discarded</a>,
-  the <a>list of known shadow roots</a> is discarded along with it.
```
whimboo, line 4379: Same as above. Wouldn't it be good to keep this one as well?

index.html
jgraham: I think it's unnecessary; it should be obvious that something owned by the Window has the same lifecycle as the Window object.

index.html
jgraham: It's called `Window` in HTML, see https://html.spec.whatwg.org/#the-window-object

whimboo:
@AutomatedTester, @shs96c, @titusfortner, @christian-bromann - given that you all commented on issue #1594 maybe you could have a look at this pull request and let us know if you would agree with those changes? 

We would like to finally discuss this in the next monthly WebDriver meeting to unblock the work for BiDi. Thanks.

CC @sadym-chromium and @foolip to get a position from Google as well.

whimboo:
To add there is also https://github.com/w3c/webdriver-bidi/pull/180 which introduces a new notion of shared references which I missed.

jgraham:
Closing this in favour of https://github.com/w3c/webdriver/pull/1705

Issue #1649, opened 2022-03-17
Custom Locator Strategies


titusfortner:
Ideally, an Appium server should be able to register with a Selenium grid and everything should just work.

The issue we're facing is that Appium has several locator strategies that are unique to mobile applications ("accessibility id", "android datamatcher", etc)

The Selenium grid is properly raising an error since those locators are not on the approved list

> If location strategy is not present as a keyword in the [table of location strategies](https://w3c.github.io/webdriver/#dfn-table-of-location-strategies), return [error](https://w3c.github.io/webdriver/#dfn-error) with [error code](https://w3c.github.io/webdriver/#dfn-error-code) [invalid argument](https://w3c.github.io/webdriver/#dfn-invalid-argument).

Is there a reason we have to be restrictive about this, or can we specify the ones that all implementations should implement, but let the endpoint node determine if it recognizes the one provided or not?

Like, the actual correct way would probably be to create a name-spaced endpoint like 
```
POST | /session/{session id}/appium/element
```
But this would require updating every single Appium driver and all the bindings, and dealing with backwards compatibility would be a pain, and possibly prohibitive right now. A quick change to the spec and I think only a minor change to the Selenium grid code, and everything should just work if it makes sense to go this route.

titusfortner:
Actually @shs96c, doesn't this mean what you did for By.id and such isn't spec compliant?

titusfortner:
Agreed that the grid should not behave like an end node per the spec, so this is not needed.

PR #1671, opened 2022-06-21
Limiting locator validation steps to endpoint node

diemol:
Context for this PR can be seen at #1649

The approach this PR takes is to limit the execution environment for the steps to validate a locator to only the endpoint node. Selenium Grid is an intermediary node and it essentially sends back and forth WebDriver commands. Blocking non-W3C locator strategies limits folks to use different types of endpoint nodes (such as cloud providers or Appium servers).

Another approach could be to indicate that the steps are optional for intermediary nodes but mandatory for endpoint nodes.

Looking forward to your feedback, 

@shs96c @AutomatedTester @whimboo @titusfortner @jgraham

### DIFF
--- a/index.html
+++ b/index.html
@@ -4709,7 +4709,8 @@ session.execute("arguments[0].remove()", [body]);
 </code></pre>
 </aside>
 
-<p>The <a>remote end steps</a> are:
+<p>The <a>remote end steps</a> in 
+<a data-lt="endpoint node">endpoint nodes</a> are:
 
 <ol>
  <li><p>Let <var>location strategy</var> be the result
@@ -4761,7 +4762,8 @@ session.execute("arguments[0].remove()", [body]);
  </tr>
 </table>
 
-<p>The <a>remote end steps</a> are:
+<p>The <a>remote end steps</a> in 
+  <a data-lt="endpoint node">endpoint nodes</a> are:
 
 <ol>
  <li><p>Let <var>location strategy</var> be the result
@@ -4808,7 +4810,8 @@ session.execute("arguments[0].remove()", [body]);
  </tr>
 </table>
 
-<p>The <a>remote end steps</a> are:
+<p>The <a>remote end steps</a> in 
+  <a data-lt="endpoint node">endpoint nodes</a> are:
 
 <ol>
  <li><p>Let <var>location strategy</var> be the result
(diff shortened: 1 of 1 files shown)


titusfortner:
It would be more work, but I was thinking it might make more sense to do something like:

In 12.3.1 Locator strategies remove reference to "enumerated":
> An element location strategy is a keyword deciding what technique should be used to search for [elements](https://w3c.github.io/webdriver/#dfn-elements) in the [current browsing context](https://w3c.github.io/webdriver/#dfn-current-browsing-context). The following table of location strategies lists the keywords and states defined for this attribute:

Remove this part from 12.3.2, 12.3.3, 12.3.4, 12.3.5, 12.3.6, and 12.3.7
> If location strategy is not present as a keyword in the [table of location strategies](https://w3c.github.io/webdriver/#dfn-table-of-location-strategies), return [error](https://w3c.github.io/webdriver/#dfn-error) with [error code](https://w3c.github.io/webdriver/#dfn-error-code) [invalid argument](https://w3c.github.io/webdriver/#dfn-invalid-argument).

Then in the Retrieval section (12.3) we replace point 4 with:
>  If selector is not present as a keyword in the [table of location strategies](https://w3c.github.io/webdriver/#dfn-table-of-location-strategies), let elements returned be the result of [trying](https://w3c.github.io/webdriver/#dfn-try) to call an implementation-specific location strategy with arguments start node, and selector. Otherwise, let elements returned be the result of [trying](https://w3c.github.io/webdriver/#dfn-try) to call the relevant [element location strategy](https://w3c.github.io/webdriver/#dfn-strategy) with arguments start node, and selector.

shs96c:
The [spec is pretty clear](https://w3c.github.io/webdriver/#dfn-intermediary-nodes) that the intermediary nodes is acting as a proxy, and doesn't need to do any validation:

> Intermediary nodes are those that act as proxies, implementing both the [local end](https://w3c.github.io/webdriver/#dfn-local-ends) and [remote end](https://w3c.github.io/webdriver/#dfn-remote-ends) of the [protocol](https://w3c.github.io/webdriver/#protocol). However they are not expected to implement [remote end steps](https://w3c.github.io/webdriver/#dfn-remote-end-steps) directly

I don't think there's any need to change the spec to limit validation steps further in this case.

diemol:
Correct, @shs96c is right and there is no need for this PR. Closing it.

PR #1676, opened 2022-07-25
Require pointerMove and wheel coordinates to be integers

jgraham:
Prevously we allowed these properties to be undefined, and then later
tried to do computations on the values. That doesn't work; they either
need to have an integer default value or require that a value is
supplied.

Since implementations currently seem to raise an exception for these
values being undefined, the easiest approach is to require that they
are supplied, so that's what's implemented here.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/pull/1676.html" title="Last updated on Jul 25, 2022, 2:28 PM UTC (b5e4651)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1676/e3b112f...b5e4651.html" title="Last updated on Jul 25, 2022, 2:28 PM UTC (b5e4651)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -8136,9 +8136,8 @@ to <var>origin</var>.
 Let <var>x</var> be the result of <a>getting the property</a>
 <code>x</code> from <var>action item</var>.
 
-<li><p>
-If <var>x</var> is not <a>undefined</a> and is not an <a>Integer</a>,
-return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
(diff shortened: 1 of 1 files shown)


jgraham:
Tests in https://phabricator.services.mozilla.com/D152645

whimboo:
@AutomatedTester could you maybe have a look? Thanks!

shs96c (approved):
LGTM

PR #1694, opened 2022-11-15
Editorial: use navigables

marcoscaceres:
Fixes cross-refs

@jakearchibald, I was unsure what to replace "nested browsing context" in this PR... I think what was currently specified here is incorrect, but *I think* it's trying to pull the [sic] browser context from the iframe (but unsure why it was trying with all the nested ones).

 Anyone who knows this spec well, can you also advise on the above? @AutomatedTester maybe?

***
<a href="https://pr-preview.s3.amazonaws.com/marcoscaceres/webdriver/pull/1694.html" title="Last updated on Nov 15, 2022, 1:56 AM UTC (710ff12)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1694/0a24679...marcoscaceres:710ff12.html" title="Last updated on Nov 15, 2022, 1:56 AM UTC (710ff12)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -69,7 +69,7 @@ ul.brief li:last-child::after { content: "" }
 /* dl.subcategories when the definition list subdivides a definition */
 dl.subcategories { margin-left: 2em }
 </style>
-<body data-cite="CSSOM-VIEW GEOMETRY-1">
+<body data-cite="CSSOM-VIEW GEOMETRY-1 PAGE-LIFECYCLE">
 <section id=abstract>
 <p>
 WebDriver is a remote control interface
@@ -1206,7 +1206,7 @@ when it receives a particular <a>command</a>.
   <td><code>no such cookie</code>
   <td>No cookie matching the given path name
    was found amongst the <a>associated cookies</a>
-   of the <a>current browsing context</a>’s <a>active document</a>.
+   of the <a>current browsing context</a>’s [=navigable/active document=].
  </tr>
 
  <tr>
@@ -2949,7 +2949,7 @@ Return <a>success</a> with data <a><code>null</code></a>.
  <li><p><a>Handle any user prompts</a> and return its value if it is an <a>error</a>.
 
  <li><p>Let <var>current URL</var> be the <a>current top-level browsing context</a>’s
- <a>active document</a>’s [=Document/URL=].
+ [=navigable/active document=]’s [=Document/URL=].
 
  <li><p>If <var>current URL</var> and <var>url</var> do not have the same
    <a>absolute URL</a>:
@@ -3009,7 +3009,7 @@ Return <a>success</a> with data <a><code>null</code></a>.
 
  <li><p>Let <var>url</var> be the <a data-lt="URL serializer">serialization</a>
   of the <a>current top-level browsing context</a>’s
-  <a>active document</a>’s [=Document/URL=].
+  [=navigable/active document=]’s [=Document/URL=].
 
  <li><p>Return <a>success</a> with data <var>url</var>.
 </ol>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1206,7 +1206,7 @@ <h3>Errors</h3>
   <td><code>no such cookie</code>
   <td>No cookie matching the given path name
    was found amongst the <a>associated cookies</a>
-   of the <a>current browsing context</a>’s <a>active document</a>.
+   of the <a>current browsing context</a>’s [=navigable/active document=].
```
domenic, line 1209: Not sure this is a great change? A navigable's active document is very different from a browsing context's active document, so this just seems misleading?

index.html
```diff
@@ -3570,7 +3570,7 @@ <h3><dfn>Switch To Frame</dfn></h3>
       return <a>error</a> with <a>error code</a> <a>no such frame</a>.
 
      <li><p><a>Set the current browsing context</a>
-     with <var>element</var>’s <a>nested browsing context</a>.
+     to the [=node navigable=] of <var>element</var>.
```
domenic, line 3573: This isn't right. The node navigable of an element is the navigable it lives in. You want https://html.spec.whatwg.org/#nested-navigable , which should probably be exported.

jakearchibald:
There's a lot of "browsing context" in this spec. The question to ask is: does this spec actually care about browsing contexts?

For example, if this spec has a reference to a top-level browsing context, and the navigation changes the isolation such that the browsing context is swapped, would this spec want to now refer to the new browsing context, or continue to refer to the old one.

If it should be referring to the new browsing context, then it probably should be anchored in navigables rather than browsing contexts. If it really needs a browsing context at some point, it can get it from the [navigable's active browsing context](https://html.spec.whatwg.org/multipage/document-sequences.html#nav-bc).

marcoscaceres:
As an aside (and totally not the right place to have this discussion, but YOLO)... I wonder if we could work together on some kind of diagram that shows how all these things relate? 

I should also read the spec more... I know, but I'd find something like the following really useful as fast reference on my day to day work (none of the below is probably right... just for illustrative purposes): 

Screenshot 2022-11-17 at 10 14 30 am

I'd love to then link these concepts to actual/real things... you know, like "browser tab", "browser window that contains tabs", "iframe", etc.

domenic:
> There's a lot of "browsing context" in this spec. The question to ask is: does this spec actually care about browsing contexts?

I am almost certain it does not.

> I wonder if we could work together on some kind of diagram that shows how all these things relate?

I think https://html.spec.whatwg.org/#infrastructure-for-sequences-of-documents provides some of that.

Your diagram is... very incorrect :(. The hierarchy is navigable > browsing context > window > Document.

index.html
marcoscaceres: yeah... sorry, I grabbed the only one that was [exported](https://respec.org/xref/?term=active+document)

marcoscaceres:
> Your diagram is... very incorrect :(. The hierarchy is navigable > browsing context/WindowProxy > Window/Document.

I know :) I was just trying to quickly (literally) illustrate a point. I need to go through the new spec text properly. 

My question was really if I should bother trying to diagram the relationships? (or can it even be mapped out)

domenic:
I think I am not the best person to ask... I find the text descriptions pretty helpful, and the diagrams redundant. E.g. there's a concurrent discussion at https://github.com/whatwg/storage/issues/151 about adding a diagram, which to me feels just duplicative. Like, it's just replacing "X has a Y" with arrows drawn between X and Y with the words "has a"? (And similar for other relationships.)

However, it does seem other people find them helpful, and I don't want to stand in the way of clarifying the spec!

marcoscaceres:
Ok, I'm going to close this PR as it's not solving the actual problems with this spec.

I'll need to brush up a bit more on the new navigable terminology before attempting this again (or hopefully one of the actual spec editors can take this on.)

PR #1702, opened 2022-12-12
Remove `UUID` restrictions from ids

sadym-chromium:
Allow `session ID`, `web element reference`,  `shadow root reference` and `input id` to be any unique string, not only a UUID.

This is required to provide [Interop of identifiers with original WebDriver spec · Issue #59 · w3c/webdriver-bidi](http://go/gh/w3c/webdriver-bidi/issues/59).


sadym-chromium:
Implemented in https://github.com/w3c/webdriver/pull/1705/files#diff-0eb547304658805aad788d320f10bf1f292797b5e6d745a3bf617584da017051R4064

PR #1703, opened 2022-12-13
Extend `get a node` and `get or create a node reference` algorithms

sadym-chromium:
To implement interoperability between WebDriver Classic's and WebDriver BiDi's references, the `browsing context` should be passed to some algorithms.

***
<a href="https://pr-preview.s3.amazonaws.com/sadym-chromium/webdriver/pull/1703.html" title="Last updated on Dec 19, 2022, 4:32 PM UTC (ae90735)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1703/6c3b67a...sadym-chromium:ae90735.html" title="Last updated on Dec 19, 2022, 4:32 PM UTC (ae90735)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4072,14 +4072,17 @@ and a <a>node id map</a>.
 <p>A <dfn>node id map</dfn> is weak map between nodes and their
 corresponding <a>WebDriver node id</a>.
 
-<p>To <dfn>get a node</dfn> given <var>session</var>
-and <var>reference</var>:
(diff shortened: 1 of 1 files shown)


jgraham (commented):
I would make the browsing context parameter non-optional and have the callers in the WebDriver spec provide current browsing context.

jgraham:
I folded my suggestion into https://github.com/w3c/webdriver/pull/1707, so if you like that one we can land it rather than this (but if there are issues with that one, we can make the same change here instead).

PR #1704, opened 2022-12-14
Fix build for HTML changes

jgraham:
This does just enough to make the build pass for the navigable changes in HTML, without actually fixing the underlying model to match.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/pull/1704.html" title="Last updated on Dec 16, 2022, 9:01 PM UTC (96f1b0c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1704/0a24679...96f1b0c.html" title="Last updated on Dec 16, 2022, 9:01 PM UTC (96f1b0c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3200,8 +3200,11 @@ Return <a>success</a> with data <a><code>null</code></a>.
  the operating system notion of a “window”
  or the DOM <a><code>Window</code></a> object.
 
+<!-- this could use a link for destroyed, but there isn't one that works
+for both top-level and nested traversables. Generally all the infrastructure
+here needs to be rewritten. -->
 <p>A <a>browsing context</a> is said to be <dfn>no longer open</dfn>
- if it has been <a>discarded</a>.
+ if its <a>navigable</a> has been destroyed..
 
 <p>Each <a>browsing context</a> has an associated
  <dfn data-lt="window handles">window handle</dfn> which uniquely
@@ -3570,7 +3573,8 @@ make the tab containing the <a>browsing context</a> the selected tab.
       return <a>error</a> with <a>error code</a> <a>no such frame</a>.
 
      <li><p><a>Set the current browsing context</a>
-     with <var>element</var>’s <a>nested browsing context</a>.
+     with <var>element</var>’s <a>nested navigable</a>'s <a>active
+     browsing context</a>.
     </ol>
  </dl>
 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -4064,10 +4068,10 @@ <h2>Elements</h2>
...
  if it has a <a>web element identifier</a> <a>own property</a>.
 
-<p>Each <a>browsing context</a> has an associated <dfn>list of
- known elements</dfn>.
- When the <a>browsing context</a> is <a>discarded</a>,
- the <a>list of known elements</a> is discarded along with it.
+<p>Each <a>browsing context</a> has an associated <dfn>list of known
+ elements</dfn>.  When the <a>browsing
+ context</a>'s <a>navigable</a>'s <a>active window</a> is destroyed,
+ the <a>list of known elements</a> is destroyed along with it.
```
sadym-chromium, line 4074: Aha. So the list is still in the browsing context, but the lifetime is limited by the navigable. I assume this is done to avoid changing references in the spec to "browsing context's navigable". SGTM

NIT: should we say "list of known elements is **_cleared_**" instead of "**_destroyed_**"?

sadym-chromium (approved):
LGTM with 1 suggestion

index.html
jgraham: This is going to go away with the Node cache changes, so I'm not going to touch it more here.

index.html
```diff
@@ -4362,10 +4366,10 @@ <h3 id=shadow-root>Shadow Roots</h3>
  <p>An ECMAScript <a>Object</a> <dfn>represents a shadow root</dfn>
   if it has a <a>shadow root identifier</a> <a>own property</a>.
 
- <p>Each <a>browsing context</a> has an associated <dfn>list of
-  known shadow roots</dfn>.
-  When the <a>browsing context</a> is <a>discarded</a>,
-  the <a>list of known shadow roots</a> is discarded along with it.
+ <p>Each <a>browsing context</a> has an associated <dfn>list of known
+  shadow roots</dfn>.  When the <a>browsing context</a>
```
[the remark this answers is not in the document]
jgraham: Good catch, but I think this will go away with the node cache changes, so I'm going to ignore it.

PR #1705, opened 2022-12-14
Node cache

jgraham:
Generalise Element and ShadowRoot storage
    
For WebDriver BiDi we want to be able to return any Node object, and
to support all reachable Nodes, not just those in the same browsing
context. Therefore the approach of having per-Window Element and
ShadowRoot caches won't work. Instead implement the following:
    
* A per session, per browsing context group Node to id mapping that
can store any Node.
* Frontends in WebDriver to store Elements and ShadowRoots in this
map, and perform the additional same-document, connected element,
checks that the protocol requires.
    
This doesn't introduce any normative changes into WebDriver, it just
refactors the infrastructure to make it reusable in the BiDi context.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/pull/1705.html" title="Last updated on Dec 16, 2022, 9:06 PM UTC (d79ca01)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1705/7ba715f...d79ca01.html" title="Last updated on Dec 16, 2022, 9:06 PM UTC (d79ca01)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3565,8 +3565,8 @@ make the tab containing the <a>browsing context</a> the selected tab.
          and return its value if it is an <a>error</a>.
 
      <li><p>Let <var>element</var> be the result
-      of <a>trying</a> to <a>get a known connected element</a>
-      by <a>web element reference</a> <var>id</var>.
+      of <a>trying</a> to <a>get a known element</a>
+      with <var>id</var>.
 
      <li><p>If <var>element</var> is not a [^frame^]
       or [^iframe^] element,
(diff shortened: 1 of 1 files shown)


jgraham:
Note that the cache as implemented here will allow access to elements that aren't actually accessible:

> annevk@annevk:matrix.org
> jgraham: that would be all the browsing contexts belonging to a single agent cluster that belongs to a browsing context groups, > though that gives you "can potentially access" (as some of those might only be same origin-domain once you use
> document.domain)
> jgraham: https://docs.google.com/presentation/d/1hi4gH7pJPHsg_hnIj77XN_ce54HIaNUnBLenVwohFVo/edit#slide=id.g5641ecbac9_0_0 and slide 13 give an overview (related slides have approximate definitions)

So we could do a post-hoc check on the Documents sharing the same agent cluster (and origin?), but maybe in that case we should just claim that the cache is global? I'm not sure. CC @foolip

index.html
```diff
@@ -4054,51 +4058,89 @@ <h2>Elements</h2>
...
+  map</var>[<var>browsing context group</var>].
+
+  <li>If <var>node id map</var> does not contain <var>node</var>,
+  set <var>node id map</var>[<var>node</var>] to a new globally
+  unique string.
+
+ <li><p>Return <var>node id map</var>[<var>node</var>].
+</ol>
+
+<p>To <dfn>get a known element</dfn> given <var>reference</var>:
```
sadym-chromium, line 4120: What is the reason in renaming the method? It still "get a known connected element", as it checks if the element is connected via `If node is stale ...`. Just to unify with "get a known shadow root"?

index.html
```diff
@@ -4054,51 +4058,89 @@ <h2>Elements</h2>
...
- <li><p>For each <var>known element</var>
-  of the <a>current browsing context</a>’s <a>list of known elements</a>:
+  <li>Let <var>node</var> be <a>get a node</a> with <a>current session</a>
+  and <var>reference</var>.
 
-  <ol>
-   <li><p>If <var>known element</var> [=Node/equals=] <var>element</var>,
-    return <a>success</a> with <var>known element</var>’s <a>web element reference</a>.
-  </ol>
+  <li>If <a>node</a> is null, or <a>node</a> is not an <a>Element</a>,
```
sadym-chromium, line 4125: Unify the check?
```suggestion
  <li>If <a>node</a> is null, or <a>node</a> does not implement {{Element}},
```

sadym-chromium (approved):
LGTM with a question and a suggestion.

index.html
jgraham: Yeah, it just seemed preferable to go with the shorter name given that the semantics are more like "get a known element connected to the document in the current browsing context", and we don't say any of that for shadow root.

jgraham:
OK, well I'm going to merge this because it's an improvement. We can always change the details of what owns the cache later.

PR #1707, opened 2022-12-20
Fix returning stale element reference

jgraham:
If we've previously seen an element in the current window, but it was GC'd we should return "stale element reference". To support this, add a set of seen node ids to each Window, so we can check if the element is stale.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/pull/1707.html" title="Last updated on Jan 6, 2023, 11:35 AM UTC (e30204d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1707/6c3b67a...e30204d.html" title="Last updated on Jan 6, 2023, 11:35 AM UTC (e30204d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4171,7 +4171,7 @@ context</var>, and <var>reference</var> if the following steps return true:
      code</a> <a>no such element</a>.
    </ol>
 
-  <li>Return <a>success</a> with <var>node</var>.
+  <li>Return <a>success</a> with data <var>node</var>.
 </ol>
 
 <p>To <dfn>get or create a web element reference</dfn>
@@ -4462,7 +4462,7 @@ is given by:
        code</a> <a>no such shadow root</a>.
      </ol>
 
-   <li>Return <a>success</a> with <var>node</var>.
+   <li>Return <a>success</a> with data <var>node</var>.
  </ol>
 
  <p>To <dfn>get or create a shadow root reference</dfn>
--- a/index.html
+++ b/index.html
@@ -4075,7 +4075,7 @@ corresponding <a>WebDriver node id</a>.
 <p>A WebDriver <a>session</a> has a <dfn>navigable seen nodes map</dfn>
 which is a weak map between a [=navigable=] and a set.
 
-<p>To <dfn>get a node</dfn> given <var>session</var>, 
+<p>To <dfn>get a node</dfn> given <var>session</var>,
 <var>browsing context</var>, and <var>reference</var>:
 <ol class="algorithm">
   <li>Let <var>browsing context group node map</var>
@@ -4090,7 +4090,7 @@ which is a weak map between a [=navigable=] and a set.
   <li>Let <var>node id map</var> be <var>browsing context group node
   map</var>[<var>browsing context group</var>].
 
-  <li>Let <a>node</a> be the entry in <var>node id map</var> whose
+  <li>Let <var>node</var> be the entry in <var>node id map</var> whose
   value is <var>reference</var>, if such an entry exists, or null
   otherwise.
 
@@ -4156,18 +4156,20 @@ context</var>, and <var>reference</var> if the following steps return true:
   <li>Let <var>node</var> be <a>get a node</a> with <a>current session</a>,
   <a>current browsing context</a>, and <var>reference</var>.
 
-   <li>If <var>node</var> is null, or <var>node</var> does not implement
-   {{Element}}, or <var>node</var> <a>is stale</a>:
-     <ol>
-       <li><p>If <var>node</var> implements {{Element}} and <a>node
-       reference is known</a> with <a>current session</a>, <a>current
-       browsing context</a>, and <var>reference</var>,
-       return <a>error</a> with <a>error code</a> <a>stale element
-       reference</a>.
+  <li>If <var>node</var> is null, or <var>node</var> does not
+  implement {{Element}} return <a>error</a> with <a>error
+  code</a> <a>no such element</a>.
 
-       <p>Otherwise return <a>error</a> with <a>error
-       code</a> <a>no such element</a>.
-     </ol>
+  <li>If <var>node</var> <a>is stale</a>:
+   <ol>
+     <li><p>If <a>node reference is known</a> with <a>current
+     session</a>, <a>current browsing context</a>,
+     and <var>reference</var>, return <a>error</a> with <a>error
+     code</a> <a>stale element reference</a>.
+
+     <p>Otherwise return <a>error</a> with <a>error
+     code</a> <a>no such element</a>.
+   </ol>
 
   <li>Return <a>success</a> with <var>node</var>.
 </ol>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -4072,13 +4072,16 @@ <h2>Elements</h2>
 <p>A <dfn>node id map</dfn> is weak map between nodes and their
 corresponding <a>WebDriver node id</a>.
 
+<p>A [=Window=] has a [=set=] <dfn>seen node ids</dfn>, which is
+initially empty.
+
 <p>To <dfn>get a node</dfn> given <var>session</var>
-and <var>reference</var>:
+<a>browsing context</a>, and <var>reference</var>:
```
sadym-chromium, line 4079:
```suggestion
<var>browsing context</var>, and <var>reference</var>:
```

index.html
```diff
@@ -4072,13 +4072,16 @@ <h2>Elements</h2>
 <p>A <dfn>node id map</dfn> is weak map between nodes and their
 corresponding <a>WebDriver node id</a>.
 
+<p>A [=Window=] has a [=set=] <dfn>seen node ids</dfn>, which is
+initially empty.
+
 <p>To <dfn>get a node</dfn> given <var>session</var>
```
sadym-chromium, line 4078:
```suggestion
<p>To <dfn>get a node</dfn> given <var>session</var>,
```

index.html
```diff
@@ -4072,13 +4072,16 @@ <h2>Elements</h2>
...
-and <var>reference</var>:
+<a>browsing context</a>, and <var>reference</var>:
 <ol>
   <li>Let <var>browsing context group node map</var>
   be <var>session</var>'s <a>browsing context group node map</a>.
 
-  <li>Let <var>browsing context group</var> be <a>current browsing
+  <li>Let <var>browsing context group</var> be <a>browsing
   context</a>'s <a>browsing context group</a>.
```
sadym-chromium, line 4085:
```suggestion
  <li>Let <var>browsing context group</var> be <var>browsing
  context</var>'s <a>browsing context group</a>.
```

index.html
```diff
@@ -4095,13 +4098,13 @@ <h2>Elements</h2>
 </ol>
 
 <p>To <dfn>get or create a node reference</dfn>
-given <var>session</var> and <var>node</var>:
+given <var>session</var> <a>browsing context</a>, and <var>node</var>:
```
sadym-chromium, line 4101:
```suggestion
given <var>session</var>, <var>browsing context</var> and <var>node</var>:
```

index.html
```diff
@@ -4110,17 +4113,30 @@ <h2>Elements</h2>
...
 <p>To <dfn>get a known element</dfn> given <var>reference</var>:
 <ol>
-  <li>Let <var>node</var> be <a>get a node</a> with <a>current session</a>
-  and <var>reference</var>.
+  <li>Let <var>node</var> be <a>get a node</a> with <a>current session</a>,
+  <a>current browsing context</a>, and <var>reference</var>.
+
+  <li>If <a>node</a> is null and <a>current browsing
+  context</a>'s <a>active window</a>'s <a>seen node ids</a>
+  [=set/contains=] reference, return <a>error</a> with error code <a>stale
```
sadym-chromium, line 4138:
```suggestion
  [=set/contains=] <var>reference</var>, return <a>error</a> with error code <a>stale
```

index.html
```diff
@@ -4401,8 +4418,13 @@ <h3 id=shadow-root>Shadow Roots</h3>
...
  and <var>reference</var>:
  <ol>
-   <li>Let <var>node</var> be <a>get a node</a> with <a>current session</a>
-   and <var>reference</var>.
+   <li>Let <var>node</var> be <a>get a node</a> with <a>current session</a>,
+   <a>current browsing context</a>, and <var>reference</var>.
+
+   <li>If <a>node</a> is null and <a>current browsing
+   context</a>'s <a>active window</a>'s <a>seen node ids</a>
+   [=set/contains=] reference, return <a>error</a> with error code <a>stale
```
sadym-chromium, line 4426:
```suggestion
   [=set/contains=] <var>reference</var>, return <a>error</a> with error code <a>stale
```

sadym-chromium (approved):
with a few suggestions

index.html
```diff
@@ -4110,25 +4113,59 @@ <h2>Elements</h2>
...
+  map</var>[<var>navigable</var>] [=set/contains=] <var>reference</var>,
+  return true, otherwise return false.
+</ol>
+
+<p>To <dfn>get a known element</dfn> given <var>reference</var>:
+<ol class="algorithm">
+  <li>Let <var>node</var> be <a>get a node</a> with <a>current session</a>,
+  <a>current browsing context</a>, and <var>reference</var>.
+
+   <li>If <a>node</a> is null, or <a>node</a> does not implement
```
sadym-chromium, line 4159:
```suggestion
   <li>If <var>node</var> is null, or <var>node</var> does not implement
```

index.html
sadym-chromium: Maybe I misunderstand the syntax, and the `<var>some_var</var>` should be declared only once, and used as `<a>some_var</a>`. If so, please decline the suggestions.

index.html
```diff
@@ -4072,13 +4072,16 @@ <h2>Elements</h2>
 <p>A <dfn>node id map</dfn> is weak map between nodes and their
 corresponding <a>WebDriver node id</a>.
 
+<p>A WebDriver <a>session</a> has a <dfn>navigable seen nodes map</dfn>
+which is a weak map between a [=navigable=] and a set.
+
 <p>To <dfn>get a node</dfn> given <var>session</var>
```
whimboo, line 4078:
```suggestion
<p>To <dfn>get a node</dfn> given <var>session</var>, 
```

index.html
```diff
@@ -4110,25 +4113,59 @@ <h2>Elements</h2>
   <li>Let <var>node id map</var> be<var>browsing context group node
```
whimboo, line 4113:
```suggestion
  <li>Let <var>node id map</var> be <var>browsing context group node
```

index.html
```diff
@@ -4110,25 +4113,59 @@ <h2>Elements</h2>
...
+
+   <li>If <var>node</var> is null, or <var>node</var> does not implement
+   {{Element}}, or <var>node</var> <a>is stale</a>:
+     <ol>
+       <li>If <a>node reference is known</a> with <a>current
+       session</a>, <a>current browsing context</a>,
+       and <var>reference</var>, return <a>error</a> with <a>error
+       code</a> <a>stale element reference</a>. Otherwise
+       return <a>error</a> with <a>error code</a> <a>no such shadow
+       root</a>.
```
whimboo, line 4167:
```suggestion
       return <a>error</a> with <a>error code</a> <a>no such element</a>.
```

index.html
```diff
@@ -4110,25 +4113,59 @@ <h2>Elements</h2>
...
+  return true, otherwise return false.
+</ol>
+
+<p>To <dfn>get a known element</dfn> given <var>reference</var>:
+<ol class="algorithm">
+  <li>Let <var>node</var> be <a>get a node</a> with <a>current session</a>,
+  <a>current browsing context</a>, and <var>reference</var>.
+
+   <li>If <var>node</var> is null, or <var>node</var> does not implement
+   {{Element}}, or <var>node</var> <a>is stale</a>:
```
whimboo, line 4160: I don't think that we can handle all of these conditions in a single if block. Assume that a `ShadowRoot` node is returned which would trigger our code to execute the `node reference is known` method and return `true`. Here we would inappropriately raise a `stale element reference` error. The same applies to the identical checks in `get a known shadow root`.

whimboo (changes requested):

index.html
jgraham: Good catch. I've changed things so that the inner condition also checks the type. I don't think you'd implement it quite like that, but I do think it makes it easier to understand what's going on.

index.html
```diff
@@ -4072,63 +4072,104 @@ <h2>Elements</h2>
...
+     and <var>reference</var>, return <a>error</a> with <a>error
+     code</a> <a>stale element reference</a>.
 
-  <li>If <var>node</var> <a>is stale</a>, return <a>error</a>
-  with <a>error code</a> <a>stale element reference</a>.
+     <p>Otherwise return <a>error</a> with <a>error
+     code</a> <a>no such element</a>.
+   </ol>
 
   <li>Return <a>success</a> with <var>node</var>.
```
whimboo, line 4174: As  you said on another PR shouldn't this be `Return success with data`? If that's the case we also need that phrasing for `get a known shadow root`.

whimboo (approved):
Thanks for the update @jgraham! This simplified versions for getting an element and shadow root are much easier to understand!

There is just a nit you might want to have a look at.

PR #1709, opened 2023-01-03
Improve handling of objects and collections in "internal JSON clone algorithm"

whimboo:
Currently in step 2 of [clone an object](https://w3c.github.io/webdriver/#dfn-clone-an-object) only the own enumerable properties are taken into account. This actually causes issues with WebIDL interfaces like `File` which won't get the `name` property serialized.

Interestingly we have a [specific wdspec test for a `FileList` case](https://github.com/web-platform-tests/wpt/blob/231ddb37db93e1e837faeebf058e72d5ae2a5285/webdriver/tests/execute_async_script/collections.py#L28-L48), which requires the `name` property serialized. And [all browsers are currently passing this test](https://wpt.fyi/results/webdriver/tests/execute_async_script/collections.py?label=master&product=firefox%5Bexperimental%5D&product=chrome%5Bexperimental%5D&product=edge%5Bexperimental%5D&product=safari&aligned&view=subtest).

As such we probably want to remove the restriction for only serializing own properties, and allow to serialize all the enumerable properties. This won't cause a regression and keeps the behavior that is currently shown by browsers.

CC @jgraham, @sadym-chromium, @patrickangle, @gsnedders, @shs96c.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/pull/1709.html" title="Last updated on Jan 6, 2023, 1:41 PM UTC (0ce48de)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1709/6c3b67a...0ce48de.html" title="Last updated on Jan 6, 2023, 1:41 PM UTC (0ce48de)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -6470,40 +6470,36 @@ a <a>remote end</a> must run the following steps:
 <dl class=switch>
  <dt><a>undefined</a>
  <dt><a><code>null</code></a>
- <dd><p><a>Success</a> with data <a><code>null</code></a>.
+ <dd><p>Return <a>success</a> with data <a><code>null</code></a>.
 
  <dt>type <a>Boolean</a>
  <dt>type <a>Number</a>
  <dt>type <a>String</a>
- <dd><p><a>Success</a> with data <var>value</var>.
-
- <dt>a <a>collection</a>
- <dd>
-  <ol>
-   <li><p>Let <var>result list</var> be an empty list.
-
-   <li><p>For each <var>item</var> of <var>value</var>,
-    add the result of <a>trying</a> to <a>clone an object</a>
-    with arguments <var>item</var> and <var>seen</var>
-    and the <a>internal JSON clone algorithm</a> as the <var>clone algorithm</var>
-    to <var>result list</var>.
-
-   <li><p>Return <a>success</a> with <var>result list</var>.
-  </ol>
+ <dd><p>Return <a>success</a> with data <var>value</var>.
 
  <dt>instance of {{Element}}</a>
  <dd><p>If the <var>element</var> <a>is stale</a>,
   return <a>error</a> with <a>error code</a> <a>stale element reference</a>.
 
-  <p>Otherwise return <a>success</a> with data set to the <a>web
-  element reference object</a> for <var>value</var>.
+  <p>Otherwise:
+  <ol>
+   <li><p>Let <var>reference</var> be the <a>web element reference object</a>
+    for <var>value</var>.
+
+   <li><p>Return <a>success</a> with data <var>reference</var>.
+  </ol>
 
  <dt>instance of {{ShadowRoot}}</a>
  <dd><p>If the <var>shadow root</var> <a>is detached</a>,
   return <a>error</a> with <a>error code</a> <a>detached shadow root</a>.
 
-  <p>Otherwise return <a>success</a> with data set to the <a>shadow
-  root reference object</a> for <var>value</var>.
+  <p>Otherwise:
+  <ol>
+   <li><p>Let <var>reference</var> be the <a>shadow root reference object</a>
+    for <var>value</var>.
+
+    <li><p>Return <a>success</a> with data <var>reference</var>.
+  </ol>
 
  <dt>a {{WindowProxy}} object
  <dd><p>If the associated <a>browsing context</a>
@@ -6511,9 +6507,13 @@ a <a>remote end</a> must run the following steps:
   in <var>value</var> has been destroyed,
   return <a>error</a> with <a>error code</a> <a>stale element reference</a>.
 
-  <p>Otherwise return <a>success</a> with data set
-   to <a><code>WindowProxy</code> reference object</a>
-   for <var>value</var>.
+  <p>Otherwise:
+  <ol>
+   <li><p>Let <var>reference</var> be the
+    <a><code>WindowProxy</code> reference object</a> for <var>value</var>.
+
+    <li><p>Return <a>success</a> with data <var>reference</var>.
+  </ol>
 
  <dt>has an <a>own property</a> named "<code>toJSON</code>" that is
   a <a>Function</a>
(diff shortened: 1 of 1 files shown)


whimboo:
Here an example snippet to replicate the behavior - no properties are returned in each of the browsers:

```
var file = new File(["foo"], "bar");
Object.getOwnPropertyNames(file);
```

whimboo:
This PR actually needs some more updates given that there are other failing cases:

* we try to serialize properties that are methods
* the serialization of a collection needs to happen after the `seen` check for `value`

index.html
```diff
@@ -6470,26 +6470,12 @@ <h3>Executing Script</h3>
 <dl class=switch>
  <dt><a>undefined</a>
  <dt><a><code>null</code></a>
- <dd><p><a>Success</a> with data <a><code>null</code></a>.
+ <dd><p><p>Return <a>success</a> with data set to <a><code>null</code></a>.
```
jgraham, line 6473: We just say "success with data [foo]", not "data set to [foo]" (here and throughout).

index.html
```diff
@@ -6517,26 +6503,20 @@ <h3>Executing Script</h3>
 
  <dt>has an <a>own property</a> named "<code>toJSON</code>" that is
   a <a>Function</a>
- <dd>Return <a>success</a> with the value returned by
+ <dd><p><p>Let <var>result</var> be the return value of
```
jgraham, line 6506: This isn't quite right, because `[[Call]]` will return a [completion record](https://tc39.es/ecma262/#sec-completion-record-specification-type) [evaluate](https://w3c.github.io/webdriver-bidi/#command-script-evaluate) in BiDi has a better example of how to run a script (and we probably do indeed need to do all the bookkeeping there). Altneratively you might be able to use https://w3c.github.io/webdriver/#dfn-execute-a-function-body although, I wonder if the alert handling there is what we want (a testcase with `{toJSON: () => alert("surprise!")}` would be informative).

index.html
```diff
@@ -6581,6 +6568,12 @@ <h3>Executing Script</h3>
 
    <li><p>Otherwise, return <var>cloned property result</var>.
   </ol>
+
+  <li>
+    <p>Remove the last element of <var>seen</var>.
```
jgraham, line 6573: Generally we keep the `<p>` on the same line as the `<li>`

jgraham (changes requested):

index.html
whimboo: I didn't modify this content for that. Would it be ok for you when we move this particular request to a follow-up issue?

index.html
whimboo: Sure. Will push an update.

index.html
whimboo: I'll fix that.

index.html
```diff
@@ -6470,39 +6470,25 @@ <h3>Executing Script</h3>
 <dl class=switch>
  <dt><a>undefined</a>
  <dt><a><code>null</code></a>
- <dd><p><a>Success</a> with data <a><code>null</code></a>.
+ <dd><p><p>Return <a>success</a> with data <a><code>null</code></a>.
```
jgraham, line 6473: If we're fixing inconsistencies anyway, we usually just write "`null`" rather than using `<code>` or linking it.

index.html
```diff
@@ -6470,39 +6470,25 @@ <h3>Executing Script</h3>
...
-   <li><p>Return <a>success</a> with <var>result list</var>.
-  </ol>
+ <dd><p>Return <a>success</a> with data <var>value</var>.
 
  <dt>instance of {{Element}}</a>
  <dd><p>If the <var>element</var> <a>is stale</a>,
   return <a>error</a> with <a>error code</a> <a>stale element reference</a>.
 
-  <p>Otherwise return <a>success</a> with data set to the <a>web
+  <p>Otherwise return <a>success</a> with data <a>web
```
jgraham, line 6484: So I think the reasons these ones used "set to the" is that the construct "with data [getter steps]" is more awkward than "with data [variable]". So I'd fix them more like

```
<p>Otherwise:
  <ol>
    <li><p>Let <var>reference</var> be the <a>web element reference object</var> for <var>value</var>.
    <li><p>Return <a>success</a> with data <var>reference</var>
  </ol>
```

But this is of course way more verbose.

index.html
```diff
@@ -6511,32 +6497,25 @@ <h3>Executing Script</h3>
...
+ <dt>Otherwise
  <dd>
   <ol>
-   <li><p>If <var>value</var> is in <var>seen</var>,
-    return <a>error</a> with <a>error code</a> <a>javascript error</a>.
-
-   <li><p>Append <var>value</var> to <var>seen</var>.
-
-   <li><p>Let <var>result</var> be the value of running the
+   <li><p>Return <a>success</a> with data return value of running the
```
jgraham, line 6515: Let's set the return value of the algorithm to a variable and return data with that.

index.html
```diff
@@ -6581,6 +6565,10 @@ <h3>Executing Script</h3>
 
    <li><p>Otherwise, return <var>cloned property result</var>.
   </ol>
+
+  <li><p>Remove the last element of <var>seen</var>.
+
+  <li>Return <a>success</a> with data <var>result</var> value.
```
jgraham, line 6571: s/ value//

index.html
jgraham: This seems like one of the two fundamental spec bugs here. If you want to leave it to a followup, I'd avoid changing this entirely rather than changing it from one wrong thing to a different but equally wrong thing.

index.html
whimboo: If we make such a change we should do that for the full document at once. This drastically enlarges the scope of this PR and we should be better of doing it in a follow up PR if really needed.

index.html
whimboo: I'm fine with that change, and will apply it.

index.html
jgraham: Ah yeah, this is apparently different compared to BiDi and other specs. We don't need to fix it here.

jgraham (approved):

PR #1710, opened 2023-01-03
Replace misleading term

jugglinmike:
This specification relies on the term "response" as defined by the Fetch standard. The "error" type which is returned by fallible algorithms does not qualify as such a response. This makes the use of the phrase "error response" to define the "error" type somewhat confusing.

Replace "error response" with the more generic "error value" to avoid insinuating a non-existent relationship.

***
<a href="https://pr-preview.s3.amazonaws.com/bocoup/webdriver/pull/1710.html" title="Last updated on Jan 3, 2023, 10:49 PM UTC (7c743be)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1710/6c3b67a...bocoup:7c743be.html" title="Last updated on Jan 3, 2023, 10:49 PM UTC (7c743be)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -349,7 +349,7 @@ In all other cases it must be true.
  <dfn class="export">success</dfn> or <dfn class="export">error</dfn>.
  A <a>success</a> value has an associated <var>data</var> field
  which encapsulates the value returned,
- whereas an <a>error</a> response has an associated <a>error code</a>.
+ whereas an <a>error</a> value has an associated <a>error code</a>.
 
 <p>When calling a fallible algorithm,
  the construct “Let <var>result</var> be the result


jugglinmike:
The type of "success" and "error" values is also under-specified. I think it can only be inferred to be "a thing with a field." While formally classifying it (perhaps as an Infra "map" value) might be more complete in some theoretical sense, I'm not sure the current ambiguity could cause confusion.

jgraham (approved):
I think this is just a typo in that it's not a response at all.

I also don't think trying to "explain" this in terms of infra values makes much sense. In particular modelling them as maps is wrong because they have fixed data, and modelling them as structs doesn't capture the essential fact that they form a sum type (i.e. a spec function using these types can _always_ return `success` _or_ `error` and you are required to handle both).

PR #1712, opened 2023-01-10
Separate out existence check of element and shadow root nodes from staleness check

### LINKED ISSUE #354: Need a way to enable / disable networking from webdriver
James Graham:
> The new generation of web apps are expected to continue to function when networking is not available for whatever reason. Obviously this is particularly important on mobile devices to close the gap between native apps and web apps.
>
> At present it isn't possible to test the behaviour of an app when it is offline, or the transition between online and offline or vice-versa. This substantially decreases the utility of WebDriver for testing contemporary web applications. Neither is it possible to write testsuites for the features underlying offline support (AppCache, Service Worker), substantially increasing the chance of buggy or non-interoperable implementations. 
>
> The most obvious way to provide this would be to expose an API to webdriver that would allow disabling "content" networking i.e. from the point of view of the webpage it would look like the browser was offline, but privileged code (in particular webdriver itself) would still be able to perform network operations.

whimboo:
Fixes #354.


whimboo:
As it looks like the proposed changes are not enough. When [get a node](https://w3c.github.io/webdriver/#dfn-get-a-node) returns `null` because an element has been garbage collected, its web element reference is still known and as such step 2 should not throw a `no such element` error.

whimboo:
There was one more regression that the most recent update fixes now. Without it we only checked for a `node reference is known` in case of the element is `null`. But that's broken and the check has to always be performed. As such we can do that at the very beginning even before `get node` given that both work independently from each other.

whimboo:
Related changes should not affect the behavior of current WebDriver classic tests. This can be [proven by the wdspec results](https://treeherder.mozilla.org/jobs?repo=try&revision=d12810578cc564bc4eb8b76a07bd1d8954d8667a) for my appropriate changes for Firefox in https://bugzilla.mozilla.org/show_bug.cgi?id=1807227.

index.html
```diff
@@ -4152,24 +4152,21 @@ <h2>Elements</h2>
...
-     session</a>, <a>current browsing context</a>,
-     and <var>reference</var>, return <a>error</a> with <a>error
-     code</a> <a>stale element reference</a>.
+  <li>If <var>node</var> is not null and <var>node</var> does not implement
+  {{Element}} return <a>error</a> with <a>error code</a> <a>no such element</a>.
 
-     <p>Otherwise return <a>error</a> with <a>error
-     code</a> <a>no such element</a>.
-   </ol>
+  <li>If <var>node</var> is null or <var>node</var> <a>is stale</a> return
```
jgraham, line 4168: Since this is apparently confusing, add a comment:
```<!-- since we already checked for known references, null here means that the object was GC'd -->```

whimboo: Sure. Let me know if the updated content that I just pushed is what you expect.

PR #1714, opened 2023-01-18
Fix detached check for "get a known shadow root" step

whimboo:
Regression from #1712.

Edit: The issue here is that by copying and pasting in the other PR I accidentally used the term `is stale` instead of `is detached` for detached checks when deserializing ShadowRoot references.

@jgraham can you please review?

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/pull/1714.html" title="Last updated on Jan 19, 2023, 12:26 PM UTC (39c13da)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1714/80b6d5d...39c13da.html" title="Last updated on Jan 19, 2023, 12:26 PM UTC (39c13da)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4459,7 +4459,7 @@ is given by:
   <!-- Since we already checked for known references, null here means that
   the object was GC'd -->
 
-  <li>If <var>node</var> is null or <var>node</var> <a>is stale</a> return
+  <li>If <var>node</var> is null or <var>node</var> <a>is detached</a> return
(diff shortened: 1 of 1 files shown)


jgraham (approved):
OK, so the point here is that just the is stale check doesn't work on the shadow root because it's not an element, and so the connectedness check doesn't work. The is detached check both checks that the shadow root's node document is the active document, and also that its host is connected.

I think that makes sense.

PR #1728, opened 2023-03-20
Add "DOMTokenList" to collections

whimboo:
Right now the `DOMTokenList` is not handled as collection and as such falls back to an arbitrary object. 

The serialization results in an empty object to be returned because there are no enumerable properties. To actually allow `Get Element Property` to return the `classList` it needs to be handled as a collection.

***
<a href="https://pr-preview.s3.amazonaws.com/whimboo/webdriver/pull/1728.html" title="Last updated on Mar 20, 2023, 5:31 PM UTC (3bbccc1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1728/a638d46...whimboo:3bbccc1.html" title="Last updated on Mar 20, 2023, 5:31 PM UTC (3bbccc1)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -6477,6 +6477,7 @@ and whose:
 <ul>
 <li><a>initial value</a> of the <code>toString</code> <a>own property</a> is "<code>Arguments</code>"
 <li>instance of {{Array}}
+<li>instance of {{DOMTokenList}}
 <li>instance of {{FileList}}
 <li>instance of {{HTMLAllCollection}}
 <li>instance of {{HTMLCollection}}


whimboo:
Tests will be added via https://bugzilla.mozilla.org/show_bug.cgi?id=1823464.

whimboo:
@gsnedders could you have a look as well? Thanks.

whimboo:
> This seems fine. But: we should probably come up with a clearer policy as to what we actually want to include here—there's plenty more interfaces we could reasonable include here, after all.

I agree but someone would have to collect these. Maybe we can put a focus for that for Webdriver BiDi first where https://github.com/w3c/webdriver-bidi/issues/318 is already filed. Then we could backport to classic if needed.

whimboo:
@jgraham and @sadym-chromium can you please review? Thanks.

sadym-chromium:
Do we really want to change WebDriver Classic serialization, as it is in support mode?

sadym-chromium:
> Do we really want to change WebDriver Classic serialization, as it is in support mode?

Even though the change seems reasonable.

jgraham:
AIUI the change aligns the spec with Chrome's existing behaviour?

gsnedders:
> Do we really want to change WebDriver Classic serialization, as it is in support mode?

I think categorising WebDriver Classic as "support mode" is disingenuous; if it were truly in some sort of support mode, we wouldn't still be adding new endpoints to it.

Regardless, as we do with the web platform, we should continue to ensure legacy functionality keeps working, even in the face of new features.

sadym-chromium (approved):

Issue #1730, opened 2023-03-24
Call of toJSON in the internal JSON clone algorithm


nechaev-chromium:
According to the [internal JSON clone algorithm](https://www.w3.org/TR/webdriver/#dfn-internal-json-clone-algorithm):

>  (if the serialized object) has an [own property](https://www.w3.org/TR/webdriver/#dfn-own-properties) named "toJSON" that is a [Function](https://www.w3.org/TR/webdriver/#dfn-function) return [success](https://www.w3.org/TR/webdriver/#dfn-success) with the value returned by [Function.[[Call]]](https://www.w3.org/TR/webdriver/#dfn-call)(toJSON) with value as the this value.

The standard does not mention what to do with [inherited property](https://262.ecma-international.org/5.1/#sec-4.3.31) toJSON of the serialized object. Would it make sense to handle toJSON uniformly for both own and inherited toJSON property?

Use case:
According to the ECMAScript standard [toJSON](https://262.ecma-international.org/5.1/#sec-15.9.5.44)  is an inherited property of Date object defined in Date.prototype. As it is not an own property we have to serialize Date as an object instead of calling Date.prototype.toJSON

Implementations:
ChromeDriver
* The serialization of Date object [used to call](https://source.chromium.org/chromium/chromium/src/+/main:chrome/test/chromedriver/js/call_function.js;drc=4a8573cb240df29b0e4d9820303538fb28e31d84;l=358) Date.prototype.toJSON until ver 111
* Starting from ver 111 ChromeDriver serializes Date as Object yielding {}. The result seems to be unexpected for the users (chromedriver:4395)[[ ](https://bugs.chromium.org/p/chromedriver/issues/detail?id=4395)].

GeckoDriver
It looks like it also calls Date.prototype.toJSON as the serialization result is a string of the form YYYY-MM-DDTHH:mm:ss.sssZ (ISO8601)

PR #1732, opened 2023-04-05
Serialization also calls inherited toJSON

nechaev-chromium:
Serialization mechanism calls not only own toJSON method but also inherited toJSON method of the serialized object

Fixing [Call of toJSON in the internal JSON clone algorithm](https://github.com/w3c/webdriver/issues/1730).

***
<a href="https://pr-preview.s3.amazonaws.com/nechaev-chromium/webdriver/pull/1732.html" title="Last updated on Apr 5, 2023, 3:28 PM UTC (1992c6e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1732/a65f9fb...nechaev-chromium:1992c6e.html" title="Last updated on Apr 5, 2023, 3:28 PM UTC (1992c6e)">Diff</a>

- - -
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1732)

### DIFF
--- a/index.html
+++ b/index.html
@@ -6581,8 +6581,8 @@ a <a>remote end</a> must run the following steps:
     <li><p>Return <a>success</a> with data <var>reference</var>.
   </ol>
 
- <dt>has an <a>own property</a> named "<code>toJSON</code>" that is
-  a <a>Function</a>
+ <dt>has an <a>own property</a> or <a>inherited property</a> named
+    "<code>toJSON</code>" that is a <a>Function</a>
  <dd>Return <a>success</a> with the value returned by
   <a>Function.[[\Call]]</a>(<code>toJSON</code>) with <var>value</var>
   as the this value.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -6581,8 +6581,8 @@ <h3>Executing Script</h3>
     <li><p>Return <a>success</a> with data <var>reference</var>.
   </ol>
 
- <dt>has an <a>own property</a> named "<code>toJSON</code>" that is
-  a <a>Function</a>
+ <dt>has an <a>own property</a> or <a>inherited property</a> named
```
whimboo, line 6584:
```suggestion
 <dt>has a property</a> named
```

Given that it doesn't matter if it's an own or inherited property we potentially could just use this term?

whimboo (commented):
Generally it looks fine to me but I wonder what might be missing in case an object inherits `toJSON()` and defines some own properties. In case handling only an own `toJSON()` call, we could assume that the object knows what to serialize, but if we call it on one of the super classes it might miss values?

@jgraham what do you think?

MurkyWaters2 (approved):

PR #1740, opened 2023-05-22
Update key codes

jrandolf:
The key codes for `\uE00B`, `\uE03D`, `\uE053` are either incorrect or missing. The key code for `\uE019` should be `NumpadEqual` to distinguish it from `=`.

See https://github.com/web-platform-tests/wpt/pull/40122 for tests

#### Manual tests

On the two major platforms (macOS and Windows), we have:

 - `=` has the code value `NumpadEqual` for the numpad value.
 - `Pause` has the code value `Pause`

This is on a US-layout extended keyboard.

#### Reference

 - https://www.w3.org/TR/uievents-code/
 - https://www.w3.org/TR/uievents-key/

***
<a href="https://pr-preview.s3.amazonaws.com/jrandolf/webdriver/pull/1740.html" title="Last updated on Jul 24, 2023, 1:11 PM UTC (9e37f98)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1740/a65f9fb...jrandolf:9e37f98.html" title="Last updated on Jul 24, 2023, 1:11 PM UTC (9e37f98)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -9029,8 +9029,9 @@ and <var>actions options</var>:
  <tr><td><code>"\uE009"</code></td><td><code></code></td><td><code>"ControlLeft"</code></td></tr>
  <tr><td><code>"\uE051"</code></td><td><code></code></td><td><code>"ControlRight"</code></td></tr>
  <tr><td><code>"\uE006"</code></td><td><code></code></td><td><code>"Enter"</code></td></tr>
- <tr><td><code>"\uE03D"</code></td><td><code></code></td><td><code>"OSLeft"</code></td></tr>
- <tr><td><code>"\uE053"</code></td><td><code></code></td><td><code>"OSRight"</code></td></tr>
+ <tr><td><code>"\uE00B"</code></td><td><code></code></td><td><code>"Pause"</code></td></tr>
+ <tr><td><code>"\uE03D"</code></td><td><code></code></td><td><code>"MetaLeft"</code></td></tr>
+ <tr><td><code>"\uE053"</code></td><td><code></code></td><td><code>"MetaRight"</code></td></tr>
  <tr><td><code>"\uE008"</code></td><td><code></code></td><td><code>"ShiftLeft"</code></td></tr>
  <tr><td><code>"\uE050"</code></td><td><code></code></td><td><code>"ShiftRight"</code></td></tr>
  <tr><td><code>" "</code></td><td><code>"\uE00D"</code></td><td><code>"Space"</code></td></tr>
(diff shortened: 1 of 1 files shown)


jrandolf:
PTAL @whimboo @sadym-chromium @jgraham

sadym-chromium (approved):

jrandolf:
We'd still like at least one other vendor to agree to these changes. Even though this is approved, PTAL @whimboo @jgraham

jrandolf:
@jgraham The test is already updated here: https://github.com/web-platform-tests/wpt/pull/40122

jrandolf:
@jgraham Could you approve this PR?

gsnedders:
> I think this kind of change requires tests. Are you able to make a test PR in web-platform-tests and link it here?

We probably want to test browsers on multiple platforms, not just the configurations that WPT tests.

whimboo:
So I had a look at [known keyCodes on MDN ](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode)but the `MetaLeft` and `MetaRight` keycodes aren't listed.  But then I found https://github.com/w3c/uievents-code/issues/7 which exactly covers the renaming of the keyCode's for Meta keys. Now I assume MDN is outdated and would need to be updated?

jrandolf:
@whimboo Yes, MDN needs to be updated.

whimboo:
> So I had a look at [known keyCodes on MDN ](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode)but the `MetaLeft` and `MetaRight` keycodes aren't listed. But then I found [w3c/uievents-code#7](https://github.com/w3c/uievents-code/issues/7) which exactly covers the renaming of the keyCode's for Meta keys. Now I assume MDN is outdated and would need to be updated?

@bsmth, could you maybe help us to figure out what needs to be updated on MDN regarding the updated key codes? Thanks!

bsmth:
Hi @whimboo - I've linked a tracking issue on MDN so we have some context for the changes. It's probably faster if I open a PR directly for the renaming if you would like to review it.

whimboo:
> Hi @whimboo - I've linked a tracking issue on MDN so we have some context for the changes. It's probably faster if I open a PR directly for the renaming if you would like to review it.

Not sure if I'm the best person to review. Instead I would like to actually leave it up to some DOM peer to review once the PR has been created.

whimboo:
> > So I had a look at [known keyCodes on MDN ](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode)but the `MetaLeft` and `MetaRight` keycodes aren't listed. But then I found [w3c/uievents-code#7](https://github.com/w3c/uievents-code/issues/7) which exactly covers the renaming of the keyCode's for Meta keys. Now I assume MDN is outdated and would need to be updated?

The page on MDN has now been updated for `MetaLeft` and `MetaRight`. That means we have to decide about the two remaining `Pause` and `NumpadEqual` keycodes.

Also possible backward incompatibilities should be checked for Selenium clients given that the keyCodes are changed. @jrandolf given that Chrome seems to use those keys already are there any tests in the Selenium repository failing and most likely marked as `xfail` because of the correct but different behavior?

whimboo:
Here is an example page that is printing all the key information:
https://hsivonen.com/test/moz/input.html

Using Firefox it is still printing `OSLeft` and `OSRight` for the meta keys on MacOS. Chrome and Safari already use `MetaLeft` and `MetaRight`.

@smaug---- or @masayuki-nakano do you know why we haven't changed it in Firefox yet? Maybe we missed that, which explains the [recent change to the MDN page](https://github.com/mdn/content/issues/27459).

masayuki-nakano:
Just nobody hasn't worked on it. (They were renamed after I confirmed the stability of key names at initial implementation...)
https://bugzilla.mozilla.org/show_bug.cgi?id=1232918

whimboo:
> Just nobody hasn't worked on it. (They were renamed after I confirmed the stability of key names at initial implementation...) https://bugzilla.mozilla.org/show_bug.cgi?id=1232918

@masayuki-nakano the bug only mentions Windows and Linux. But given that MacOS is affected as well we probably need to adjust the bug?

jgraham (approved):
I think this makes sense. The NumpadEquals thing in particular looks like a mistake in the original spec, given that one can already send `=` as the keycode for normal non-numpad equals.

PR #1757, opened 2023-08-24
Point Apple Safari Vendor status to W3C WebDriver protocol supported commands

aprotyas:
In this commit we update the Safari vendor status document to point to the set of commands supporting the W3C WebDriver protocol. The previous link redirects to [macOS WebDriver Commands for Safari 11.1 and earlier](https://developer.apple.com/documentation/webkit/macos_webdriver_commands_for_safari_11_1_and_earlier?language=objc), which highlights the WebDriver Commands supporting the Selenium JSON Wire Protocol instead.


whimboo (approved):
Thanks!

PR #1790, opened 2024-02-08
Add userAgent capability

OrKoN:


whimboo:
This looks fine to me. @jgraham and @gsnedders mind checking as well?

gsnedders:
Does this not also need processing in "validate capabilities" (and maybe more in "matching capabilities"? it's not clear to me whether that's deliberately only a subset of capabilities).

As far as I can tell, with this PR passing:

```json
{
  "capabilities": {
    "alwaysMatch": {
      "userAgent": "This/1.0 Is/2.0 Not/3.0 (A User-Agent String"
    }
  }
}
```

…does nothing, and it will still manage to satisfy the capabilities?

OrKoN:
@gsnedders I intended to add userAgent as an output capability only as described [by this definition](https://w3c.github.io/webdriver/#dfn-matched-capability-serialization-algorithm)("provide any computed value to return to the user") so it should not be interpreted as part of the capabilities request (and my understanding is that in this case updates to the validation algorithms are not needed).

whimboo (approved):

OrKoN:
So I believe the following in the request

```
{
  "capabilities": {
    "alwaysMatch": {
      "userAgent": "This/1.0 Is/2.0 Not/3.0 (A User-Agent String"
    }
  }
}
```

would result in a failure to create a session according to https://w3c.github.io/webdriver/#dfn-matching-capabilities (Step 3, the Otherwise clause).

We had a discussion with @sadym-chromium and @jrandolf and we are wondering if we should instead allow clients to opt-in into getting the capability result?  E.g., by sending

```
{
  "capabilities": {
    "alwaysMatch": {
      "userAgent": true
    }
  }
}
```

the client will receive the userAgent string in the response.

OrKoN:
Updated the PR to handle userAgent when matching capabilities.

nechaev-chromium (approved):

index.html
```diff
@@ -1569,6 +1569,13 @@ <h2>Capabilities</h2>
   <td>Describes the <a>current session</a>’s <a>user prompt handler</a>.
    Defaults to the <a>dismiss and notify state</a>.
  </tr>
+
+ <tr>
+  <td><dfn>User Agent</dfn>
+  <td>"<code>userAgent</code>"
+  <td>string
+  <td>Provides the <a>remote end</a>'s <a>default User-Agent value</a>.
```
whimboo, line 1577: To be in sync with the other application / os specific capabilities:

```suggestion
  <td>Identifies the <a>remote end</a>'s <a>default User-Agent value</a>.
```

index.html
```diff
@@ -2106,6 +2116,11 @@ <h3>Processing capabilities</h3>
        used. The intent is that if this is not possible a new session
        will not be established.
 
+      <dt>"<code>userAgent</code>"
+      <dd><p>return <a>success</a> with data <a><code>null</code></a>.
+
+        <p class=note>Matching is not supported for this capability.
+
```
whimboo, line 2123: Shall we leave this out for now until it has been agreed on how to handle https://github.com/w3c/webdriver/issues/1792?

gsnedders: +1 for leaving this out until we sort out the (pre-existing, broader) issue.

OrKoN: Removed this part for now.

jgraham:
FWIW I'm happy with this only being returned, not used as the basis for a match.

index.html
```diff
@@ -1992,6 +1999,9 @@ <h3>Processing capabilities</h3>
 
    <dt>"<code>setWindowRect</code>"
    <dd>Boolean indicating whether the <a>remote end</a> supports all of the <a href=#resizing-and-positioning-windows>resizing and positioning</a> <a>commands</a>.
+
+   <dt>"<code>userAgent</code>"
+   <dd>String containing the <a>default User-Agent value</a>.
```
whimboo, line 2004: To be in sync with the other string capabilities:

```suggestion
   <dd>The <a>default user agent value</a> as a <a>string</a>.
```

Also for both the the `browserName` and `browserVersion` we use `user agent`. Should we use that here as well and update the dfn below?

index.html
```diff
@@ -1569,6 +1569,13 @@ <h2>Capabilities</h2>
   <td>Describes the <a>current session</a>’s <a>user prompt handler</a>.
    Defaults to the <a>dismiss and notify state</a>.
  </tr>
+
+ <tr>
+  <td><dfn>User Agent</dfn>
+  <td>"<code>userAgent</code>"
+  <td>string
+  <td>Identifies the <a>remote end</a>'s <a>default User-Agent value</a> of the <a>endpoint node</a>.
```
whimboo, line 1577: An `endpoint node` is actually a subset of `remote end`. So I assume we can get rid of the latter similar to the other prefs?

```suggestion
  <td>Identifies the <a>default User-Agent value</a> of the <a>endpoint node</a>.
```

whimboo (changes requested):

index.html
OrKoN: "default User-Agent value" is a link referencing another spec. I think the user agent in other capabilities is the infra spec user agent.

OrKoN: Would linking still work if the link name is not an exact match?

jgraham (approved):

index.html
whimboo: Looks like we can resolve this issue.

whimboo (approved):

whimboo:
@OrKoN I assume that you want to add a new test for classic that checks for the user agent to be part of the session capabilities? We should probably have a separate test for now until all browsers have support for it.

whimboo:
@OrKoN are you going to add a wpt test for it?

OrKoN:
Yes, I am on it

Issue #1793, opened 2024-02-14
setWindowRect should be matched as a capability


gsnedders:
c.f. https://matrix.to/#/!zvtoFTlOBeOwHrBlEK:matrix.org/$yumW_YPZ6kAsKlD3SB_muMP9ImAUoF893INwdNegTEw?via=matrix.org&via=mozilla.org&via=igalia.com

Nowhere does https://w3c.github.io/webdriver/#dfn-matching-capabilities match setWindowRect, so it effectively is ignored when passed as a capability (see: https://github.com/w3c/webdriver/issues/1792).

However, per @whimboo:

> In Firefox we match that capability and fail if the value is not appropriate

I have not tested what other implementations do.

whimboo:
Matching on our side takes place as:

https://searchfox.org/mozilla-central/rev/9e5b13e92b50c61e60a42bdc2cb2e5966ae28795/remote/shared/webdriver/Capabilities.sys.mjs#560-573

Note scenarios like window manipulation tests that want to position or resize browser windows. If you would run these tests on desktop it should basically work in nearly all the cases (exception would be Wayland) but on Android it should basically fail.

With matching clients can request that this feature is present to actually run all the tests on that given platform and if it's not supported the session will not be created.

[wpt.fyi shows the difference](https://wpt.fyi/results/webdriver/tests/classic/set_window_rect/set.py?label=master&label=experimental&product=firefox_android&product=chrome_android&aligned&view=subtest) between Firefox on Android and Chrome on Android. For Chrome an assertion only fails without any details that the command actually doesn't work, while for Firefox an unsupported operation is raised.

With matching the capability we would get a clear failure right in `New Session` for all browsers.

OrKoN:
Chrome does not do matching for this capability and it is a return-only value: https://source.chromium.org/chromium/chromium/src/+/main:chrome/test/chromedriver/session_commands.cc;l=189;drc=21802bc29a3f1f96a333bc52039dc56764a85524

whimboo:
@shs96c @AutomatedTester @titusfortner @jimevans from the Selenium side of view how would you expect a remote end to behave when it doesn't support window manipulation? Should matching the `setWindowRect` capability fail if it's value in the requested capabilities is set to `true`? Or would you expect all the tests trying to manipulate the window to fail because the requested operation cannot be achieved?

gsnedders:
It likewise is return-only in safaridriver.

jimevans:
[...]

PR #1795, opened 2024-02-15
Handle setWindowRect in matching algorithm

OrKoN:
Closes https://github.com/w3c/webdriver/issues/1793
WPT: https://github.com/web-platform-tests/wpt/pull/44602

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/pull/1795.html" title="Last updated on Feb 15, 2024, 9:56 AM UTC (57a0467)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1795/a04dbf7...57a0467.html" title="Last updated on Feb 15, 2024, 9:56 AM UTC (57a0467)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2106,6 +2106,11 @@ with a "<code>moz:</code>" prefix:
        used. The intent is that if this is not possible a new session
        will not be established.
 
+     <dt>"<code>setWindowRect</code>"
+     <dd><p>return <a>success</a> with data <a><code>null</code></a>.
+
+      <p class=note>Matching is not supported for this capability.
+
      <dt><strong>Otherwise</strong>
        <dd>
          <ul>


index.html
```diff
@@ -2106,6 +2106,11 @@ <h3>Processing capabilities</h3>
        used. The intent is that if this is not possible a new session
        will not be established.
 
+     <dt>"<code>setWindowRect</code>"
+     <dd><p>return <a>success</a> with data <a><code>null</code></a>.
+
+      <p class=note>Matching is not supported for this capability.
```
whimboo, line 2112: The referenced issue actually requests matching, and I've given an example with some more details. So I think that not matching the capability is not what we probably want.

I would suggest to get it discussed on the issue which direction we want to go before continuing here?

PR #1797, opened 2024-02-23
Add autofill endpoints and processing model

### LINKED ISSUE #1796: Add ability to save autofill data and trigger autofill on form elements
## Problem
Currently there is no way for developers to automate autofill in browsers and thus they can't test their HTML with the browser's autofill behaviour. Adding such an API would allow such testing and enable developers to write automated autofill tests (e.g. WPT).

## Proposal
We are proposing to add 2 new end points to webdriver:

#### Endpoint /session/{session id}/autofill/save
This endpoint can be used to save an address in the autofill manager of the browser which is necessary to set up any automated tests. It should take a JSON object as input with the keys being the [field name](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill-field) and the values being the actual value for that field.

#### Endpoint /session/{session id}/autofill/trigger
This endpoint will be used to trigger the autofill behaviour on a specific input field, similar to how a user would do it when using the browser manually. It should take the field's `id` attribute as parameter to specify which input field autofill should be triggered on.

yoavweiss:
This is a rough sketch PR to demonstrate how one can resolve #1796

As this is our first time contributing to WebDriver, feedback on the overall approach, as well as the details is more than welcome! :)

Co-authored by @theindra

***
<a href="https://pr-preview.s3.amazonaws.com/yoavweiss/webdriver/pull/1797.html" title="Last updated on May 6, 2024, 11:49 AM UTC (c6c07c2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1797/a04dbf7...yoavweiss:c6c07c2.html" title="Last updated on May 6, 2024, 11:49 AM UTC (c6c07c2)">Diff</a>

- - -
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1797)

### DIFF
--- a/index.html
+++ b/index.html
@@ -10865,7 +10865,7 @@ A <a>browsing context</a> has a <dfn>autofill store for testing</dfn>, an
  </tr>
  <tr>
   <td>POST
-  <td>/session/{<var>session id</var>}/autofill/trigger
+    <td>/session/{<var>session id</var>}/autofill/{<var>element id</var>}/trigger
  </tr>
 </table>
 
@@ -10874,8 +10874,15 @@ A <a>browsing context</a> has a <dfn>autofill store for testing</dfn>, an
 <ol>
  <li><p>If the <a>current top-level browsing context</a> is <a>no longer open</a>,
   return <a>error</a> with <a>error code</a> <a>no such window</a>.
+
  <li><p>Let <var>element</var> be the result of <a>trying</a> to
-  <a>deserialize a web element</a> given request body.
+ <a>get a known element</a> given <var>session id</var> and <var>element id</var>.
+
+ <li><p>If <var>element</var> is not a <a data-cite="HTML#form-associated-element">form
+    associated</a> or the <var>element</var>'s <a data-cite="HTML#form-owner">form
+    owner</a> is null, return an <a>error</a> with <a>error code</a> <a>invalid
+    element state</a>.
+                                                     
  <li><p>The <a>user agent</a> should <a data-cite="HTML#concept-fe-autofill">autofill</a>
   <var>element<var> and <var>element</var>'s <a data-cite="HTML#form-owner">form owner</a>,
   while taking into account the contents of the <a>current top-level browsing context</a>'s
--- a/index.html
+++ b/index.html
@@ -1039,6 +1039,18 @@ when it receives a particular <a>command</a>.
   <td>/session/{<var>session id</var>}/print</td>
   <td><a>Print Page</a></td>
  </tr>
+
+ <tr>
+  <td>POST</td>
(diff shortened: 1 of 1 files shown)


sadym-chromium:
Please note that webdriver classic is in maintenance mode, while new features are expected to be added to [webdriver bidi](https://w3c.github.io/webdriver-bidi/).

index.html
```diff
@@ -10782,6 +10794,96 @@ <h3><dfn>Print Page</dfn></h3>
...
+  <td>/session/{<var>session id</var>}/autofill/trigger
+ </tr>
+</table>
+
+<p>The <a>remote end steps</a> are:
+
+<ol>
+ <li><p>If the <a>current top-level browsing context</a> is <a>no longer open</a>,
+  return <a>error</a> with <a>error code</a> <a>no such window</a>.
+ <li><p>Let <var>element</var> be the result of <a>trying</a> to
```
theindra, line 10877: @yoavweiss  `element` should be derived from a specified `elementId`, right?

yoavweiss: Not necessarily. IIUC (and it's highly likely I don't), [Web elements](https://www.w3.org/TR/webdriver2/#dfn-web-elements) can be represented using their [node ID](https://www.w3.org/TR/webdriver2/#dfn-webdriver-node-id).

index.html
```diff
@@ -10782,6 +10794,96 @@ <h3><dfn>Print Page</dfn></h3>
...
+  <a>deserialize a web element</a> given request body.
+ <li><p>The <a>user agent</a> should <a data-cite="HTML#concept-fe-autofill">autofill</a>
```
AutomatedTester, line 10879: What happens if the page doesn't have a form in it. Currently this looks like it should attempt to do the right thing but the user is never given any feedback.

index.html
AutomatedTester, line 10877: Since this for a web element it should probably be in the URL like other element commands and we should change the line to be more like the following to match the rest of the spec.
```suggestion
 <li><p>Let <var>element</var> be the result of <a>trying</a> to
 <li><p>Let element be the result of [trying](https://w3c.github.io/webdriver/#dfn-try) to [get a known element](https://w3c.github.io/webdriver/#dfn-get-a-known-element) with session and element id.
```

index.html
yoavweiss: I can add a check to enforce the existence of a form owner (and throw if there isn't one?)

index.html
yoavweiss: Done

index.html
yoavweiss: Ended up returning an error. Please make sure I did that correctly :)

PR #1799, opened 2024-02-26
Update pointer up action process

jrandolf:
This PR updates the pointer up action process to reflect the conclusions in https://github.com/w3c/webdriver-bidi/pull/667

***
<a href="https://pr-preview.s3.amazonaws.com/jrandolf/webdriver/pull/1799.html" title="Last updated on Feb 26, 2024, 6:05 PM UTC (f32e26f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1799/a65f9fb...jrandolf:f32e26f.html" title="Last updated on Feb 26, 2024, 6:05 PM UTC (f32e26f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -8394,6 +8394,10 @@ and <var>action</var>:</p>
  <li><p>Set the <code>button</code> property of <var>action</var>
   to <var>button</var>.
 
+ <li><p>If <var>action item</var> has subtype property set to
+ "<code>pointerUp</code>", return success with
+ data <a><code>null</code></a>.
+
  <li><p>Let <var>width</var> be the result
   of getting the property <code>width</code>
   from <var>action item</var>.


index.html
```diff
@@ -8394,6 +8394,10 @@ <h3>Processing actions</h3>
  <li><p>Set the <code>button</code> property of <var>action</var>
   to <var>button</var>.
 
+ <li><p>If <var>action item</var> has subtype property set to
+ "<code>pointerUp</code>", return success with
+ data <a><code>null</code></a>.
```
whimboo, line 8399: It's still not clear to me. The [`pointerUp` event](https://developer.mozilla.org/en-US/docs/Web/API/Element/pointerup_event) has all these common pointer properties as well. So why shall we return early and not correctly handle all the properties instead?

jrandolf: The pointer up event doesn't use any of the below properties for emission.

OrKoN (approved):
LGTM since the properties are not used by the dispatch spec https://pr-preview.s3.amazonaws.com/w3c/webdriver/1799/a65f9fb...jrandolf:f32e26f.html#dfn-dispatch-a-pointerup-action

index.html
jgraham: The functional difference is that we don't even validate these properties. So `{type: "pointerUp", button: 0, width: "invalid value"}` wouldn't throw.

But that seems more or less OK to me; it's similar to what we do for other unused properties (e.g. if you provided a `context: []` property we wouldn't notice that it's not valid as a context). However it's a bit of a shift from the original model where all pointer types were treated as more or less interchangable. Probably if we'd started with this model from the start we'd just have ended up with `mouseDown`, `touchDown` and `penDown` rather than `pointerDown`.

From an author's point of view, neither approach helps you catch the error where you get the wrong pointer type and end up sending well-formed (say) pen inputs as a mouse pointer type, with most of the properties being ignored.

jrandolf:
Closing in favor of https://github.com/w3c/webdriver/pull/1800

PR #1800, opened 2024-03-04
Update pointer option processing

jrandolf:
For specific pointer types, some properties are not supported and should not be configurable and must be set to the default recommended in https://w3c.github.io/pointerevents/#pointerevent-interface (not the interface definition; each property has a default written in their description).


OrKoN (approved):
LGTM (to only validate properties that can be set on the particular event type, and to set the default values based on other specs instead of defining WebDriver-specific defaults). Are there any WPT changes needed?

jrandolf:
> LGTM (to only validate properties that can be set on the particular event type, and to set the default values based on other specs instead of defining WebDriver-specific defaults). Are there any WPT changes needed?

Yes, several WPT tests need to be changed.

index.html
```diff
@@ -8555,113 +8624,164 @@ <h3>Processing actions</h3>
...
+ <li><p>Let <var>pointerType</var> be equal to
+  <var>action</var>’s <code>pointerType</code> property.
+  
+ <li><p>Run the substeps of the first matching value of <var>pointerType</var>
 
- <li><p>If <var>azimuthAngle</var> is not <a>undefined</a>
-  and <var>azimuthAngle</var> is not a <a>Number</a> greater than
-  or equal to 0 and less than or equal to 2π return <a>error</a> with
-  <a>error code</a> <a>invalid argument</a>.
+  <dl class=switch>
```
jgraham, line 8632: This switch needs to be exhaustive i.e. you need a `"<code>mouse</code>"` option, even if it's a no-op.

jgraham (commented):
I wonder if @gsnedders should look at this? I think Safaridriver's implementation is nonstandard in that touch inputs will fall back to mouse inputs if there isn't touch support. If the opposite is also true (i.e. mouse becomes touch on devices with only touch input) this might cause problems.

gsnedders:
> I wonder if @gsnedders should look at this? I think Safaridriver's implementation is nonstandard in that touch inputs will fall back to mouse inputs if there isn't touch support. If the opposite is also true (i.e. mouse becomes touch on devices with only touch input) this might cause problems.

This is true, both ways around:

https://github.com/WebKit/WebKit/blob/fbe4a30494efb2f9a31d2a6c2866780889869737/Source/WebKit/UIProcess/Automation/WebAutomationSession.cpp#L2182-L2191

I don't think we can inject mouse events in any way on iOS, because I don't believe iOS has any support for a mouse. (Assistive Touch, as the name implies, still just dispatches touch events based on a physical mouse.) With the level we're injecting events at, we basically can't dispatch anything the OS cannot. And it turns out most existing WebDriver usage assumes there is a mouse—thus aliasing is practically a necessity for existing usage to work on a touch-only device.

AIUI, we fallback from touch to mouse because it turns out web developers ended up expecting this to work, but I don't know if it originally working was deliberate.

jrandolf:
@gsnedders Then would this PR cause of a problem? It seems like a non-issue for webkit.

PR #1805, opened 2024-03-07
Fix unreferenced definitions

jgraham:
This takes us down to zero respec warnings when bulding the spec.

We still have terms which are only referenced in non-normative contexts. That's probably also a bug, and could be fixed in a followup. However this at least provides a clean baseline for future spec work.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/pull/1805.html" title="Last updated on Mar 7, 2024, 6:45 PM UTC (48a3ef8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1805/52c3e3a...48a3ef8.html" title="Last updated on Mar 7, 2024, 6:45 PM UTC (48a3ef8)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -288,12 +288,9 @@ It is acknowledged that this is complementary to the Evil Bit [[RFC3514]].
  <dd>Intermediary nodes are those that act as proxies, implementing
   both the <a>local end</a> and <a>remote end</a> of
   the <a href=#protocol>protocol</a>.  However they are not expected
-  to implement <a>remote end steps</a> directly. All nodes between a
-  specific <a>intermediary node</a> and a <a>local end</a> are said to
-  be <dfn data-lt="downstream node">downstream</dfn> of that
-  node. Conversely, any nodes between a specific <a>intermediary
-  node</a> and an <a>endpoint node</a> are said to
-  be <dfn data-lt="upstream node">upstream</dfn>.
+  to implement <a>remote end steps</a> directly. Nodes between
+  a specific <a>intermediary node</a> and an <a>endpoint node</a> are
+  said to be <dfn>upstream</dfn> of the <a>endpoint node</a>.
 
  <dt><dfn>Endpoint node</dfn>
  <dd>An endpoint node is the final <a>remote end</a>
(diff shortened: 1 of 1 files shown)


OrKoN (approved):

index.html
```diff
@@ -1502,21 +1500,21 @@ <h2>Capabilities</h2>
  </thead>
 
  <tr>
-  <td><dfn>Browser name</dfn>
+  <td>Browser name
```
whimboo, line 1503: Below we use `<code>` for the individual entries. Is there a reason why that one and others are not wrapped?

```suggestion
  <td><code>Browser name</code>
```

index.html
```diff
@@ -5746,24 +5739,24 @@ <h4><dfn>Get Element Rect</dfn></h4>
 The returned value is a dictionary with the following members:
 
 <dl>
-<dt><dfn data-lt="elementrect-x">x</dfn>
+<dt>x
```
whimboo, line 5742: We probably want to use `<code>` here as well. But this all might go into a different PR.

index.html
```diff
@@ -11105,19 +11098,23 @@ <h2>Element displayedness</h2>
...
+<p>The <dfn>element displayed state</dfn> is a boolean representing
+whether an element is currently visible.
+
+<p>To get the <a>element displayed state</a> using
+the <a><code>bot.dom.isShown</code></a> Selenium atoms,
+given <var>element</var>:
+ <ol>
+  <li><p>Let <var>result</var> be
+  <a>Function.[[\Call]]</a>(<a><code>null</code></a>, <var>element</var>,
+  <code>false</code>), with <a><code>bot.dom.isShown</code></a> as the
```
whimboo, line 11110: I wonder what the `false` argument stands for here. Nowadays the atoms get called with just the element.

whimboo (changes requested):

index.html
jgraham: I don't agree with this one; they're not code strings, they're just names. The proxy table below is different in that the first column is supposed to represent actual strings. We could make the tables more similar but I think that should be out of scope for this PR.

whimboo: Ah, good point. You are right here.

whimboo (approved):

PR #1807, opened 2024-04-05
Clarify when an element is disabled

whimboo:
Selenium's atom "Is Element Enabled" special-cases the disabled state for "option" and "optgroup" elements when the nearest anchestor "select" is disabled as well.

@jgraham please let me know if that is what you would imagine based on your comment from:
https://bugzilla.mozilla.org/show_bug.cgi?id=1863266#c12

***
<a href="https://pr-preview.s3.amazonaws.com/whimboo/webdriver/pull/1807.html" title="Last updated on Apr 16, 2024, 12:51 PM UTC (8f951ba)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1807/43903d0...whimboo:8f951ba.html" title="Last updated on Apr 16, 2024, 12:51 PM UTC (8f951ba)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4505,8 +4505,8 @@ and <var>element</var> is:
 <dl class=subcategories>
  <dt><dfn data-lt="mutable form control element">Mutable form control elements</dfn>
  <dd><p>Denotes [^input^] elements
-  that are <a>mutable</a> (e.g. that are not <a>read only</a> or <a>actually disabled</a>)
-  and whose [^input/type^] attribute
+  that are <a>mutable</a> (e.g. that are not <a>read only</a> or
+  <a>disabled</a>) and whose [^input/type^] attribute
   is in one of the following states:
 
   <ul class=brief>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -4603,6 +4603,11 @@ <h3>Interactability</h3>
  <li><p>Return the pair of (<var>x</var>, <var>y</var>).
 </ol>
 
+<p>An <a>element</a> is <dfn>disabled</dfn> when it is an <a>element</a>
```
jgraham, line 4606: That's not quite what your gecko patch does; it looks for any disabled `<select>` or `<optgroup>` ancestor, not just the nearest `<select>` ancestor. The gecko patch seems correct to me, since it makes the `<select disabled><optgroup><option>` and `<select><optgroup disabled><option>` cases work.

I'd be somewhat tempted to write this as an algorithm, like

```
An element |element| is <dfn>disabled</dfn> if the following steps return true:

1. If |element| is an <{option}> element or |element| is an <{optgroup}> element:

  1. For each [=inclusive ancestor=] |ancestor| of |element|:

    1. If |ancestor| is an <{optgroup}> element or |ancestor| is a <{select}> element, and |ancestor| is [=actually disabled=], return true.

  1. Return false.

1. Return|element| is [=actually disabled=].
```

jgraham (changes requested):

index.html
whimboo: Oh, I actually forgot to update the PR here after applying the changes to the Gecko patch. So yes the `optgroup` needs to be added here as well.

Your suggestion looks fine but would require some small updates for the references. Does it look fine to you now? Whom else do we have to ask for review?

whimboo:
There is currently a bug in the Actions workflow for `action-setup@v3` which breaks our build. I've filed https://github.com/w3c/spec-prod/issues/178 to get this fixed.

AutomatedTester (approved):

whimboo:
As per Matrix @jgraham is fine with the rest of the changes. And thanks to @AutomatedTester we got also feedback from the Selenium side. So lets get this PR merged.

jgraham (approved):

PR #1791, opened 2024-02-09
User prompt handler capabilities

jgraham:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/pull/1791.html" title="Last updated on Mar 11, 2024, 4:01 PM UTC (6a39106)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1791/6c86972...6a39106.html" title="Last updated on Mar 11, 2024, 4:01 PM UTC (6a39106)">Diff</a>

PR #1812, opened 2024-05-16
Fix issues with User Prompt Handler after PR #1791

whimboo:
With PR #1791 we got extended support for the `unhandledPromptBehavior` landed which added a new Object type to define handlers for all the prompt handlers more granularly.

While implementing the capability handling for that new type in Firefox I noticed a couple of issue. This PR fixes these and also updates some cases to make it better understandable what specific variables actually are.

FYI @sadym-chromium @nechaev-chromium @gsnedders for when you get started to implement these changes in the drivers.

Updated tests will be part of my work on https://bugzilla.mozilla.org/show_bug.cgi?id=1895738.

***
<a href="https://pr-preview.s3.amazonaws.com/whimboo/webdriver/pull/1812.html" title="Last updated on May 21, 2024, 9:01 PM UTC (76573d3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1812/3b12477...whimboo:76573d3.html" title="Last updated on May 21, 2024, 9:01 PM UTC (76573d3)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -10510,6 +10510,9 @@ argument <var>value</var>:
    set <var>handler<var> to "<code>dismiss</code>"
    and <var>notify</var> to true.
 
+   <li><p>If <var>handler</var> is "<code>ignore</code>",
+   set <var>notify</var> to true.
+
    <li><p>Let <var>configuration</var> be a <a>prompt handler
    configuration</a> with [=prompt handler
    configuration/handler=] <var>handler</var> and [=prompt handler
@@ -10659,19 +10662,18 @@ given <var>requested prompt handler</var>:
  on <var>handler</var>&apos;s [=prompt handler configuration/handler=]:
 
 <dl class=switch>
-<dt>"<code>accept</code>"
-<dd><p><a>Accept</a> the <a>current user prompt</a>.
+ <dt>"<code>accept</code>"
+ <dd><p><a>Accept</a> the <a>current user prompt</a>.
 
-<dt>"<code>dismiss</code>"
-<dd><p><a>Dismiss</a> the <a>current user prompt</a>.
+ <dt>"<code>dismiss</code>"
+ <dd><p><a>Dismiss</a> the <a>current user prompt</a>.
 
-<dt>"<code>ignore</code>"
-<dd><p>Do nothing.
+ <dt>"<code>ignore</code>"
+ <dd><p>Do nothing.
 </dl>
 
-<li><p>If <var>handler</var>&apos;s [=prompt handler
-   configuration/notify=] is true, return <a>annotated unexpected
-   alert open error</a>.
+<li><p>If <var>handler</var>&apos;s [=prompt handler configuration/notify=]
+is true, return <a>annotated unexpected alert open error</a>.
 
 <li><p>Return <a>success</a>.
 </ol>
--- a/index.html
+++ b/index.html
@@ -10484,6 +10484,9 @@ argument <var>value</var>:
  <li><p>If <var>value</var> is a <a>string</a> set <var>value</var> to
  the <a data-cite=infra>map</a> «["<code>default</code>" → <var>value</var>]».
 
+ <li><p>If <var>value</var> is not an <a>Object</a> return <a>error</a>
+ with <a>error code</a> <a>invalid argument</a>.
+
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -10484,6 +10484,9 @@ <h3>User Prompt Handler</h3>
  <li><p>If <var>value</var> is a <a>string</a> set <var>value</var> to
  the <a data-cite=infra>map</a> «["<code>default</code>" → <var>value</var>]».
 
+ <li><p>If <var>value</var> is not an <a>Object</a> return <a>error</a>
```
jgraham, line 10487: It can't be an Object here since we converted to infra values, but it could be an infra Map and indeed not being a Map would be an error.

index.html
```diff
@@ -10559,20 +10562,21 @@ <h3>User Prompt Handler</h3>
...
+   to <var>request handler</var>.
   </ol>
 
 </ol>
 
 <p>To <dfn>serialize the user prompt handler</dfn>:
 
 <ol class=algorithm>
- <li><p>If the <a>user prompt handler</a> is null, return null.
+ <li><p>If the <a>user prompt handler</a> is null, return
```
jgraham, line 10578: This is a behaviour change, right? I don't disagree with making this change if it aligns with more implementations, but we shouldn't sneak it in in a bugfix PR. Also you need to update the caller to not handle null if we're always returning a string.

index.html
```diff
@@ -10656,19 +10660,19 @@ <h3>User Prompt Handler</h3>
...
-<dt>"<code>ignore</code>"
-<dd><p>Do nothing.
+ <dt>"<code>ignore</code>"
+ <dd><p>Do nothing.
 </dl>
 
-<li><p>If <var>handler</var>&apos;s [=prompt handler
-   configuration/notify=] is true, return <a>annotated unexpected
-   alert open error</a>.
+<li><p>If <var>handler</var>&apos;s [=prompt handler configuration/notify=]
```
jgraham, line 10673: Returning unexpected alert if `notify` isn't true seems like it misses the point of `notify`.

jgraham (changes requested):

index.html
whimboo: Good point! I completely overlooked that even through the map is used around those lines already.

index.html
whimboo: This is the change that we discussed on Matrix. If there is no `unhandledPromptBehavior` capability specified we won't return anything. But at the moment both Firefox and Chrome return the default which is `dismiss and notify`. 

If we are sure that we absolutely don't want to return the default value if the capability wasn't set when calling the new session command, I'm happy to revert that and add a test to check that more strict behavior (which aligns to what Safari is doing).

index.html
whimboo: The `ignore` handler causes all prompts to not be closed, but an error reported. Should a new case be allowed for not raising an error? This would require quite a lot of changes especially if we want to keep the backward compatibility with classic. Not sure if that is really doable.

index.html
jgraham: I think the point is that you should split that change into a separate PR and make sure we get input from other vendors, since it's not just an editorial change.

index.html
jgraham: In that case set /notify/ to true when we get "ignore" as the capability value, rather than adding a special case here.

whimboo: Oh right. That is indeed a better approach.

index.html
whimboo: Ok. I'm going to remove this changeset then and will file a new issue to discuss if we should return the capability or not for the default case.

jgraham (approved):

index.html
```diff
@@ -10507,13 +10510,17 @@ <h3>User Prompt Handler</h3>
    set <var>handler<var> to "<code>dismiss</code>"
    and <var>notify</var> to true.
 
+   <li><p>If <var>handler</var> is "<code>ignore</code>",
```
OrKoN, line 10513: So if the handler is already "ignore", we set it to "ignore" again?

whimboo: Ah good point. I should not have simply copy and pasted. Yes, we only need to set `notify` to `true` here.

PR #1818, opened 2024-06-12
Don't require user prompt handler notify to match when creating session

jgraham:
This would prevent creating a BiDi session for a browser that already has a HTTP session with the notify flag set for some prompt handler.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/pull/1818.html" title="Last updated on Jun 14, 2024, 9:53 AM UTC (d95609c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1818/34f9efa...d95609c.html" title="Last updated on Jun 14, 2024, 9:53 AM UTC (d95609c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -10545,11 +10545,7 @@ given <var>requested prompt handler</var>:
      <li><p>If the <var>requested prompt handler</var>&apos;s [=prompt
      handler configuration/handler=] is not equal to the <a>user
      prompt handler</a>&apos;s [=prompt handler
-     configuration/handler=], or the <var>requested prompt
-     handler</var>&apos;s [=prompt handler configuration/notify=]
-     is not equal to the <a>user prompt
-     handler</a>&apos;s [=prompt handler configuration/handler=],
-     return false.
+     configuration/handler=], return false.
 
     </ol>
    </li>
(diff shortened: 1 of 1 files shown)


OrKoN (approved):

index.html
```diff
@@ -10545,11 +10545,7 @@ <h3>User Prompt Handler</h3>
      <li><p>If the <var>requested prompt handler</var>&apos;s [=prompt
      handler configuration/handler=] is not equal to the <a>user
      prompt handler</a>&apos;s [=prompt handler
-     configuration/handler=], or the <var>requested prompt
-     handler</var>&apos;s [=prompt handler configuration/notify=]
-     is not equal to the <a>user prompt
-     handler</a>&apos;s [=prompt handler configuration/handler=],
-     return false.
+     configuration/handler=], return false.
```
whimboo, line 10548: So, removing this extra check for `notify` means that for a classic session and multiple `alwaysMatch` fields of the requested capabilities the value for `unhandledPromptBehavior` could vary between `accept/dismiss` and `accept/dismiss and notify` without causing a mismatch, right?

jgraham: It doesn't apply in that case, this is only matching against the current configured value. If you're starting a new session the current configured value is always null.

whimboo: Ah, right. That's perfect then! Thanks.

whimboo (approved):

PR #1830, opened 2024-07-16
Allow ignoring position or dimensions in set window rect

jgraham:
Previously it was unclear if it was acceptable to return without error from set window rect if the implementation is unable to set some of the properties.

This updates the spec so that:

* It's unambiguously OK to run the command even if you don't support some aspects of the functionality on the current platform / configuration (you are expected to continue to return an error if you don't support any aspect of the functionality).

* It's clear that both position and dimension properties are best effort, and even if you are able to adjust them there's no guarantee that you get the exact requested values.

* It's clear what to do if you get just an x coordinate but not y, or just width but not height, or vice-versa. Previously this was undefined.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/pull/1830.html" title="Last updated on Jul 16, 2024, 8:41 AM UTC (dae8395)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1830/af51973...dae8395.html" title="Last updated on Jul 16, 2024, 8:41 AM UTC (dae8395)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4044,25 +4044,23 @@ variables</var> and <var>parameters</var> are:
 <ol>
  <li><p>Let <var>width</var> be the result of
   <a>getting a property</a> named "<code>width</code>"
-  from <var>parameters</var>, else let it be <a><code>null</code></a>.
+  from <var>parameters</var></a>.
 
  <li><p>Let <var>height</var> be the result of
   <a>getting a property</a> named "<code>height</code>"
-  from <var>parameters</var>, else let it be <a><code>null</code></a>.
+  from <var>parameters</var>.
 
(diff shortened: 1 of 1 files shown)


whimboo (approved):
Thanks James! The changes look good to me.

OrKoN:
cc @nechaev-chromium

OrKoN (approved):

jgraham:
After merging this I noticed we have some uniformly passing tests that assume that input like `{x: 200}` is just silently ignored rather than either producing an error or setting the x-coordinate to 200 (where possible). Silently ignoring broken input seems like bad design, but I guess I should revert that part on the basis that's the defacto behviour.

Issue #1792, opened 2024-02-14
Output-only capabilities


gsnedders:
Split from discussion on https://github.com/w3c/webdriver-bidi/issues/446 (c.f. https://github.com/w3c/webdriver/pull/1790#issuecomment-1935920012).

Not everything in https://w3c.github.io/webdriver/#capabilities has defined matching behaviour under https://w3c.github.io/webdriver/#dfn-matching-capabilities.

If I pass `setWindowRect` as an `alwaysMatch` capability, what should behaviour be?

Currently, under "For each name and value corresponding to capability’s own properties:", this falls through out of the "Otherwise" clause because:

 * "If name is the name of an additional WebDriver capability" doesn't apply because an additional WebDriver capability is one defined in another spec, and setWindowRect is defined in this spec

 * "Otherwise, if name is the key of an extension capability" also obviously doesn't apply

This effectively means that we end up not doing any matching and we also don't fail as a result, and the returned value may well be different to the `alwaysMatch` value.

From a protocol point-of-view, it is _exceptionally_ surprising that we end up with the returned value being different to the `alwaysMatch` value.

(Note I'm not super interested in `setWindowRect` specifically here—and indeed this is arguably a bug—but we do want to allow for other output-only capabilities, and we should define *in general* what should happen when they are passed in as a capability.)

whimboo:
CC @jgraham

PR #1833, opened 2024-07-23
Add a capability for returning the main PID of the browser

### LINKED ISSUE #1823: Exposing the PID of the browser process from the remote end to the local end - in Capabilities?
Hi!

We'd love to be able to get the main PID of the browser process back from WebDriver implementations (like chromedriver) in order to test the exposed accessibility API on the browser. We need the PID in order to reliably find the the accessibility API of the browser through the platform's accessibility API (otherwise, we would have to find the application by name, which is tricky if there are more than one Firefoxes open). We only need the main PID of the browser in all cases.

We'd like to get this information from WebDriver implementations like chromedriver specifically because it would allow us to use WebDriver in our test suite. We would use WebDriver to start the browser, load test pages, and interact with the browser (like clicking buttons or running scripts), then separately query the browser through the accessibility API.

For some context: accessibility APIs are what assistive technologies like screen readers use to interact with applications like web browsers. We are working on extending WPT with tests that test the html -> accessibility API mappings. If you are curious to understand more, you can read this explainer from my colleague @alice: https://github.com/Igalia/wpt/blob/explainer/docs/wpt-for-aams.md

I'm new to reading the WebDriver spec, but it seems to me that a process ID can be returned in the Capabilities object returned when starting a new Session. This might not makes sense to do when using WebDriver over a network. In our case, we can only query the accessibility API of a browser on the same local machine.

So my questions:
- Can we return the PID in the Capabilities object?
- Is there a better way to return the PID (like through a request)?
- Is it ok to specify to return the PID only if the Session request came from localhost?

Looking at the code for the chromedriver implementation, I think this won't be hard to do, and I would be happy to supply a patch if this change is support! :)

spectranaut:
Closes https://github.com/w3c/webdriver/issues/1823

As discussed in the issue, returning the main PID of the browser will allow us to do accessibility testing of platform accessibility APIs exposed by the browser. Otherwise, we have to rely on the browser name, which is less reliable.

This value obviously cannot be used for matching and only is an returned value.

It doesn't look like there is precedent for this, but I wonder if this should be optional capability, in case there is a reason to no return the PID?

***
<a href="https://pr-preview.s3.amazonaws.com/Igalia/webdriver/pull/1833.html" title="Last updated on Jul 24, 2024, 4:47 PM UTC (172882a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1833/7a23ea0...Igalia:172882a.html" title="Last updated on Jul 24, 2024, 4:47 PM UTC (172882a)">Diff</a>

- - -
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1833)

### DIFF
--- a/index.html
+++ b/index.html
@@ -1578,6 +1578,13 @@ with a "<code>moz:</code>" prefix:
   <td>string
   <td>Identifies the <a>default User-Agent value</a> of the <a>endpoint node</a>.
  </tr>
+
+ <tr>
+  <td>ProcessID
+  <td>"<code>processID</code>"
+  <td>string
+  <td>Identifies the main process ID of the <a>endpoint node</a>.
+ </tr>
 </table>
 
 <section>
@@ -2009,6 +2016,9 @@ with a "<code>moz:</code>" prefix:
 
    <dt>"<code>userAgent</code>"
    <dd>String containing the <a>default User-Agent value</a>.
+
+   <dt>"<code>processID</code>"
+   <dd>The main PID of the user agent as a <a>number</a>.
   </dl>
 
  <li><p>If <var>flags</var> contains "<code>http</code>", add the


index.html
```diff
@@ -2009,6 +2016,9 @@ <h3>Processing capabilities</h3>
 
    <dt>"<code>userAgent</code>"
    <dd>String containing the <a>default User-Agent value</a>.
+
+   <dt>"<code>processID</code>"
```
spectranaut, line 2020: For example, this could be:

```suggestion
   <dt>"<code>processID</code> (optional)"
```

alice: Right; do we want to say anything here about suggesting that the processID be included iff the client is on localhost?

spectranaut:
@whimboo and  @gsnedders -- can you review this change based on our discussion? :)

jgraham (changes requested):
This PR is currently only affecting the non-normative parts of the spec. To actually return a value in capabilities you need to set it in the [matching capabilities](https://w3c.github.io/webdriver/#dfn-matching-capabilities) algorithm.

spectranaut:
> This PR is currently only affecting the non-normative parts of the spec. To actually return a value in capabilities you need to set it in the [matching capabilities](https://w3c.github.io/webdriver/#dfn-matching-capabilities) algorithm.

Ah thanks @jgraham for pointing this out. I was wondering myself about that, but I looked at the previous capability added (https://github.com/w3c/webdriver/pull/1790) and it also didn't update the matching algorithm. I wonder what the point of this previous PR was? Did it also make a mistake?

I also have a question -- and sorry if it's a silly one, I've only used webdriver minimally -- it seems to me, according to [matching capabilities step 4 and later](https://w3c.github.io/webdriver/#dfn-matching-capabilities), the returned capabilities seems to only include keys that were specifically set in the new session command? Is that true?

I'm pretty sure that returning the PID should be optionally, and isn't not need and shouldn't be returned in all cases. So I'm trying to wrap my head around how to "optionally" return the PID. The other option is to use a extension capabilities to do this, like firefox does, returning "moz:processID".

gsnedders:
> Ah thanks @jgraham for pointing this out. I was wondering myself about that, but I looked at the previous capability added (https://github.com/w3c/webdriver/pull/1790) and it also didn't update the matching algorithm. I wonder what the point of this previous PR was? Did it also make a mistake?

The addition at lines 2011–3 does that, as that is within the matching algorithm.

#1792 is related here (don't expect you to solve this, just want the cross-reference!).

PR #1847, opened 2024-10-22
Input events dispatch to top-level frame

sadym-chromium:
As discussed in https://github.com/w3c/webdriver-bidi/issues/795, the actions should be dispatched from the top-level browsing context.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/pull/1847.html" title="Last updated on Oct 24, 2024, 8:08 AM UTC (adf94ad)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1847/db06684...adf94ad.html" title="Last updated on Oct 24, 2024, 8:08 AM UTC (adf94ad)">Diff</a>

- - -
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1847)

### DIFF
--- a/index.html
+++ b/index.html
@@ -8339,12 +8339,14 @@ from a machine utilisation perspective.
  input JSON, such that the actions to be performed in a single <a>tick</a>
  are grouped together.
 
-<p>To <dfn>get parent offset</dfn> of <var>context</var>:
+ <p>To <dfn>get parent offset</dfn> of <var>context</var>:
  <ol class="algorithm">
   <li>Let <var>offsetLeft</var> equal to 0 and <var>offsetTop</var> equal to 0.
   </li>
-  Let <var>navigable</var> be <var>context</var>'s <a>active document</a>'s
-  [=navigable/parent=].  </p></li>
+  <li>
+   Let <var>navigable</var> be <var>context</var>'s <a>active document</a>'s
+   [=navigable/parent=].
+  </li>
   <li>Let <var>parent navigable</var> be <var>navigable</var>'s parent.</li>
   <li>If <var>parent navigable</var> is not null:
    <ol>
@@ -8375,10 +8377,11 @@ from a machine utilisation perspective.
      <code>offsetTop</code>.
     </li>
    </ol>
+  </li>
   <li>Return (<var>offsetLeft</var>, <var>offsetTop</var>).</li>
-  <span class=issue>TODO: clarify if the algo respect transforms?</span>
-  <span class=issue>TODO: respect iframe's viewport's intersections</span>
-</ol>
+ </ol>
+ <span class=issue>TODO: clarify if the algo respect transforms?</span>
+ <span class=issue>TODO: respect iframe's viewport's intersections</span>
 
 <p>To <dfn>get coordinates relative to an origin</dfn>
 given <var>source</var>, <var>x offset</var>, <var>y offset</var>,
--- a/index.html
+++ b/index.html
@@ -2212,8 +2212,8 @@ sessions</a> will be the only entry.
  browsing context</dfn>, which is set to the parent of the <a>current
  browsing context</a> when changing browsing contexts, and an
  associated <dfn>current top-level browsing context</dfn>, which is
- set to the <a>top-level browsing context of</a> the <a>current browsing
- context</a>, when changing browsing contexts.</p>
+ set to the [=browsing context/top-level browsing context=] of the
+ <a>current browsing context</a>, when changing browsing contexts.</p>
 
 <p>An <a>HTTP session</a> has an associated <dfn>session
 timeouts</dfn> which is a <a>timeouts configuration</a>. This is
@@ -8286,10 +8286,10 @@ and <var>subtype</var>:
  content observable effects that must be consistent across
  implementations. To accommodate this, the specification requires
  that <a>remote ends</a> <dfn>perform implementation-specific action
- dispatch steps</dfn> on a <a>top-level browsing context of</a>
+ dispatch steps</dfn> on a [=browsing context/top-level browsing context=] of
  <var>context</var>, and a <var>list of events</var> and their properties.
  These steps must be equivalent to performing the given input device
- manipulations on the <a>top-level browsing context of</a> the
+ manipulations on the [=browsing context/top-level browsing context=] of the
  <var>context</var>, such that trusted events corresponding to the entries in
  <var>list of events</var> are dispatched.
 
@@ -11743,7 +11743,7 @@ to automatically sort each list alphabetically.
    <!-- Simple dialogs --> <li><dfn data-lt="simple dialog"><a href=https://html.spec.whatwg.org/#simple-dialogs>Simple dialogs</a></dfn>
    <!-- Steps to fire beforeunload --> <li><dfn><a href=https://html.spec.whatwg.org/#steps-to-fire-beforeunload>Steps to fire beforeunload</a></dfn>
    <!-- Suffering from bad input --> <li><dfn><a href=https://html.spec.whatwg.org/#suffering-from-bad-input>Suffering from bad input</a></dfn>
-   <!-- Top-level browsing context of --> <li><dfn><a href=https://html.spec.whatwg.org/#bc-tlbc>Top-level browsing context of</a></dfn>
+   <!-- Top-level browsing context --> <li><dfn data-dfn-for="browsing context"><a href=https://html.spec.whatwg.org/#bc-tlbc>Top-level browsing context</a></dfn>
    <!-- Traverse the history by a delta --> <li><dfn><a href=https://html.spec.whatwg.org/#traverse-the-history-by-a-delta>Traverse the history by a delta</a></dfn>
    <!-- Unfocusing steps --><li><dfn><a href=https://html.spec.whatwg.org/#unfocusing-steps>unfocusing steps</a></dfn>
    <!-- User prompt --> <li><dfn data-lt="user prompts"><a href=https://html.spec.whatwg.org/#user-prompts>User prompt</a></dfn>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -8289,8 +8289,9 @@ <h3>Ticks</h3>
  dispatch steps</dfn> on a <a>browsing context</a> <var>context</var>,
  and a <var>list of events</var> and their properties. These steps
  must be equivalent to performing the given input device manipulations
- on <var>context</var>, such that trusted events corresponding to the
- entries in <var>list of events</var>are dispatched.
+ on the top-browsing context ancestor of the <var>context</var>, such
```
OrKoN, line 8292: isn't the term "top-level browsing context"? could you link to the getter steps here https://html.spec.whatwg.org/multipage/document-sequences.html#bc-tlbc?

sadym-chromium: Thanks! Replaced both usages.

index.html
```diff
@@ -11742,6 +11743,7 @@ <h2>Index</h2>
    <!-- Simple dialogs --> <li><dfn data-lt="simple dialog"><a href=https://html.spec.whatwg.org/#simple-dialogs>Simple dialogs</a></dfn>
    <!-- Steps to fire beforeunload --> <li><dfn><a href=https://html.spec.whatwg.org/#steps-to-fire-beforeunload>Steps to fire beforeunload</a></dfn>
    <!-- Suffering from bad input --> <li><dfn><a href=https://html.spec.whatwg.org/#suffering-from-bad-input>Suffering from bad input</a></dfn>
+   <!-- Top-level browsing context of --> <li><dfn><a href=https://html.spec.whatwg.org/#bc-tlbc>Top-level browsing context of</a></dfn>
```
OrKoN, line 11746: I think we should probably leave of out of the definition and only have "top-level browsing context" as the definition

sadym-chromium: https://w3c.github.io/webdriver/#dfn-browsing-contexts:~:text=to%20capture%20the-,top%2Dlevel%20browsing%20context,-%27s%20initial%20viewport

OrKoN: I believe scoped concepts are applicable here https://github.com/speced/respec/wiki/Shorthands-Guide#scoped-concepts (example in https://github.com/w3c/webdriver/pull/1849/commits/bc1b1cfc409c527030f68ffcf49b9a26368e2746)

sadym-chromium: applied

OrKoN (approved):

sadym-chromium:
@whimboo WDYT?

OrKoN:
The following two thing seem to be still missing:

1) steps to convert the coordinates to the top-level browsing context coordinate space if they are frame scoped.
2) apply the same fix to other actions (I assume we do actually want to dispatch keyboard events through the top-level frame as well)

whimboo:
> The following two things seem to be still missing:
>     1. steps to convert the coordinates to the top-level browsing context coordinate space if they are frame scoped. Edit: unless this is covered by the implementation-defined bits?

Yes, absolutely. By changing the way where we dispatch the events the coordinates should be exactly the same. Not taking care of offsets would cause quite a lot of regressions for those users who make use of actions a lot.

Therefore see:

https://github.com/web-platform-tests/wpt/pull/48147
https://github.com/web-platform-tests/wpt/pull/48123

Given that Chrome seems to already dispatch actions in the top-level browsing context the referenced tests are failing.

>     2. apply the same fix to other actions (I assume we do actually want to dispatch keyboard events through the top-level frame as well)

Yes, we would have to do it for all the pointer and wheel input sources. 

Keyboard actions are more challenging because they could allow users to trigger shortcuts that access restricted browser features. This would require additional checks to determine what actions are permitted and which ones should be blocked. Maybe we should have a follow-up issue for it?

sadym-chromium:
> 2. apply the same fix to other actions (I assume we do actually want to dispatch keyboard events through the top-level frame as well)

Aren't all the actions end up in [**perform implementation-specific action dispatch steps**](https://w3c.github.io/webdriver/#dfn-perform-implementation-specific-action-dispatch-steps)?

sadym-chromium:
@OrKoN @whimboo I added calculation of the offset relative to the parent browsing context, but IDK how well it will work

index.html
```diff
@@ -8286,11 +8286,12 @@ <h3>Ticks</h3>
...
- and a <var>list of events</var> and their properties. These steps
- must be equivalent to performing the given input device manipulations
- on <var>context</var>, such that trusted events corresponding to the
- entries in <var>list of events</var>are dispatched.
+ dispatch steps</dfn> on a <var>context</var>, and a <var>list of events</var>
+ and their properties. These steps must be equivalent to performing the given
+ input device manipulations on the
+ [=browsing context/top-level browsing context=] of the <var>context</var>,
+ such that trusted events corresponding to the entries in
+ <var>list of events</var> are dispatched.
```
sadym-chromium, line 8294: We can "hand-wave" hear like "These steps must be equivalent to user trying to perform the given input device manipulations on context through the top-level browsing context."

OrKoN:
@sadym-chromium it looks like each action type dispatches on the context parameter https://w3c.github.io/webdriver/#dfn-dispatch-a-keydown-action 

> > 2. apply the same fix to other actions (I assume we do actually want to dispatch keyboard events through the top-level frame as well)
> 
> Aren't all the actions end up in [**perform implementation-specific action dispatch steps**](https://w3c.github.io/webdriver/#dfn-perform-implementation-specific-action-dispatch-steps)?

that's right, overlooked that.

OrKoN:
> Keyboard actions are more challenging because they could allow users to trigger shortcuts that access restricted browser features. This would require additional checks to determine what actions are permitted and which ones should be blocked. Maybe we should have a follow-up issue for it?

so actually I am not sure, even with this change it is not quite defined what it means that the actions are dispatched to the top-level browsing context. For example, in Chrome that is the case but still would not give you access to browser shortcuts. 
Perhaps we need a better definition of what the dispatch of an event means, perhaps hook into native event handles in https://www.w3.org/TR/uievents/#handle-native-mouse-move-id?

whimboo:
> so actually I am not sure, even with this change it is not quite defined what it means that the actions are dispatched to the top-level browsing context. For example, in Chrome that is the case but still would not give you access to browser shortcuts.

Yes you are right. I mixed it up with the native event dispatching that I'm working on as well right now. In those cases we would have that particular issue but not when dispatching it in the content process of the top-level browsing context.

index.html
```diff
@@ -8286,11 +8286,12 @@ <h3>Ticks</h3>
  content observable effects that must be consistent across
  implementations. To accommodate this, the specification requires
  that <a>remote ends</a> <dfn>perform implementation-specific action
- dispatch steps</dfn> on a <a>browsing context</a> <var>context</var>,
- and a <var>list of events</var> and their properties. These steps
- must be equivalent to performing the given input device manipulations
- on <var>context</var>, such that trusted events corresponding to the
- entries in <var>list of events</var>are dispatched.
+ dispatch steps</dfn> on a <var>context</var>, and a <var>list of events</var>
```
OrKoN, line 8289: I think we should keep <a>browsing context</a> indicating the context var type

index.html
```diff
@@ -8392,7 +8438,7 @@ <h3>Processing actions</h3>
    </dd>
   </dl>
 
- <li><p>Return (<var>x</var>, <var>y</var>)
+ <li><p>Return (<var>x</var> + <var>parentOffsetLeft</var>, <var>y</var> + <var>parentOffsetTop</var>)
```
OrKoN, line 8441: should the offset be applied to all origins or only the viewport? I think the origin pointer might already be in the top level context coordinates and not sure if the element origin does some adjustments already.

index.html
```diff
@@ -8338,13 +8339,58 @@ <h3>Processing actions</h3>
  input JSON, such that the actions to be performed in a single <a>tick</a>
  are grouped together.
 
+ <p>To <dfn>get parent offset</dfn> of <var>context</var>:
+ <ol class="algorithm">
+  <li>Let <var>offsetLeft</var> equal to 0 and <var>offsetTop</var> equal to 0.
+  </li>
+  <li>
+   Let <var>navigable</var> be <var>context</var>'s <a>active document</a>'s
+   [=navigable/parent=].
```
OrKoN, line 8348: [=navigable/parent=] is a property of a navigable and not the document. Did you mean to get the current navigable instead of a parent (e.g., https://html.spec.whatwg.org/#node-navigable)?

index.html
sadym-chromium: Yeah, I missed it.

index.html
OrKoN: That does not seem to follow from the spec text since input sources are per top-level browsing context.

index.html
sadym-chromium: https://github.com/w3c/webdriver/pull/1850

sadym-chromium: @OrKoN @whimboo we can keep specification implementation-specific, just clarify the events are dispatched via top-level browsing context. I'm not sure if this approach would work with the events intercepted by the top-level frame.

jgraham (changes requested):
I agree that in principle this is the model we need to change to in order to work with modern browser designs.

However I think the change as written is insufficient, and makes the spec contradictory. In each action we still claim we're dispatching events to `|context|` which, afaict after these changes, is not the parent context but the actual child context. The spec is unfortunately rather badly written here, so the text you've modified is apparently normative, but exists in what is essentially an informative section describing the overall model.

I also suggest (perhaps as a followup) that we make it possible for specific actions to provide the context that is used for calculating origins (which must in all cases be an ancestor of the top-level command context). That makes it much easier to construct action chains that interact with both the top level document and iframes.

whimboo:
We should as well clarify what should happen if a frame gets closed by an action. Following actions in the same chain should still be dispatched even through they will reach some other document and elements?

PR #1875, opened 2025-01-07
Improve support for actions involving multiple navigables

jgraham:
This makes the following changes:

* Actions are always dispatched first to the top-level traversable rather than directly to the target navigable. It is assumed that the action will end up interacting with the target traversable, but may not e.g. if there is an element obscuring the target, or the focus is in the wrong place.

* To support the above, coordinates are computed relative to the viewport of the top-level traversable rather than the target navigable.

* A new coordinate origin type is introduced which has the form: ``` { type: "viewport",
    context: <context id>
  } ```

  This is designed to allow targeting inside a specific iframe without that iframe needing to be the target of the overall actions chain.

The existing behaviour that if the main target navigable is deleted the action chain is terminated remains. This makes the behaviour consistent between top-level traversables and iframes, and also between different actions.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/pull/1875.html" title="Last updated on Jan 7, 2025, 9:34 AM UTC (1866cff)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1875/9cb696c...1866cff.html" title="Last updated on Jan 7, 2025, 9:34 AM UTC (1866cff)">Diff</a>

- - -
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1875)

### DIFF
--- a/index.html
+++ b/index.html
@@ -8093,37 +8093,36 @@ An <a>input state</a> has the following items:
 </ul>
 
 <p>To <dfn>get the input state</dfn> given <var>session</var>
-and <var>browsing context</var>:
+and <var>top-level traversable</var>:
 
 <ol class="algorithm">
- <li><p>Assert: <var>browsing context</var> is a <a>top-level browsing
+ <li><p>Assert: <var>top-level traversable</var> is a <a>top-level browsing
  context</a>.
 
  <li><p>Let <var>input state map</var>
  be <var>session</var>&apos;s <a>browsing context input state map</a>.
 
  <li><p>If <var>input state map</var> does not
- [=map/exist|contain=] <var>browsing context</var>, set <var>input
- state map</var>[<var>browsing context</var>] to <a>create an input
+ [=map/exist|contain=] <var>top-level traversable</var>, set <var>input
+ state map</var>[<var>top-level traversable</var>] to <a>create an input
  state</a>.
 
- <li><p>Return <var>input state map</var>[<var>browsing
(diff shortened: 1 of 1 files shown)


jgraham:
There are still a number of open questions here:

* The new coordinate origin is the first thing to introduce the idea of a global id for navigables (particularly iframes) into WebDriver classic. The design is a natural fit to BiDi, but there's a question about whether it's appropriate for Classic.
* Similarly the current element origin depends on the element being in the target navigable. In BiDi it's rather easy to add a context parameter to the element origin to specify where to look for the element, but that doesn't work for web elements. The alternative would be to allow searching the entire tree of navigables under the top-level traversable for the element. That would be easy to use, but might have perf implications.
* It's unclear that we're maintaining the current behaviour that are inside the viewport of an iframe but outside the top-level viewport (e.g. consider a case where a button is inside an iframe, and would be visible if you scrolled the top level document, but where the entire iframe is currently entirely out of view).
* I worry a bit about how backwards incompatible the change is. I know it's similar to what Chrome is already doing, but I want to check that there aren't any features of the Chrome implementation that aren't captured here (e.g. implicitly moving focus to the target navigable at the start of the action chain). With wheel and pointer events the fact that they might be captured by the top-level document is considered a feature, but I'm not so sure about key events.

whimboo:
@jgraham is this PR somewhat ready to get review or does it require further work?

PR #1907, opened 2025-06-27
Added note about possibility of capping the cookie expiry value

whimboo:
Remote ends can cap the expiry value to a certain amount of days in the future. For now this is only listed in https://httpwg.org/http-extensions/draft-ietf-httpbis-rfc6265bis.html#name-cookie-lifetime-limits so do we need to add a new reference?

***
<a href="https://pr-preview.s3.amazonaws.com/whimboo/webdriver/pull/1907.html" title="Last updated on Jul 1, 2025, 10:31 AM UTC (4d93620)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1907/6dc16d1...whimboo:4d93620.html" title="Last updated on Jul 1, 2025, 10:31 AM UTC (4d93620)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -7631,6 +7631,11 @@ variables</var> and <var>parameters</var> are:
    <dd><p>The value if the entry exists, otherwise leave unset to
     indicate that this is a session cookie.
 
+    <aside class=note>
+      <p>The cookie&apos;s expiry value might be limited by the remote end in
+       accordance with the <a>Cookie Lifetime Limits</a>.
+    </aside>
+
    <dt><a>Cookie same site</a>
    <dd><p>The value if the entry exists, otherwise leave unset to
     indicate that no same site policy is defined.
@@ -7638,7 +7643,6 @@ variables</var> and <var>parameters</var> are:
 
   <p>If there is an <a>error</a> during this step,
    return <a>error</a> with <a>error code</a> <a>unable to set cookie</a>.
-
 
  <li><p>Return <a>success</a> with data <a><code>null</code></a>.
 </ol>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -7631,6 +7631,12 @@ <h3><dfn data-lt="adding a cookie">Add Cookie</dfn></h3>
    <dd><p>The value if the entry exists, otherwise leave unset to
     indicate that this is a session cookie.
 
+    <aside class=note>
+      <p>The cookie’s expiry value may be limited by the remote end
+       according to an implementation-defined cap, as permitted by
+       the [[RFC6265bis]].</p>
```
jgraham, line 7637:
```suggestion
      <p>The cookie’s expiry value may be limited by the remote end
       as required by [[RFC6265bis]].</p>
```

But also, can we link to https://httpwg.org/http-extensions/draft-ietf-httpbis-rfc6265bis.html#name-cookie-lifetime-limits specifically rather than to the general RFC, I think that would make things clearer.

jgraham (changes requested):

index.html
whimboo: Updated to a direct link. Once the draft is merged we probably want to add a new ref.

whimboo:
A test will be added as part of the work on https://bugzilla.mozilla.org/show_bug.cgi?id=1974394.

sadym-chromium (approved):

whimboo:
@jgraham the latest update uses now a link to the latest samesite draft document so that we don't need an external link.

index.html
```diff
@@ -7631,6 +7631,11 @@ <h3><dfn data-lt="adding a cookie">Add Cookie</dfn></h3>
    <dd><p>The value if the entry exists, otherwise leave unset to
     indicate that this is a session cookie.
 
+    <aside class=note>
+      <p>The cookie&apos;s expiry value might be limited by the remote end in
+       accordance with the <a>Cookie Lifetime Limits</a>.
```
whimboo, line 7636: @jgraham I replaced `as required` by `in accordance` because `required` is not allowed per Bikeshed and would cause a failure when building the BiDi specification, and I think that we should have the same wording in both specs.

Issue #1470, opened 2019-12-11
Lack of normalisation in date formats


andrewnicols:
This is a bit of a weird one, so apologies in advance.

In chapter [E. Index](https://www.w3.org/TR/webdriver/#dfn-date-state#e-index):
> Date state
> Local Date and Time state

The display format for the date, and local date and time, input types is left to be implementation specific, meanwhile the value attribute is always normalised to `dd-mm-yyyy`.
The lack of standardisation makes it difficult to set the value using the element-send-keys endpoint (https://www.w3.org/TR/webdriver/#dfn-element-send-keys) as the keys sent mimic the display format.

As an example, if I set the value of a `date` element to `2012-12-25` I can do so successfully on Firefox, but Chrome uses a different date format of `dd/mm/yyyy` for display. As a result my attempt to set the value leads to a garbage result of `1225-12-20`.

I feel that there should be some part of the spec which dictates that the accepted values must be normalised in some way.

In the current state it is not possible to write a cross-browser way of setting the date using the sendKeys endpoint.

andreastt:
[Mutable form control elements](https://w3c.github.io/webdriver/#dfn-mutable-form-control-element) are special cased in the remote end steps for [Element Send Keys](https://w3c.github.io/webdriver/#element-send-keys), step 8, match clause two, substep 3:

> ↪ The user agent renders element as something other than a text input control (for example an input element in the color state being presented as a colorwheel):
>   …
>   3. Set a property value to text on element. 

(It is a bug in the above quote from the spec that the match clause doesn’t use the _mutable form control elements_ definition directly, but the effect is the same.)

This means for `<input type=date>`, drivers will modify the `value` own  property directly and simulate the necessary DOM events. The following test case shows that this produces the correct result in both Firefox and Chrome:

```html
<!doctype html>
<meta charset=utf-8>
<title>date</title>

<input type=date />

<script>
const date = document.querySelector("input");
date.value = "2019-12-11";
console.log(date.value);  // prints "2019-12-11"
</script>
```

To my big surprise [we don’t seem to have any interoperability tests](https://wpt.fyi/results/webdriver/tests/element_send_keys?label=experimental&label=master&aligned) for this, which means the behaviour you are observing with chromedriver is likely a bug.

andrewnicols:
Thanks @andreastt - I'll have a re-read of that part of the spec tomorrow (10pm here in Aus).

Should I report this as a bug against chromedriver then?

[...]

PR #1910, opened 2025-07-06
Add new category for typeable form controls with normalization

Holo-xy:
There looks to be a popular demand for normalized date and datetime-local as seen in [1470](https://github.com/w3c/webdriver/issues/1470) and [42322342](https://issues.chromium.org/issues/42322342).
This introduces a new category to make the expected behavior explicit.

***
<a href="https://pr-preview.s3.amazonaws.com/Holo-xy/webdriver/pull/1910.html" title="Last updated on Jul 6, 2025, 11:43 PM UTC (f8c808c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1910/c1a2419...Holo-xy:f8c808c.html" title="Last updated on Jul 6, 2025, 11:43 PM UTC (f8c808c)">Diff</a>

- - -
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1910)

### DIFF
--- a/index.html
+++ b/index.html
@@ -6400,6 +6400,13 @@ and a number keypad for [^input^] elements
 in the <a><code>number</code> state</a> on non-desktop devices.
 </div>
 
+<p>
+A <dfn>typeable form control with normalization</dfn> 
+is an [^input^] element 
+whose [^input/type^] attribute state 
+requires that user-provided values be normalized before being assigned to the element's value attribute.
+This applies to <code>type="date"</code> and <code>type="datetime-local"</code>.
+
 <p>The <a>key input source</a> used for input may be cleared mid-way
  through “typing” by sending the <dfn>null key</dfn>, which is U+E000
  (NULL).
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -6400,6 +6400,13 @@ <h4><dfn>Element Send Keys</dfn></h4>
 in the <a><code>number</code> state</a> on non-desktop devices.
 </div>
 
+<p>
+A <dfn>typeable form control with normalization</dfn> 
+is an [^input^] element 
+whose [^input/type^] attribute state 
+requires that user-provided values be normalized before being assigned to the element's value attribute.
```
OrKoN, line 6407: I think we should probably be more specific and define how exactly the input values should be normalized for date types. Also I wonder if it could be made optional because I imagine some existing clients would want to keep the current non-normalized behavior.

cc @whimboo @jgraham @sadym-chromium

Holo-xy: Right now the implementation accepts non-normalized values (MM-DD-YYYY format correctly), so we could keep the current implementation and just modify it to convert normalized values (YYYY-MM-DD or YYYYMMDD formats) to MM-DD-YYYY.
This way It would be techincally optional and require minmal implementation changes, but I'm not sure if this would be appropriate to add to the spec

sadym-chromium:
The approach LGTM, but the expected formats should be strictly defined.

whimboo:
> I chose to normalize to MM-DD-YYYY and MM-DD-YYYYTHH:MM:SS since that seems to be what Chromium expects as input, which keeps changes minimal, but I'm not sure if this is the correct approach.

Does Chromium not support [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601)? I think that we should follow that format.

OrKoN:
@Holo-xy I wonder if using locale overrides https://www.w3.org/TR/webdriver-bidi/#command-emulation-setLocaleOverride works around this issue? I imagine if you the locale is fixed, there could be no need for additional normalization?

PR #1916, opened 2025-08-05
Fix wrong link for first element of `getClientRects`

yezhizhen:
Previously the `element` leads us to https://dom.spec.whatwg.org/#concept-element. But it really means the first item of the array.


yezhizhen:
@whimboo Hi can you check this? Just a simple change.

PR #1927, opened 2025-09-11
Fix description of computing pointer-interactable paint tree

yezhizhen:
The step of [in-view center point](https://w3c.github.io/webdriver/#dfn-center-point) already computes the first object of `getClientRects` on element. This PR makes the description more accurate.

***
<a href="https://pr-preview.s3.amazonaws.com/yezhizhen/webdriver/pull/1927.html" title="Last updated on May 7, 2026, 9:02 AM UTC (974fd42)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1927/785bc01...yezhizhen:974fd42.html" title="Last updated on May 7, 2026, 9:02 AM UTC (974fd42)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4695,13 +4695,13 @@ and <var>element</var> is:
  is produced this way:
 
 <ol>
- <li><p>If <var>element</var> is <a>not in the same tree</a>
+ <li><p>If <a><var>element</var></a> is <a>not in the same tree</a>
   as <var>session</var>&apos;s <a>current browsing context</a>&apos;s
   <a>active document</a>, return an empty sequence.
 
  <li><p>Let <var>rectangles</var> be
   the {{DOMRect}} sequence
-  returned by calling {{Element/getClientRects()}}.
+  returned by calling {{Element/getClientRects()}} on <a><var>element</var></a>.
 
  <!--
   An element which style property `display` is "none"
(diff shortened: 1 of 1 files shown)


yezhizhen:
@whimboo Can you help check this small one too?

index.html
```diff
@@ -4686,13 +4686,13 @@ <h3>Interactability</h3>
  is produced this way:
 
 <ol>
- <li><p>If <var>element</var> is <a>not in the same tree</a>
+ <li><p>If <a><var>element</var></a> is <a>not in the same tree</a>
```
whimboo: I think that this is correct, bu in other places we use `<var><a>` so we should be consistent. 

But I wonder if we always want to link the element. There are a lot of other instances where we don't do that. So maybe a separate PR fixing all the cases might be good if we wanna have that? But that's not blocking landing this PR.

whimboo (commented):
Thanks for the fixes. That was indeed broken. Maybe you can take a look at the inline comment?

index.html
yezhizhen: That's a good point. Sometimes we call "element element", https://w3c.github.io/webdriver/#dfn-disabled
where the first element links to the Element interface.

For this one, I will just remove the link.

yezhizhen: I have refactored this same way as how we deal with disabled.

whimboo (approved):
This looks good now and matches what we have for `An element element is disabled`.

Thanks!

PR #1950, opened 2026-02-25
Add safePrintableInset parameter for printing.

mstensho:
This is for testing the `page-margin-safety` descriptor in `@page` and
page margin box contexts.

Spec: https://drafts.csswg.org/css-page-3/#page-margin-safety
Spec discussion: https://github.com/w3c/csswg-drafts/issues/11395
RFC: https://github.com/web-platform-tests/rfcs/pull/233

wptrunner and webdriver code changes:
https://github.com/web-platform-tests/wpt/pull/58030

***
<a href="https://pr-preview.s3.amazonaws.com/mstensho/webdriver/pull/1950.html" title="Last updated on Aug 14, 2026, 11:55 AM UTC (5f98639)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1950/cc4c0d2...mstensho:5f98639.html" title="Last updated on Aug 14, 2026, 11:55 AM UTC (5f98639)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -11313,6 +11313,14 @@ variables</var> and <var>parameters</var> are:
  <li><p>If <var>shrinkToFit</var> is not a <a>Boolean</a>
  return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
 
+ <li><p>Let <var>safePrintableInset</var> be the result of <a>getting a
+  property with default</a> named "<code>safePrintableInset</code>" and with
+  default <code>0</code> from <var>parameters</var>.
+
+ <li><p>If <var>safePrintableInset</var> is not a <a>Number</a>, or is a
+  negative number, return <a>error</a> with <a>error code</a> <a>invalid
+  argument</a>.
+
  <li><p>Let <var>pageRanges</var> be the result of <a>getting a
   property with default</a> named "<code>pageRanges</code>" from
   <var>parameters</var> with default of an
--- a/index.html
+++ b/index.html
@@ -11375,8 +11375,9 @@ variables</var> and <var>parameters</var> are:
    each page edge needed to steer clear of the area that the printer is potentially
    incapable of marking reliably. Most printers have a small region along each
    edge of the page sheet which is unprintable, typically due to the printer's
-   paper handling mechanism. The document that is being printed may have CSS rules
-   to adjust page margins based on this. TODO: Link to spec, when it has landed?
+   paper handling mechanism. The document that is being printed may have
+   <a href="https://drafts.csswg.org/css-page-3/#page-margin-safety">CSS rules</a>
+   to adjust page margins based on this.
 
   <li><p>If <var>pageRanges</var> is not an empty <a>Array</a>,
  Let <var>pages</var> be the result of <a>trying</a> to <a>parse a
--- a/index.html
+++ b/index.html
@@ -11317,9 +11317,8 @@ variables</var> and <var>parameters</var> are:
   property with default</a> named "<code>safePrintableInset</code>" and with
   default <code>0</code> from <var>parameters</var>.
 
- <li><p>If <var>safePrintableInset</var> is not a <a>Number</a>, or is a
-  negative number, return <a>error</a> with <a>error code</a> <a>invalid
-  argument</a>.
+ <li><p>If <var>safePrintableInset</var> is not a <a>Number</a>, or is less than 0,
+ return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
 
  <li><p>Let <var>pageRanges</var> be the result of <a>getting a
   property with default</a> named "<code>pageRanges</code>" from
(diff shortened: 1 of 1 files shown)


foolip:
@gsnedders @jgraham could you take a look?

index.html
```diff
@@ -11313,6 +11313,14 @@ <h3><dfn>Print Page</dfn></h3>
...
  return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
 
+ <li><p>Let <var>safePrintableInset</var> be the result of <a>getting a
+  property with default</a> named "<code>safePrintableInset</code>" and with
+  default <code>0</code> from <var>parameters</var>.
+
+ <li><p>If <var>safePrintableInset</var> is not a <a>Number</a>, or is a
+  negative number, return <a>error</a> with <a>error code</a> <a>invalid
+  argument</a>.
+
```
whimboo: Lets stay in sync with other params and avoid adding line breaks within a definition:

```suggestion
<li><p>If <var>safePrintableInset</var> is not a <a>Number</a>, or is less then 0, return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
```

index.html
```diff
@@ -11313,6 +11313,14 @@ <h3><dfn>Print Page</dfn></h3>
  <li><p>If <var>shrinkToFit</var> is not a <a>Boolean</a>
  return <a>error</a> with <a>error code</a> <a>invalid argument</a>.
 
+ <li><p>Let <var>safePrintableInset</var> be the result of <a>getting a
+  property with default</a> named "<code>safePrintableInset</code>" and with
+  default <code>0</code> from <var>parameters</var>.
```
whimboo: This only defines the extraction of the `safePrintableInset` parameter but I do not see how it actually affects printing. You need to define as well what should be done (see `shrinkToFit` as example).

whimboo (changes requested):
As referenced this is not yet a public available feature so we need to wait until the feature spec is done. Nevertheless here a quick feedback.

Upon agreement it might be good to as well to get it added to WebDriver BiDi.

index.html
mstensho: Ok. but not taken verbatim. Less *than* 0.

index.html
mstensho: This value is used to implement `page-margin-safety`. I added some spec text now.

For real printers in a real browser, try printing with Margins: Minimum. This is what this value is about. With `page-margin-safety`, this becomes web-exposed.

mstensho:
The spec changes have landed, and I've updated the description here accordingly. Please take a look.

index.html
```diff
@@ -11364,6 +11371,13 @@ <h3><dfn>Print Page</dfn></h3>
        page width specified in the content
      </dl>
 
+   <p>The value of <var>safePrintableInset</var> defines the minimum inset along
+   each page edge needed to steer clear of the area that the printer is potentially
+   incapable of marking reliably. Most printers have a small region along each
+   edge of the page sheet which is unprintable, typically due to the printer's
+   paper handling mechanism. The document that is being printed may have CSS rules
+   to adjust page margins based on this. TODO: Link to spec, when it has landed?
```
gsnedders: https://github.com/w3c/csswg-drafts/pull/13190 has now landed, so this should now cross-reference to that.

mstensho:
This has been stuck in review for some time.

index.html
```diff
@@ -11364,6 +11371,14 @@ <h3><dfn>Print Page</dfn></h3>
        page width specified in the content
      </dl>
 
+   <p>The value of <var>safePrintableInset</var> defines the minimum inset along
+   each page edge needed to steer clear of the area that the printer is potentially
+   incapable of marking reliably. Most printers have a small region along each
+   edge of the page sheet which is unprintable, typically due to the printer's
+   paper handling mechanism. The document that is being printed may have
+   <a href="https://drafts.csswg.org/css-page-3/#page-margin-safety">CSS rules</a>
+   to adjust page margins based on this.
```
gsnedders: This should cross-reference https://drafts.csswg.org/css-page-3/#printable-area, should it not? We don't need to redefine what a printable/unprintable area is when we can just cross-reference it.

I'd be tempted to say something more along the lines of:

> The value of `safePrintableInset` defines the width \[ XXX: in what units? presumably cm to match everything else in WebDriver? \] of the [non-printable area](https://drafts.csswg.org/css-page-3/#printable-area) on all sides of the page sheet.

With the current text here I genuinely don't know where to start, because I don't know what this corresponds to, and what I'm meant to be changing as an implementer.

But also: shouldn't this just be in the table above? You're changing properties of the "paper" that CSS is targeting, right?

gsnedders (changes requested):

index.html
mstensho: Please take another look now.

PR #1955, opened 2026-03-26
Throw "invalid cookie domain" as its definition implies

yezhizhen:
See https://github.com/servo/servo/pull/43690#discussion_r2994771111

The definition of [InvalidCookieDomain](https://w3c.github.io/webdriver/#dfn-invalid-cookie-domain) is: An illegal attempt was made to set a cookie under a different domain than the current page. 

But somehow, the spec asks to throw "invalid argument" error for this case.
However, all browsers indeed throw "invalid cookie domain" for this case.

Following test aligns with current spec, yet fails for all browsers.
```python
def test_add_cookie_with_domain_not_matching_current_domain(session, url):
    session.url = url("/common/blank.html")

    new_cookie = {
        "name": "hello",
        "value": "world",
        "domain": "example.com",
        "path": "/",
        "httpOnly": False,
        "secure": False
    }

    response = add_cookie(session, new_cookie)
    assert_error(response, "invalid argument")
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1955)

***
<a href="https://pr-preview.s3.amazonaws.com/yezhizhen/webdriver/pull/1955.html" title="Last updated on Apr 1, 2026, 3:57 AM UTC (555772e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1955/1d4b691...yezhizhen:555772e.html" title="Last updated on Apr 1, 2026, 3:57 AM UTC (555772e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -7603,6 +7603,12 @@ variables</var> and <var>parameters</var> are:
  <li><p><a>Try</a> to <a>handle any user prompts</a>
   with <var>session</var>.
 
+  <li><p>If <a>cookie name</a> or <a>cookie value</a> is <a><code>null</code></a>,
+  <a>cookie secure only</a> or <a>cookie HTTP only</a> are not boolean types,
+  or <a>cookie expiry time</a> is not an integer type, or it less than 0 or greater than
+  the <a>maximum safe integer</a>, return <a>error</a> with <a>error
+  code</a> <a>invalid argument</a>.
+
  <li><p>If <var>session</var>&apos;s <a>current browsing
   context</a>&apos;s <a>document element</a> is
   a <a>cookie-averse <code>Document</code> object</a> or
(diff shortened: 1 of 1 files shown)


yezhizhen:
cc @whimboo @xiaochengh

whimboo:
This makes sense. Thanks for discovering that discrepancy. The [referenced tests can be found here](https://wpt.fyi/results/webdriver/tests/classic/add_cookie/add.py?label=master&label=experimental&product=firefox&product=firefox_android&product=chrome&product=chrome_android&product=safari&view=subtest) but I have to add that we only test unsupported schemes for `invalid cookie domain`. We should indeed at least one more test with a valid domain. @yezhizhen mind adding one respectively? Thanks!

yezhizhen:
Adding one more in https://github.com/web-platform-tests/wpt/pull/58801

index.html
whimboo: Would you mind to move this second block up before the `invalid cookie domain` one so that we have the type checks for the cookie data together? In the moment the browsing context check separates them. 

This will also fix the issue that we should fail with `invalid argument` when the `cookie domain` is invalid. We should only see `invalid cookie domain` if the type check for `cookie domain` succeeded but the actual domain is incorrect.

This is what browsers actually should do these days as well and we actually miss all those invalid test cases for this command. :/ But to add those is not a blocker for this PR to get merged.

whimboo (changes requested):

whimboo (approved):
Thanks!

whimboo:
@yezhizhen given that you contributed a lot already please try to join the W3C group if possible so that the IP check wouldn't cause us trouble.

yezhizhen:
> @yezhizhen given that you contributed a lot already please try to join the W3C group if possible so that the IP check wouldn't cause us trouble.

I just joined!

PR #1960, opened 2026-05-12
Get Accessibility Properties

spectranaut:
See https://github.com/WICG/aom/issues/203

Added/update sections:
* [Added route under "Element" section](https://pr-preview.s3.amazonaws.com/Igalia/webdriver/pull/1960.html#get-accessibility-properties)
* [Added "Accessibility" Section](https://pr-preview.s3.amazonaws.com/Igalia/webdriver/pull/1960.html#accessibility)
* [Add references](https://pr-preview.s3.amazonaws.com/Igalia/webdriver/pull/1960.html#index)

Things to maybe revisit:
* [Discussion on errors in explainer](https://github.com/WICG/aom/issues/203#issuecomment-1765432713) - should we/can we have more specific errors?
* [Discussion on IDs in the explainer](https://github.com/WICG/aom/issues/203#issuecomment-3963870273) - should the accessibility ID be more defined?

- - -
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1960)

***
<a href="https://pr-preview.s3.amazonaws.com/Igalia/webdriver/pull/1960.html" title="Last updated on Aug 18, 2026, 5:03 PM UTC (97a068d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1960/6bd15e4...Igalia:97a068d.html" title="Last updated on Aug 18, 2026, 5:03 PM UTC (97a068d)">Diff</a>


index.html
```diff
@@ -11116,6 +11182,68 @@ <h3><dfn>Take Element Screenshot</dfn></h3>
...
+<p> <dfn>Ignored accessible objects</dfn> are accessible objects that represent <a>elements which are excluded from the accessibility tree</a>. The <a>accessibility properties</a> of these <a>accessible objects</a> must only have the keys <va>accessibilityID</var>, <va>parent</var>, <va>children</var> and <va>role</var> set to <code>"none"</code>.
+
+<p>To <dfn>get a known accessible object</dfn> given <var>session</var> and <var>reference</var>:
+
+<ol class="algorithm">
+  <li><p>TO DO: Should we create a parallel to the specification of <a>get a known element</a>? Do we need to make a parallel structure to <a>browsing context group node map</a> and <a>node id map</a>?
+</ol>
+
+<section>
+<h3><dfn>Get Accessibility Properties for Accessible Object</dfn></h3>
```
spectranaut: I used `accessible object` instead of `accessible node` because `accessible object` is defined in ARIA.

index.html
```diff
@@ -1193,6 +1206,14 @@ <h3>Errors</h3>
    and cannot be brought into that viewport.
  </tr>
 
+ <tr>
+  <td><dfn>no such accessible</dfn>
+  <td>404
+  <td><code>no such accessible</code>
```
cookiecrook: Unless we’re trying to match existing precedent for this “no such” phrasing, I’d prefer:
```suggestion
  <td><code>accessible node not found</code>
```

index.html
```diff
@@ -1255,6 +1276,14 @@ <h3>Errors</h3>
   <td>A new <a>session</a> could not be created.
  </tr>
 
+ <tr>
+  <td><dfn>stale accessible reference</dfn>
+  <td>404
+  <td><code>stale accessible reference</code>
```
cookiecrook: Do we have confidence that the implementations will retain the difference between “was here but is now stale” and “not found for some other reason”? If not, the first 404 may suffice.

index.html
cookiecrook: Or if people really want to keep the precedent, let’s call it “no such accessible node” rather than “no such accessible”

index.html
```diff
@@ -11116,6 +11182,68 @@ <h3><dfn>Take Element Screenshot</dfn></h3>
 </section> <!-- /Take Element Screenshot -->
 </section> <!-- /Screen capture -->
 
+<section>
+<h2>Accessibility</h2>
+
+<p> An <dfn>accessibility ID</dfn> is a string value representing a handle to an <a>accessible object</a> in a specific WebDriver session.
+
+<p> <dfn>Accessibility properties</dfn> is a JSON <a>Object</a> that contains the relevant <a>WAI-ARIA states and properties</a> of an <a>accessible object</a>, as well as the following properties:
```
cookiecrook: Not to downgrade ARIA, but it’s a subset here:

```suggestion
<p> <dfn>Accessibility properties</dfn> is a JSON <a>Object</a> that contains the relevant <a>accessibility properties</a> of an <a>accessible object</a>, as well as the following properties:
```

index.html
```diff
@@ -11116,6 +11182,68 @@ <h3><dfn>Take Element Screenshot</dfn></h3>
...
+
+  <dl>
+   <dt>"<code>accessibilityId</code>"
+   <dd>The <a>accessibility ID</a> of this <a>accessible object</a>.
```
cookiecrook: Object or node?

index.html
```diff
@@ -11116,6 +11182,68 @@ <h3><dfn>Take Element Screenshot</dfn></h3>
...
+
+   <dt>"<code>parent</code>"
+   <dd>The <a>accessibility ID</a> of the parent of this <a>accessible object</a> in the <a>accessibility tree</a>, including <a>ignored accessible objects</a>.
+
+   <dt>"<code>children</code>"
+   <dd>A list of <a>accessibility IDs</a> representing the children of this <a>accessible object</a> in the <a>accessibility tree</a>, including <a>ignored accessible objects</a>.
```
cookiecrook:
```suggestion
   <dd>A list of <a>accessibility IDs</a> representing the child nodes of this <a>accessible node</a> in the implementation’s <a>accessibility tree</a>.
```

index.html
```diff
@@ -11116,6 +11182,68 @@ <h3><dfn>Take Element Screenshot</dfn></h3>
...
+  </dl>
+
+<p> <dfn>Ignored accessible objects</dfn> are accessible objects that represent <a>elements which are excluded from the accessibility tree</a>. The <a>accessibility properties</a> of these <a>accessible objects</a> must only have the keys <va>accessibilityID</var>, <va>parent</var>, <va>children</var> and <va>role</var> set to <code>"none"</code>.
```
cookiecrook: Ignored accessible children may not be implementable as written. The “ignored” state is usually on the non-accessibility rendered object which can—modulo implementation details—result in no backing accessibility node to return in the child list.

index.html
```diff
@@ -11116,6 +11182,68 @@ <h3><dfn>Take Element Screenshot</dfn></h3>
...
+
+<p>To <dfn>get a known accessible object</dfn> given <var>session</var> and <var>reference</var>:
+
+<ol class="algorithm">
+  <li><p>TO DO: Should we create a parallel to the specification of <a>get a known element</a>? Do we need to make a parallel structure to <a>browsing context group node map</a> and <a>node id map</a>?
```
cookiecrook: I don’t think we need it, unless the element ID can’t be returned in the property bag.

index.html
cookiecrook: Node not Object, right?

index.html
```diff
@@ -11602,7 +11730,11 @@ <h2>Index</h2>
    <p>The following terms are defined
      in the Accessible Rich Internet Applications (WAI-ARIA) 1.2 specification: [[wai-aria-1.2]]
      <ul>
+       <li><dfn><a href="https://w3c.github.io/aria/#dfn-accessible-object"></a>accessible object</dfn>
```
cookiecrook: Ditto node?

cookiecrook: Maybe we should sync an ARIA edit with this PR?

index.html
```diff
@@ -11602,7 +11730,11 @@ <h2>Index</h2>
...
+       <li><dfn><a href="https://w3c.github.io/aria/#accessibility_tree">accessibility tree</a></dfn>
+       <li><dfn><a href="https://w3c.github.io/aria/#tree_exclusion">elements which are excluded from the accessibility tree</a></dfn>
        <li><dfn><a href="https://w3c.github.io/aria/#introroles">WAI-ARIA role</a></dfn>
+       <li><dfn><a href="https://w3c.github.io/aria/#introstates">WAI-ARIA states and properties</a></dfn>
```
cookiecrook: We’ll need a landing spot for the new “accessibility properties” ref above because this ARIA list is a subset.

index.html
spectranaut: following convention with "no such accessible node"

index.html
spectranaut: This suggestion is a bit tautological :)

I wonder if we should defined "computed accessibility properties" like we defined "computed role" in core-aam, and reference that...?

index.html
spectranaut: good point, I think we don't need this definition using your definition above.

index.html
spectranaut: I think we need a term for the implementation's accessibility tree that is not "accessibility tree", maybe "internal accessibility tree" or "computed accessibility tree", because "accessibility tree" alone is defined in ARIA and it basically means the platform API's tree.

index.html
spectranaut: Right so I originally used "accessible object" everywhere because that is defined in aria and "accessibility node" is not defined anywhere. But given your other comments... I'm realizing we can't define things in terms of ARIA's "accessibility tree" because that (and "accessibility object") is defined in terms of the platform accessibility API. And we are talking about something else, here.

spectranaut (commented):
First pass at fixes to your comments, @cookiecrook, but I still need to work on definitions and where they belong!

index.html
cookiecrook: To discuss in next week's API sync with @jcsteh et al.

index.html
cookiecrook: From he API meeting yesterday, this will likely land in either Core-AAM as a formalized property list... Most will reference ARIA attribute concepts, but some, like [size/bounds](https://github.com/web-platform-tests/interop-accessibility/issues/220), ignored (maybe), clickable, etc. may be specified in Core-AAM or another relevant AAM.

index.html
jcsteh: FWIW, I was going to make the same observation and then saw that it had already been made in this comment. As it currently stands, Gecko's implementation has no concept of stale nodes, though it could be added if needed since we have a specific accessibility id map for WebDriver/Marionette anyway. That said, I don't see a use case for this distinction, though I could potentially be convinced otherwise.

index.html
```diff
@@ -6028,6 +6057,43 @@ <h4><dfn>Get Computed Label</dfn></h4>
...
+        return <a>error</a> with <a>error code</a> <a>no such window</a>.
+
+    <li><p><a>Try</a> to <a>handle any user prompts</a>
+     with <var>session</var>.
+
+    <li><p>Let <var>element</var> be the result
+        of <a>trying</a> to <a>get a known element</a>
+        with <var>URL variables</var>["<code>element id</code>"].
+
+    <li><p>Let <var>accessible</var> be the <a>accessible node</a> that corrosponds to this element in the <a>accessibility tree</a>.
```
jcsteh: nit:
```suggestion
    <li><p>Let <var>accessible</var> be the <a>accessible node</a> that corresponds to this element in the <a>accessibility tree</a>.
```

index.html
```diff
@@ -11116,6 +11182,70 @@ <h3><dfn>Take Element Screenshot</dfn></h3>
 </section> <!-- /Take Element Screenshot -->
 </section> <!-- /Screen capture -->
 
+<section>
+<h2>Accessibility</h2>
+
+<p> An <dfn>accessibility ID</dfn> is a string value representing a handle to an <a>accessible node</a> in a specific WebDriver session.
+
+<p> An <dfn>accessibility node</dfn> is an node in the implementation defined <a>internal accessibility tree</a>. These nodes frequently but not always correlated to DOM elements.
```
jcsteh: Elsewhere, @spectranaut noted:

> because "accessibility tree" alone is defined in ARIA and it basically means the platform API's tree.

I think I'm still missing something here. I don't see how these trees are fundamentally different. From ARIA:

> The accessibility tree and the DOM tree are parallel structures. The accessibility tree includes the user interface objects of the user agent and the objects of the document. Accessible objects are created in the accessibility tree for every DOM element that should be exposed to an assistive technology, either because it might fire an accessibility event or because it has a property, relationship or feature which needs to be exposed. 

As I see it, all of that applies here. There are some subtle differences between platforms which we need to eliminate here because we're doing platform independent testing, but for the most part, the trees have a similar structure. While some browsers do have an internal concept of ignored nodes, I don't think those are exposed to any platform, though please correct me if I'm wrong.

In any case, on the call today, we discussed using a term like "implementation defined platform independent accessibility tree" or similar to clarify this a little. It's still a little wishy-washy, but it's probably the best we can do here for now.

index.html
```diff
@@ -1255,6 +1276,14 @@ <h3>Errors</h3>
   <td>A new <a>session</a> could not be created.
  </tr>
 
+ <tr>
+  <td><dfn>stale accessible node reference</dfn>
+  <td>404
+  <td><code>stale accessible node reference</code>
+  <td>A <a>command</a> failed because
+   the referenced <a>accessible node</a> is no available.
```
jcsteh: nit, though this may be irrelevant if we strike this error code:
```suggestion
   the referenced <a>accessible node</a> is no longer available.
```

index.html
jcsteh: nit:
```suggestion
<p> An <dfn>accessibility node</dfn> is a node in the implementation defined <a>internal accessibility tree</a>. These nodes frequently but not always correlated to DOM elements.
```

index.html
```diff
@@ -11116,6 +11182,70 @@ <h3><dfn>Take Element Screenshot</dfn></h3>
...
+   <dd>The <a>accessibility ID</a> of this <a>accessible node</a>.
+
+   <dt>"<code>parent</code>"
+   <dd>The <a>accessibility ID</a> of the parent of this <a>accessible node</a> in the <a>accessibility tree</a>, including <a>ignored accessible nodes</a>.
+
+   <dt>"<code>children</code>"
+   <dd>A list of <a>accessibility IDs</a> representing the child nodes of this <a>accessible node</a> in the implementation’s <a>internal accessibility tree</a>
+  </dl>
+
+<p> <dfn>Ignored accessible nodes</dfn> are accessible nodes that represent <a>elements which are excluded from the accessibility tree</a>. The <a>accessibility properties</a> of these <a>accessible nodes</a> must only have the keys <va>accessibilityID</var>, <va>parent</var>, <va>children</var> and <va>role</var> set to <code>"none"</code>.
```
jcsteh: On the call today, we discussed excluding ignored nodes from this tree on the grounds that they are internal only and not meant to be perceived by users or assistive technologies. Note that we can still test that something is excluded from the accessibility tree by testing for the non-existence of the node.

index.html
```diff
@@ -11603,6 +11733,7 @@ <h2>Index</h2>
      in the Accessible Rich Internet Applications (WAI-ARIA) 1.2 specification: [[wai-aria-1.2]]
      <ul>
        <li><dfn><a href="https://w3c.github.io/aria/#introroles">WAI-ARIA role</a></dfn>
+       <li><dfn><a href="https://w3c.github.io/aria/#fixme">Computed acccessible properties</a></dfn>
```
jcsteh: nit:
```suggestion
       <li><dfn><a href="https://w3c.github.io/aria/#fixme">Computed accessible properties</a></dfn>
```

jcsteh (commented):
Thanks for all of your work on this.

index.html
spectranaut: updated to accessibility node.

spectranaut:
I can't actually add people as reviewers, but, this is ready for re-review after all the feedback from the meetings the last few days, and the review comments from James and Jamie.

@jcsteh @cookiecrook @jugglinmike @lolaodelola , if you could all take a look :)

index.html
spectranaut: Latest update: I ended up removing the term "computed accessibility tree" and just refer to the ARIA's "accessibility tree", but I left a definition of "accessibility node", instead of referring to ARIA "accessible object",

index.html
```diff
@@ -1040,6 +1040,19 @@ <h3>Endpoints</h3>
   <td>/session/{<var>session id</var>}/print</td>
   <td><a>Print Page</a></td>
  </tr>
+
+<tr>
+  <td>GET</td>
+  <td>/session/{<var>session id</var>}/element/{<var>element id</var>}/accessibilityproperties</td>
+  <td><a>Get Accessibility Properties</a></td>
```
jcsteh: Sorry I missed this in the last pass. I understand that this links to a section under Elements, but looking at it here in the endpoints section, we see this ambiguity between "get accessibility properties" and "get accessibility properties for accessibility node". Would it make sense to rename this to "Get Accessibility Properties for Element"?

index.html
```diff
@@ -11116,6 +11173,60 @@ <h3><dfn>Take Element Screenshot</dfn></h3>
 </section> <!-- /Take Element Screenshot -->
 </section> <!-- /Screen capture -->
 
+<section>
+<h2>Accessibility</h2>
+
+<p> An <dfn>accessibility ID</dfn> is a string value representing a handle to an <a>accessibility node</a> in a specific WebDriver session.
+
+<p> An <dfn>accessibility node</dfn> is a platform-independent, implementation-defined node in the platform-independent, implementation-defined accessibility tree. The platform-independent, implementation defined accessibility tree is built from the DOM tree for the purposes of interacting with the web page via an <a>accessibility API</a>. Accessibility nodes might exist in this tree for which there is no DOM element back it (for example, one might be create to represent a CSS pseudo element). Additionally, an accessibility node might not be created for every DOM node (for example, if something is intentionally hidden from accessibility APIs using <a>aria-hidden</a>). See the ARIA definition of the <a>accessibility tree</a> for more information.
```
jcsteh: nit:
```suggestion
<p> An <dfn>accessibility node</dfn> is a platform-independent, implementation-defined node in the platform-independent, implementation-defined accessibility tree. The platform-independent, implementation defined accessibility tree is built from the DOM tree for the purposes of interacting with the web page via an <a>accessibility API</a>. Accessibility nodes might exist in this tree for which there is no DOM element back it (for example, one might be created to represent a CSS pseudo element). Additionally, an accessibility node might not be created for every DOM node (for example, if something is intentionally hidden from accessibility APIs using <a>aria-hidden</a>). See the ARIA definition of the <a>accessibility tree</a> for more information.
```

jcsteh (approved):
Thanks. Looks good to me.

index.html
spectranaut: ty for catching all my typos/grammar mistakes :)

index.html
spectranaut: you're right that seem reasonable! switched!

lolaodelola (approved):
This looks good to me

index.html
```diff
@@ -11116,6 +11173,60 @@ <h3><dfn>Take Element Screenshot</dfn></h3>
 </section> <!-- /Take Element Screenshot -->
 </section> <!-- /Screen capture -->
 
+<section>
+<h2>Accessibility</h2>
+
+<p> An <dfn>accessibility ID</dfn> is a string value representing a handle to an <a>accessibility node</a> in a specific WebDriver session.
```
cookiecrook: Accessibility **_node_** ID? 

Those of us that live in an accessibility headspace full time may take this for granted, but I'm trying to be explicit (through redundancy) so web developers and browser engine implementors don't confuse this as being a property of a DOM node, for example.

index.html
```diff
@@ -11603,6 +11714,10 @@ <h2>Index</h2>
      in the Accessible Rich Internet Applications (WAI-ARIA) 1.2 specification: [[wai-aria-1.2]]
      <ul>
        <li><dfn><a href="https://w3c.github.io/aria/#introroles">WAI-ARIA role</a></dfn>
+       <li><dfn><a href="https://w3c.github.io/aria/#accessibility_tree">Accessibility tree</a></dfn>
```
cookiecrook, line 11701: **_browser_** accessibility tree? b/c people may confuse this with the **_platform_** accessibility tree.

cookiecrook: Now that we're formalizing portions of the browser accessibility tree, ARIA also needs a spec issue to distinguish the **_browser_** accessibility tree from the **_platform_** accessibility tree.

cookiecrook (approved):

index.html
cookiecrook: - https://github.com/w3c/aria/issues/2817

index.html
spectranaut: Good idea, switched.

index.html
spectranaut: I'm not totally sure we need to this... I originally thought we did, but let's discuss it in the issue. For now, the information in the accessibility tree as it is defined in ARIA is the information in this platform-agnostic implementation defined tree -- which backs the platform accessibility trees. If we do decided to make a definition for the browser accessibility tree, we can update this spec text!

cookiecrook: okay to submit this part as-is, and address in the other issue.

index.html
cookiecrook: Any reason why you didn't make the change in the WebDriver path too? Search PR for three remaining instances of `accessibility id`

index.html
spectranaut: I never replied to this comment, but I did update the language a bit, just commenting to resolve this. Also we are going to discuss the ARIA accessibility tree definition here: https://github.com/w3c/aria/issues/2817, which your comments are relevant for!

index.html
```diff
@@ -11116,6 +11173,60 @@ <h3><dfn>Take Element Screenshot</dfn></h3>
 </section> <!-- /Take Element Screenshot -->
 </section> <!-- /Screen capture -->
 
+<section>
+<h2>Accessibility</h2>
+
+<p> An <dfn>accessibility node ID</dfn> is a string value representing a handle to an <a>accessibility node</a> in a specific WebDriver session.
+
+<p> An <dfn>accessibility node</dfn> is a platform-independent, implementation-defined node in the platform-independent, implementation-defined accessibility tree. The platform-independent, implementation defined accessibility tree is built from the DOM tree for the purposes of interacting with the web page via an <a>accessibility API</a>. Accessibility nodes might exist in this tree for which there is no DOM element back it (for example, one might be created to represent a CSS pseudo element). Additionally, an accessibility node might not be created for every DOM node (for example, if something is intentionally hidden from accessibility APIs using <a>aria-hidden</a>). See the ARIA definition of the <a>accessibility tree</a> for more information.
```
jugglinmike: There's a plurality mismatch here ("Accessibility nodes [...] it") and we need to do something with "back" (e.g. "to back" or "backing").

index.html
```diff
@@ -11116,6 +11173,60 @@ <h3><dfn>Take Element Screenshot</dfn></h3>
...
+
+  <dl>
+   <dt>"<code>accessibilityId</code>"
+   <dd>The <a>accessibility node ID</a> of this <a>accessibility node</a>.
+
+   <dt>"<code>parent</code>"
+   <dd>The <a>accessibility node ID</a> of the parent of this <a>accessibility node</a> in the <a>accessibility tree</a>.
+
+   <dt>"<code>children</code>"
+   <dd>A list of <a>accessibility node IDs</a> representing the child nodes of this <a>accessibility node</a> in the <a>accessibility tree</a>.
```
jugglinmike: Technically, since we're constructing a "JSON object", I think this should be an Array (an ECMAScript value) rather than a list (an Infra value).

index.html
```diff
@@ -11116,6 +11173,60 @@ <h3><dfn>Take Element Screenshot</dfn></h3>
...
+<h3><dfn>Get Accessibility Properties for Accessibility Node</dfn></h3>
+
+  <table class="simple jsoncommand">
+    <tr>
+      <th>HTTP Method</th>
+      <th>URI Template</th>
+    </tr>
+    <tr>
+      <td>GET</td>
+      <td>/session/{<var>session id</var>}/accessibiltiy/properties/{<var>accessibility id</var>}/</td>
```
jugglinmike: A couple typos: "accessibility", and we can nix the trailing forward-slash.

Taking a step back, though, I wonder if we should design this resource more like the element commands. That is, given that we have:

    /session/{session ID}/element/{element ID}/rect

A resource like:

    /session/{session ID}/accessibility/node/{accessibility node ID}/properties

Looks a little more familiar (and hopefully more predictable) than

    /session/{session ID}/accessibility/properties/{accessibility node ID}

I'd personally be in favor of such an alteration purely on the basis of parity, but it would help if there was a credible expectation for further accessible-node-based commands in the future. Do you know any? Maybe `/element` to get the backing DOM element (where available)?

index.html
```diff
@@ -11116,6 +11173,60 @@ <h3><dfn>Take Element Screenshot</dfn></h3>
...
 </section> <!-- /Screen capture -->
 
+<section>
+<h2>Accessibility</h2>
+
+<p> An <dfn>accessibility node ID</dfn> is a string value representing a handle to an <a>accessibility node</a> in a specific WebDriver session.
+
+<p> An <dfn>accessibility node</dfn> is a platform-independent, implementation-defined node in the platform-independent, implementation-defined accessibility tree. The platform-independent, implementation defined accessibility tree is built from the DOM tree for the purposes of interacting with the web page via an <a>accessibility API</a>. Accessibility nodes might exist in this tree for which there is no DOM element back it (for example, one might be created to represent a CSS pseudo element). Additionally, an accessibility node might not be created for every DOM node (for example, if something is intentionally hidden from accessibility APIs using <a>aria-hidden</a>). See the ARIA definition of the <a>accessibility tree</a> for more information.
+
+<p> <dfn>Accessibility properties</dfn> is a JSON <a>Object</a> that contains the relevant <a>computed accessibility properties</a> of an <a>accessibility node</a>, as well as the following properties:
```
jugglinmike: An unadorned "relevant" seems a little dubious for conformance-seekers. I imagine that the properties are implementation-defined (though I can't find "computed accessibility properties" in CORE-AAM 1.2 or its issue tracker); if that's part of the definition, is "relevant" necessary here?

jugglinmike (changes requested):

index.html
```diff
@@ -1040,6 +1040,19 @@ <h3>Endpoints</h3>
   <td>/session/{<var>session id</var>}/print</td>
   <td><a>Print Page</a></td>
  </tr>
+
+<tr>
```
OrKoN, line 1044: We already have some a11y based tooling in the WebDriver BiDi spec (in locateNodes). Is it possible to move the spec work to the WebDriver BiDi spec instead of this spec?

index.html
```diff
@@ -11614,6 +11729,14 @@ <h2>Index</h2>
               Name and Description Computation</a></dfn>
       </ul>
 
+ <dd>
+   <p>The following terms are defined
+     in the Core Accessibility API Mappings (Core-AAM) 1.2 specification: [[core-aam-1.2]]
+     <ul>
+       <li><dfn><a href="https://w3c.github.io/core-aam/#computed_accessibility_properties">Computed accessibility properties</a></dfn>
```
OrKoN: this link does not seem to work for me? is the content for this in a PR?

index.html
```diff
@@ -11116,6 +11173,60 @@ <h3><dfn>Take Element Screenshot</dfn></h3>
...
+ <ol>
+    <li><p>If <var>session</var>&apos;s <a>current browsing context</a> is <a>no longer open</a>,
+        return <a>error</a> with <a>error code</a> <a>no such window</a>.
+
+    <li><p><a>Try</a> to <a>handle any user prompts</a>
+     with <var>session</var>.
+
+    <li><p>Let <var>node</var> be the <a>accessibility node</a> with <a>accessibility node ID</a> matching the <var>URL variables</var>["<code>accessibility id</code>"]. If no such <a>accessibility node</a> exists, return <a>error</a> with <a>error code</a> <a>no such accessibility node</a>.
+
+    <li><p>Let <var>properties</var> be the result of computing the <a>accessibility properties</a> of <var>node</var>.
```
OrKoN, line 11204: Do you know how well does this corresponds to https://chromedevtools.github.io/devtools-protocol/tot/Accessibility/#type-AXNode?

index.html
spectranaut: ty! fixed.

index.html
spectranaut: ty! fixed.

index.html
spectranaut: Right, there is a PR for "computed accessibility properties": https://github.com/w3c/aria/pull/2800

To see the spec text: https://deploy-preview-2800--wai-aria.netlify.app/core-aam/#computed_accessibility_properties

The spec text does explain exactly which properties should appear for a given accessibility node. So maybe you are right, the relevant is superfluous, the "computer accessibility properties" is a defined set, I'll remove it :)

index.html
spectranaut: Thanks for the ideas here, I wonder what the other reviews would think, and we should discuss this at the next accessibility interop meeting.

index.html
spectranaut: So sorry -- I should have made that more clear. This is also currently being specified and there is a PR: w3c/aria#2800

To see the spec text: https://deploy-preview-2800--wai-aria.netlify.app/core-aam/#computed_accessibility_properties

We need to get agreement between all the browsers on this, and we are working on that actively, it seems like it's close to landing.

index.html
spectranaut: Honestly, I'm not sure. I don't know that much about WebDriver BiDi -- I think we are targeting WebDriver Classic because Safari doesn't implement WebDriver BiDi.

Safari is already working on a POC of for this endpoint, and Firefox implemented the endpoint in MarionetteDriver. So it would open a lot of accessibility testing in WPT, for all three browsers, if it lands in WebDriver. I would like to hear about any reasons it should be in BiDi instead, though.

Would it make sense for it to be in WebDriver BiDi as well as Classic?

index.html
spectranaut: Yeah, actually, this will be backed by the devtools node -- the webdriver command goes through the devtools protocol to get this information. The difference is that to start it will be a subset of the properties -- just the ones all three browsers agree to expose, and the agreement is being worked out in the other PR (w3c/aria#2800).

spectranaut (commented):
Thanks @jugglinmike and @OrKoN ! I've some fixes, answers and questions.

index.html
OrKoN: WebDriver BiDi is an improvement over WebDriver classic and I think there is already feature parity plus additional features that WebDriver classic is lacking (network interception, emulation, events). In Chromium, we are actually not recommending to add new features to WebDriver classic https://chromium.googlesource.com/chromium/src/+/HEAD/docs/testing/web_platform_tests.md#tests-that-require-testing-apis as we want to avoid maintaining multiple implementations. AFAIK webkit has some WebDriver BiDi support.

index.html
OrKoN: are you also working on the implementation in Chromium?

index.html
```diff
@@ -1040,6 +1040,19 @@ <h3>Endpoints</h3>
   <td>/session/{<var>session id</var>}/print</td>
   <td><a>Print Page</a></td>
  </tr>
+
+<tr>
+  <td>GET</td>
+  <td>/session/{<var>session id</var>}/element/{<var>element id</var>}/accessibilityproperties</td>
```
OrKoN, line 1046: would these endpoints allow getting the entire a11y tree? what is the expected workflow? getting accessibilityproperties for the root element and then using `accessibility/properties/<accessibility id>` to traverse?

index.html
```diff
@@ -11116,6 +11173,60 @@ <h3><dfn>Take Element Screenshot</dfn></h3>
...
 </section> <!-- /Screen capture -->
 
+<section>
+<h2>Accessibility</h2>
+
+<p> An <dfn>accessibility node ID</dfn> is a string value representing a handle to an <a>accessibility node</a> in a specific WebDriver session.
+
+<p> An <dfn>accessibility node</dfn> is a platform-independent, implementation-defined node in the platform-independent, implementation-defined accessibility tree. The platform-independent, implementation defined accessibility tree is built from the DOM tree for the purposes of interacting with the web page via an <a>accessibility API</a>. Accessibility nodes might exist in this tree for which there are no DOM element backing them (for example, one might be created to represent a CSS pseudo element). Additionally, an accessibility node might not be created for every DOM node (for example, if something is intentionally hidden from accessibility APIs using <a>aria-hidden</a>). See the ARIA definition of the <a>accessibility tree</a> for more information.
+
+<p> <dfn>Accessibility properties</dfn> is a JSON <a>Object</a> that contains the <a>computed accessibility properties</a> of an <a>accessibility node</a>, as well as the following properties:
```
OrKoN: does it mean that the computed properties are keys on the same level as accessibilityId, parent, children? is there a chance to have conflicts in the future?

index.html
```diff
@@ -11116,6 +11173,60 @@ <h3><dfn>Take Element Screenshot</dfn></h3>
...
+
+  <dl>
+   <dt>"<code>accessibilityId</code>"
```
OrKoN: as far as I see the a11y properties do not link back to DOM element IDs? it seems like it might limit usefulness of the a11y properties?

index.html
```diff
@@ -11116,6 +11173,60 @@ <h3><dfn>Take Element Screenshot</dfn></h3>
...
+   <dd>The <a>accessibility node ID</a> of this <a>accessibility node</a>.
+
+   <dt>"<code>parent</code>"
+   <dd>The <a>accessibility node ID</a> of the parent of this <a>accessibility node</a> in the <a>accessibility tree</a>.
+
+   <dt>"<code>children</code>"
+   <dd>An <a>Array</a> of <a>accessibility node IDs</a> representing the child nodes of this <a>accessibility node</a> in the <a>accessibility tree</a>.
```
OrKoN: Is there something to link to that defines what child nodes are? should this extend into iframes or is it expected to work with https://www.w3.org/TR/webdriver2/#switch-to-frame?

index.html
OrKoN: it also seems like it would require many roundtrips to get properties for all elements.

index.html
OrKoN: Although I think the main part of this feature is in https://deploy-preview-2800--wai-aria.netlify.app/core-aam/#computed_accessibility_properties so we can re-use it in WebDriver BiDi and adapt the interfaces to WebDriver BiDi conventions (still this means somewhat duplicate efforts).

index.html
spectranaut: Yes, I sent it to you, but just in case here is is again: https://chromium-review.googlesource.com/c/chromium/src/+/7922212

index.html
spectranaut: @cookiecrook, how is the bidi support in Safari? Do you know/can you share whether it is coming soon enough to provide this feature there instead of webdriver classic?

index.html
spectranaut: Yes we will have some "subtree" tests which walk the accessibility tree by accessibility ID, for example: https://github.com/web-platform-tests/wpt/blob/master/wai-aria/subtree/tablist.tentative.html

(this test runs on firefox because of the marionette endpoint: https://wpt.fyi/results/wai-aria/subtree?label=experimental&label=master&aligned)

In general, though, the subtrees are not testable, because they are pretty different between browsers, and we don't think we will ever have full alignment. The main issue is how much "generic" nodes there are (things that represent something meaningless in the a11y tree, like a div for styling purposes). We expect only to test specific scenarios. We discussed having an endpoint for the whole subtree, but we think it's more comment to just test a node or a few children/parents, thus the current design.

index.html
spectranaut: Good question. From my perspective, working on ARIA and being an editor of Core-AAM, I'd guess we won't end up defining it there... but it's possible it would be a fall out of testing these internal accessibility trees and trying to get more alignment on them.

These computed accessibility trees started out as something like an "implementation detail", the original specified accessibility trees are in terms of the browser-exposed accessibility API (these are the things the screen reader actually uses to interact with the browser). This "computed accessibility tree", however, is useful to test because it is the platform independent and the same thing you see in the devtools. Getting alignment on it for role and name has solved a lot of interop issues, so we want to surface more properties. Sorry I don't know how much you know about the accessibility part of the browser, just thought I'd provide more information if it's helpful :) 

Ultimately, thought, I think it's fine if we bundle the computed properties in a key like "properties." I bring it up in the next WPT accessibility interop meeting.

index.html
spectranaut: They don't link back, it's true. Can you explain why it might limit the usefulness? Oh maybe you mean, if you have an accessibility node, and you want to click it, you don't know where to send that click?

That is a problem, I'm not sure right now how to get around, I'll bring it up.

spectranaut (commented):
Thanks again @OrKoN ! I'm out for a week, but wanted to reply to a few things. There is an WPT accessibility interop meeting July 1, I'll bring up some points there.

index.html
OrKoN: yes, this would be a use case we have in Puppeteer or in Chrome DevTools (a11y tree view), for example, but also if the a11y tree is traversed using the methods in PR, there seems to be no mechanism to do something with the backing DOM node or even know what DOM contributed this a11y node (if any). I understand that the proposed extension is probably useful for the WPT use cases but I also wonder how it would work for other WebDriver clients.

index.html
OrKoN: Thanks, I think it would be useful to put the computed properties into a separate key.

index.html
OrKoN: Thanks, that makes sense but I wonder once this is published if other WebDriver clients might start building out features based on the subtree iteration mechanisms available (thus starting to rely on the non-specified behavior) and require many roundtrips to implement the subtree retrieval.

index.html
cookiecrook: > Do you know/can you share whether it is coming soon enough to provide this feature there instead of webdriver classic?

All I can say is that WebDriver BiDi remains an area of interest for WebKit.

The Incubation group along with feedback from the the Browser Testing & Tools WG agreed at TPAC 2023 and then TPAC 2024 to include this work in WebDriver Classic but block both the Outgoing Notifications and Synthesized AT Events on WebDriver BiDi. So I would strongly prefer to continue our progress here in WebDriver Classic.

index.html
cookiecrook: I expect that will be the case, and don't see harm in testing the subtree of smaller features, like a known list or table with a known style sheet... However, it won't be a reliable cross-UA WPT test yet to include something like `dumpAccessiblityTree()` at the root node for a complex web site. Perhaps someday if and when we are able to standardize the rest.

index.html
cookiecrook: I would want these are same level as parentNode/childNodes/id, etc. I don't see a lot of benefit in putting an arbitrary collection of them into a sub-level `properties` container. Many of the to-be-specified don't align with attributes per se... location and size/bounds for example, would be computed using CSS and other features, but should that any different in the final object representation? What about the other relationships like `aria-controls` (or `label-from`)? Why should those be any different from the parent and children relationships?

index.html
cookiecrook: There should be an optional DOM Node ID reference from the Accessibility Node, if relevant... The reason it's optional is that not all Accessibility Nodes have a relevant DOM Node, and vice versa, not all DOM Elements have a backing Accessibility Node... 

An example of the latter is any hidden DOM node; not mapped in the AX Tree. An example of the former (AX Nodes without a DOM Element) is any CSS generated content, or sometimes (depending on implementation details, dynamic nodes like table columns or other in-betweener AX nodes)

index.html
OrKoN: Thanks, that sounds fine to me.

index.html
OrKoN: would it make sense to add a note about this to the spec text here?

index.html
OrKoN: I do not see an optional DOM Node ID reference in the spec. Could you please point me to the definition?

index.html
OrKoN: The reason I asked this question is to avoid potential conflicts if some properties defined by the `computed accessibility properties` end up clashing with properties defined here. Alternatively, we could say that the properties coming from computed accessibility properties override conflicting keys if any (or the other way around). I think called the key `properties` would be fine because it the algorithm is data it provides is called `computed accessibility properties`. If you think conflicts are unlikely, I think the current version looks good to me.

I am also wondering if perhaps the `accessibility node ID` term might move to the core-aam spec so that WebDriver spec uses it instead of defining it. Then the Core-AAM spec can define how to produce all of the attributes including parentNode, childNodes and all keys on the same level would be part of the same spec. It would also make it easier to re-use the definitions in WebDriver BiDi spec later on.

This is also somewhat related to the other question I had about how child nodes should be computed and whether the a11y nodes should span over iframes or if the client needs to use SwitchToFrame to inspect a11y tree that belongs to iframes.

index.html
cookiecrook: That could be defined here, but would more likely come from the TBD property bag definition in Core-AAM. @spectranaut?

index.html
spectranaut: We talked about this in the interop meeting, I'm going to try to define it along side the computed accessibility properties in Core-AAM, for re-using in bidi -- but I won't be able to move this forward for a few weeks!

index.html
spectranaut: In the case of webdriver, it would be the webdriver element ID. Maybe we will make some kind of abstraction in order to define this in Core-AAM -- because other testing frameworks (like marionette driver) will have a different kind of ID references to get back to the original DOM element.

index.html
spectranaut: We agreed to make this switch!

index.html
cookiecrook: It's a good question. The cross-frame and cross-process hop makes it tricky, and I'd defer to a security analysis of whether its wise to allow that here. @gsnedders, @twilco, @minorninth, @jcsteh, do you have any thoughts?

minorninth: I don't think it's a security concern, because my understanding is that a WebDriver client has full access to the whole browser - all tabs, all frames. So exposing the whole composed accessibility tree makes sense.

It might be trickier for browsers to implement globally-unique node IDs across all frames. But we definitely need to pick a direction to go here.

Ideas:
(1) browsers could handle all of the details transparently, either coordinating between frames to make globally unique ids, or making ids that are the composition of a frame id and node id
(2) we could expose both a frame id and node id for each node, mirroring the internal structure that (some) browsers actually use
(3) we could scope this API to a single frame and ensure there's a way to identify an iframe element and continue the search from there
...

lolaodelola:
The build is failing because it can't find `accessibility properties` or `accessibility node ID` in the CORE-AAM spec which makes sense because those changes haven't landed yet. There is a recommendation to use `lint-ignore` for those dfns but I think we should just wait until the changes in CORE have landed before merging this (unless someone is going to remember to remove the lint-ignores once CORE-AAM changes have landed).

The problem could be something completely different, have I missed something?

spectranaut:
> The build is failing because it can't find `accessibility properties` or `accessibility node ID` in the CORE-AAM spec which makes sense because those changes haven't landed yet. There is a recommendation to use `lint-ignore` for those dfns but I think we should just wait until the changes in CORE have landed before merging this (unless someone is going to remember to remove the lint-ignores once CORE-AAM changes have landed).
> 
> The problem could be something completely different, have I missed something?

Thanks for these changes Lola! I agree that we should land the Core-AAM change first.

index.html
jcsteh: I agree security isn't a concern here.

I don't think ids are a major concern here either. The id just has to be unique and constant for the lifetime of the accessibility node; the specific format is opaque. Fwiw, Gecko's internal ids are not globally unique, but for WebDriver, Gecko exposes GUIDs and has an internal map from GUIDs to accessibility objects, making our WebDriver accessibility node ids globally unique.

The bigger problem is that an iframe may not be able to access information about its embedded document at all if it's an out-of-process iframe due to site isolation. Rather than doing something specific to accessibility, I wonder if it'd be better to expose the embedded BrowsingContext id or similar, since (I think?) all browsers can get that information from the iframe, even if it's out-of-process. So iframes would be a bit special, but that doesn't seem unreasonable.

Issue #1961, opened 2026-05-19
"in-view center point" is broken when an inline element contains block-level elements


julienw:
In chapter 12. Elements, section [12.1 Interactability](https://w3c.github.io/webdriver/#interactability):
The first step of the algorithm to calculate `in-view center point ` says: 
```
1. Let rectangle be the first object of the DOMRect collection returned by calling getClientRects() on element. 
```

That rectangle is then used to find the center of the element.

The problem is when that first object has zero width and height, when another of the returned rects in the collection contain the main part of the element. This happens when an inline element such as a `<a>` contains block-level elements.

When coordinates are not integer, the final `floor` operation will yield coordinates that are outside the element. The `floor` operation doesn't make a lot of sense to me (the user _can_ click on non-integer CSS pixels when the dpi isn't 1), but just removing this `floor` operation wouldn't be good enough: clicks would work, but they wouldn't be at the center of the element.

So I think the step 1. above should be changed to:
```
1. Let rectangle be the first object of the DOMRect collection returned by calling getClientRects() on element, that has non-null width and height. If all have a null width and height, let rectangle be the first object of the returned collection.
```
(maybe this could be extracted to another algorithm)

Here is a test that demonstrates the problem in Firefox:
```python
def test_inline_element_with_zero_size_rects_at_fractional_position(
    session, inline
):
    # An inline <a> wrapping a block-level child produces zero-size line-box
    # rects at the start and end of its getClientRects() list: [0x0, 32x32,
    # 0x0]. Placing the element at a fractional y-coordinate (via
    # margin-top: 0.5px) means Math.floor on the zero-size rect's y lands
    # outside the element and elementsFromPoint misses it.
    # This test verifies that clicking such an element succeeds despite the
    # zero-size first rect, i.e. that the non-zero rect is used to determine
    # the element's position.
    session.url = inline("""
        <style>
          body { margin: 0; }
        </style>
        <div style="margin-top: 0.5px">
          <a id="link" href="#"
             onclick="document.getElementById('log').textContent = 'clicked'">
            <div style="width: 32px; height: 32px;"></div>
          </a>
        </div>
        <div id="log"></div>
    """)
    element = session.find.css("#link", all=False)

    # Verify the layout actually produces zero-size first rect; if this fails
    # the test premise has changed and the test should be revisited.
    first_rect = session.execute_script(
        """
        const r = arguments[0].getClientRects()[0];
        return { w: r.width, h: r.height };
        """,
        args=[element],
    )
[...]

PR #1964, opened 2026-05-27
Add scrollIntoView action

### LINKED ISSUE #1005: Missing "scroll into view" Actions primitive
Mid-way through executing a sequence of actions it's useful to be able to scroll an element into view. A "scroll into view" primitive would enable this.

hbenl:
This PR adds `scrollIntoView` to the general actions.
The action should also be able to scroll text nodes (which can be passed in from WebDriver BiDi) and elements with `display: contents` into view. To this end, the "scroll into view" steps are updated to create a `Range` object for such nodes and call the "Scroll target into view" steps in the CSSOM View spec (which support `Range`s) directly.
Note that this is a breaking change for interaction commands that implicitly scroll an element into view: previously, the implicit attempt to scroll an element with `display: contents` into view would do nothing but now it will be scrolled into view.

Fixes #1005.

- - -
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1964)

***
<a href="https://pr-preview.s3.amazonaws.com/hbenl/webdriver/pull/1964.html" title="Last updated on May 27, 2026, 1:21 PM UTC (efcfe63)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1964/905860d...hbenl:efcfe63.html" title="Last updated on May 27, 2026, 1:21 PM UTC (efcfe63)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4496,27 +4496,43 @@ and <var>element</var> is:
  or if it is not <a>connected</a>.
 
 <p>To <dfn data-lt="scrolls into view|scrolled into view">scroll into view</dfn>
- an <var><a>element</a></var>
- perform the following steps
- only if the element is not already <a>in view</a>:
+ a <var><a>node</a></var>
+ given optional arguments <var>onlyIfNecessary</var> (default true),
+ <var>behavior</var> (default "<code>instant</code>"),
+ <var>block</var> (default "<code>end</code>")
+ and <var>inline</var> (default "<code>nearest</code>"):
 
 <ol>
- <li><p>Let <var>options</var> be
-  the following <a><code>ScrollIntoViewOptions</code></a>:
+ <li><p>Let <var>target</var> be determined as follows:
 
-  <dl>
-   <dt>"<code>behavior</code>"
-   <dd>"<code>instant</code>"
+  <dl class=switch>
+   <dt><var>node</var> implements {{Element}}
+    and has an associated [=CSS/box=]
+   <dd>
+    <ol>
+     <li><p>Let <var>target</var> be <var>node</var>.
 
-   <dt><a>Logical scroll position "<code>block</code>"</a>
-   <dd>"<code>end</code>"
+     <li><p>If <var>onlyIfNecessary</var> is true
+      and <var>target</var> is <a>in view</a>, return.
+    </ol>
 
-   <dt><a>Logical scroll position "<code>inline</code>"</a>
-   <dd>"<code>nearest</code>"
+   <dt>Otherwise
+   <dd>
+    <ol>
+     <li><p>Let <var>target</var> be a new {{Range}} obtained by calling
+      {{Document/createRange()}} on <var>node</var>&apos;s [=Node/node document=]
+      and then invoking {{Range/selectNodeContents()}} on it with <var>node</var>
+      as argument.
+
+     <li><p>If <var>onlyIfNecessary</var> is true
+      and <var>target</var> is <a data-lt="range is in view">in view</a>, return.
+    </ol>
   </dl>
 
- <li><p>Run <a>Function.[[\Call]]</a>(<a>scrollIntoView</a>, <var>options</var>)
- with <var>element</var> as the this value.
+ <li><p><a>Scroll <var>target</var> into view</a>
+  with scroll behavior <var>behavior</var>,
+  block flow direction position <var>block</var>,
+  and inline base direction position <var>inline</var>.
 </ol>
 
 <p><dfn>Editable</dfn> <a>elements</a>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -4496,27 +4496,43 @@ <h2>Elements</h2>
...
 
 <p>To <dfn data-lt="scrolls into view|scrolled into view">scroll into view</dfn>
- an <var><a>element</a></var>
- perform the following steps
- only if the element is not already <a>in view</a>:
+ a <var><a>node</a></var>
+ given optional arguments <var>onlyIfNecessary</var> (default true),
+ <var>behavior</var> (default "<code>instant</code>"),
+ <var>block</var> (default "<code>end</code>")
+ and <var>inline</var> (default "<code>nearest</code>"):
```
whimboo, line 4503: Not sure if `nearest` would work as default. Usually we use the element's in-view center point for the interaction and scrolling only as little as possible most likely would still keep the majority of the element outside of the viewport. Should we maybe use `end` here as well similar to what we use for `block`?

index.html
```diff
@@ -4496,27 +4496,43 @@ <h2>Elements</h2>
...
-   <dd>"<code>instant</code>"
+  <dl class=switch>
+   <dt><var>node</var> implements {{Element}}
+    and has an associated [=CSS/box=]
+   <dd>
+    <ol>
+     <li><p>Let <var>target</var> be <var>node</var>.
+
+     <li><p>If <var>onlyIfNecessary</var> is true
+      and <var>target</var> is <a>in view</a>, return.
```
whimboo: This is the same for both blocks so you can factor it out to after the switch.

index.html
```diff
@@ -4496,27 +4496,43 @@ <h2>Elements</h2>
...
+    <ol>
+     <li><p>Let <var>target</var> be a new {{Range}} obtained by calling
+      {{Document/createRange()}} on <var>node</var>&apos;s [=Node/node document=]
+      and then invoking {{Range/selectNodeContents()}} on it with <var>node</var>
+      as argument.
 
-   <dt><a>Logical scroll position "<code>inline</code>"</a>
-   <dd>"<code>nearest</code>"
+     <li><p>If <var>onlyIfNecessary</var> is true
+      and <var>target</var> is <a data-lt="range is in view">in view</a>, return.
```
whimboo, line 4528: This renders `in view` similarly to the above option and is confusing. We most likely want `target's <a>range is in view</a>` or similar. 

Also here we only test the viewport overlapping for the first client rect (shouldn't we take all into account?). 

Per AI it might cause issues with `overflow: hidden` when `in view` would return true but it's actually clipped by the container. So `onlyIfNecessary` would wrongly skip the scroll. If that's acceptable it might warrant at least a note. By lets see what @jgraham will think about it.

index.html
```diff
@@ -9330,6 +9491,39 @@ <h4>General actions</h4>
...
+ <li><p>Let <var>behavior</var> be the <code>behavior</code>
+  property of <var>action object</var>.
+
+ <li><p>Let <var>block</var> be the <code>block</code>
+  property of <var>action object</var>.
+
+ <li><p>Let <var>inline</var> be the <code>inline</code>
+  property of <var>action object</var>.
+
+ <li><p><a>Scroll into view</a> <var>node</var> with
```
whimboo, line 9521: If we have a scrolling behavior of `auto` or `smooth` the scrolling will be asynchronous. We should wait until the scrolling is done before continuing dispatching the remaining actions.

index.html
```diff
@@ -4496,27 +4496,43 @@ <h2>Elements</h2>
...
-   <dt><a>Logical scroll position "<code>inline</code>"</a>
-   <dd>"<code>nearest</code>"
+     <li><p>If <var>onlyIfNecessary</var> is true
+      and <var>target</var> is <a data-lt="range is in view">in view</a>, return.
+    </ol>
   </dl>
 
- <li><p>Run <a>Function.[[\Call]]</a>(<a>scrollIntoView</a>, <var>options</var>)
- with <var>element</var> as the this value.
+ <li><p><a>Scroll <var>target</var> into view</a>
```
whimboo, line 4532: Is it ok to put a `var` inside an anchor?

index.html
```diff
@@ -8833,6 +8920,76 @@ <h3>Processing actions</h3>
  <li><p>Return success with data <var>action</var>.
 </ol>
 
+<p>To <dfn>process a scrollIntoView action</dfn> given <var>action item</var>,
+<var>action</var>, and <var>actions options</var>:</p>
+
+<ol class="algorithm">
+ <li><p>Let <var>node</var> be the result of <a>getting the property</a>
```
whimboo, line 8927: Some lines up we use `getting a property`. What shall we use?

whimboo (changes requested):
Thanks @hbenl. I did a first review pass for this proposed enhancement and I have a couple of comments. But I would also see feedback from @jgraham.

index.html
hbenl: I wanted the `scrollIntoView` action defaults to match [what WebDriver does](https://www.w3.org/TR/webdriver2/#dfn-scrolls-into-view) for the interaction commands.

index.html
hbenl: I assumed this is already covered by waiting for "pending asynchronous waits arising from the last invocation of the dispatch tick actions steps" [after dispatching an action](https://www.w3.org/TR/webdriver2/#dfn-dispatch-actions-inner) because [scroll a target into view](https://drafts.csswg.org/cssom-view/#scroll-a-target-into-view) returns a Promise that resolves when scrolling is done.
@jgraham did I understand this correctly?

index.html
hbenl: It's not ideal, so I changed it to `<a>Scroll into view</a> <var>target</var>`.

index.html
hbenl: Both `getting the property` and `getting a property` are used throughout the spec.

index.html
hbenl: I changed it to `the <var>target</var> <a>range is in view</a>` (not `target's range` because `target` itself is a range here).

> Also here we only test the viewport overlapping for the first client rect (shouldn't we take all into account?).

Again I tried to match how scrolling into view is done for WebDriver interaction commands: the [in view check](https://www.w3.org/TR/webdriver2/#dfn-in-view) uses the [ pointer-interactable paint tree](https://www.w3.org/TR/webdriver2/#dfn-pointer-interactable-paint-tree) which uses the element's [in-view center point](https://www.w3.org/TR/webdriver2/#dfn-center-point) which is "the intersection between the element's **first** [DOMRect](https://www.w3.org/TR/geometry-1/#domrect) of [getClientRects](https://www.w3.org/TR/cssom-view-1/#dom-element-getclientrects)() and the [initial viewport](https://www.w3.org/TR/webdriver2/#dfn-viewport)" (but I was surprised by this logic as well).

index.html
whimboo: I didn't check that already existing algorithm. Thanks for pointing out. In that case we probably can leave it as `nearest` yes.

index.html
whimboo: Mind then filing a new issue so that we can get this cleaned-up?

index.html
whimboo: You are probably right. But lets get a final review from James anyway.

index.html
whimboo: Oh I was actually thinking of https://github.com/w3c/webdriver/issues/1961 but as I can see now this is still open.

index.html
```diff
@@ -7942,8 +8003,9 @@ <h4>Key input source</h4>
  </tr>
 </table>
 
-<p>A <a>key input source</a> supports the same <a>pause</a> action
-as a <a>null input source</a> plus the following actions:
+<p>A <a>key input source</a> supports the same <a>pause</a>
```
jgraham, line 8006: Why do these input sources support `scrollIntoView`? `pause` is a special case here because it's used as the no-op action.

index.html
jgraham: No you'd need to specifically say to asynchronously wait for that promise to resolve.

jgraham (changes requested):

index.html
hbenl: Done.

index.html
hbenl: The action is not associated with any specific device, so I treated it like the `pause` action. Which input sources should we add it to? I think it makes most sense to add it to `pointer` actions because its main use case is to prepare for a click. Should we also add it to the `null` input source?

PR #1965, opened 2026-07-01
fix: prefer first non-zero rect for interactability checks

### LINKED ISSUE #1961: "in-view center point" is broken when an inline element contains block-level elements

juliandescottes:
Fixes #1961 

This matches the implementation as updated in Firefox. WPT tests already cover potential edge cases related to this (https://github.com/web-platform-tests/wpt/pull/60371)

- - -
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1965)

***
<a href="https://pr-preview.s3.amazonaws.com/juliandescottes/webdriver/pull/1965.html" title="Last updated on Jul 8, 2026, 2:27 PM UTC (8e07cfd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1965/926cd31...juliandescottes:8e07cfd.html" title="Last updated on Jul 8, 2026, 2:27 PM UTC (8e07cfd)">Diff</a>


juliandescottes:
I don't understand the build failure here, it seems to complain about <a>node</a> but this is unrelated to my change?

Edit: No idea why this only started failing now, but let's see: https://github.com/w3c/webdriver/pull/1966

whimboo:
@juliandescottes can you please rebase? CI actions should no longer show this error.

juliandescottes:
> @juliandescottes can you please rebase? CI actions should no longer show this error.

Sure,done!

index.html
```diff
@@ -4591,13 +4591,16 @@ <h3>Interactability</h3>
  is the origin position of the rectangle
  that is the intersection between
  the element&apos;s first {{DOMRect}} of {{Element/getClientRects()}}
+ with non-zero width and height (or the first {{DOMRect}} if none have non-zero dimensions)
  and the <a>initial viewport</a>.
```
whimboo, line 4595:
```suggestion
<p>An <a>element</a>&apos;s <dfn data-lt="center point">in-view center point</dfn>
is the origin position of the rectangle
that is the intersection between
the element&apos;s {{DOMRect}} and the <a>initial viewport</a>. 
The {{DOMRect}} is the first result of {{Element/getClientRects()}}
with non-zero width and height, or the first result if none qualify.
```

index.html
```diff
@@ -4591,13 +4591,16 @@ <h3>Interactability</h3>
...
  It can be calculated this way:
 
 <ol>
  <li><p>Let <var>rectangle</var> be
-  the first object of the {{DOMRect}} collection
-  returned by calling {{Element/getClientRects()}} on <a><var>element</var></a>.
+  the first object in the {{DOMRect}} collection
+  returned by calling {{Element/getClientRects()}} on <a><var>element</var></a>
+  that has non-zero width and height,
+  or the first object in the collection if no such object is found.
```
whimboo, line 4604: Another suggestion so that it reads better IMHO. Sadly I cannot get GitHub to let me create a suggestion. :(

> Let rectangle be the first object in the DOMRect collection returned by calling getClientRects() on element whose width and height are both non-zero. If no such object exists, let rectangle be the first object in the collection.

whimboo (commented):
Just two wording comments from my side.

PR #1966, opened 2026-07-01
fix: update incorrect node references

juliandescottes:
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1966)

***
<a href="https://pr-preview.s3.amazonaws.com/juliandescottes/webdriver/pull/1966.html" title="Last updated on Jul 2, 2026, 12:18 PM UTC (4263ffe)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webdriver/1966/7329641...juliandescottes:4263ffe.html" title="Last updated on Jul 2, 2026, 12:18 PM UTC (4263ffe)">Diff</a>


juliandescottes:
See https://github.com/w3c/webdriver/pull/1967 for an example of a "blank" PR which fails.

index.html
```diff
@@ -5485,16 +5485,16 @@ <h3>State</h3>
  <li><p>Return a pair of (<var>x</var>, <var>y</var>).
 </ol>
 
-<p>To determine if <a>node</a> is <dfn>not in the same tree</dfn>
- as another <a>node</a>, <var>other</var>,
+<p>To determine if <var>node</var> is <dfn>not in the same tree</dfn>
+ as another {{Node}}, <var>other</var>,
```
whimboo: The wording here is still inconsistent with the other algorithms. We usually use the pattern *"To <dfn>name</dfn> given <var>arg1</var> and <var>arg2</var>"*. Here, it isn't clear that `node` is actually an argument, which is likely the source of this issue. We probably can't change the algorithm name itself because it would no longer fit naturally at the call sites.

Perhaps something like the following would work better:

> `<p>To determine if a {{Node}} <var>node</var> is <dfn>not in the same tree</dfn> as a {{Node}} <var>other</var>,`

This makes both `node` and `other` explicit inputs while preserving the existing algorithm name.

whimboo:
```suggestion
<p>To determine if a {{Node}} <var>node</var> is <dfn>not in the same tree</dfn> as a {{Node}} <var>other</var>,
```

whimboo (approved):
Looks good to me with my proposed changes applied.

jgraham (approved):

PR #1974, opened 2026-09-17
fix: Restrict range of `Integer` to be within maximum safe integer

yezhizhen:
1. Change from `integer` to `Integer`, which is how it is referred to.
2. Restrict the absolute value to be less than or equal to [maximum safe integer](https://262.ecma-international.org/6.0/#sec-number.max_safe_integer). Previous definition can be as large as $1.7976931348623157 × 10^{308}$ according to definition.

This restriction is implicitly imposed by tests and used widely in spec, such as [process a pause action](https://w3c.github.io/webdriver/#dfn-process-a-pause-action):
> If duration is not [undefined](https://w3c.github.io/webdriver/#dfn-undefined) and duration is not an [Integer](https://w3c.github.io/webdriver/#dfn-integer) greater than or equal to 0, return [error](https://w3c.github.io/webdriver/#dfn-error) with [error code](https://w3c.github.io/webdriver/#dfn-error-code) [invalid argument](https://w3c.github.io/webdriver/#dfn-invalid-argument).

https://github.com/web-platform-tests/wpt/blob/87812b1aecbf4a2b892f540e077d9ac6fd90455c/webdriver/tests/classic/perform_actions/invalid.py#L10

- - -
This change is [Reviewable](https://reviewable.io/reviews/w3c/webdriver/1974)


yezhizhen:
cc @whimboo @gsnedders Can you check? This is quite broken otherwise for many operations. And tests already reflect it.

index.html
```diff
@@ -190,8 +190,9 @@ <h2>Terminology</h2>
 <a href=https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap04.html#tag_04_15>section 4.15</a> (IEEE Std 1003.1).
 
 <p>
-An <dfn>integer</dfn> is a <a>Number</a> that is unchanged
-under the <a>ToInteger</a> operation.
+An <dfn>Integer</dfn> is a <a>Number</a> that is unchanged
```
whimboo: I'm not sure if the rename to `Integer` with a capital starting letter is fine here given that we have references to `integer` a lot in this document. I wonder if we can leave the name as is and then can remove all the other mentions of maximum safe integer from its usages. It's quite a lot of duplication at the moment and no command can support integers that are larger than that. 

CC'ing @jgraham as well.

yezhizhen: > I'm not sure if the rename to Integer with a capital starting letter is fine here given that we have references to integer a lot in this document. 

Yeah I can change the renaming to another PR. It seems we only use "integer" once and "Integer" elsewhere.
> I wonder if we can leave the name as is and then can remove all the other mentions of maximum safe integer from its usages.

I can do that.

index.html
```diff
@@ -7604,9 +7605,8 @@ <h3><dfn data-lt="adding a cookie">Add Cookie</dfn></h3>
 
   <li><p>If <a>cookie name</a> or <a>cookie value</a> is <a><code>null</code></a>,
   <a>cookie secure only</a> or <a>cookie HTTP only</a> are not boolean types,
-  or <a>cookie expiry time</a> is not an integer type, or it less than 0 or greater than
-  the <a>maximum safe integer</a>, return <a>error</a> with <a>error
```
yezhizhen, line 7608: > I wonder if we can leave the name as is and then can remove all the other mentions of maximum safe integer from its usages.

I removed ref to max safe int here. But should probably keep the other one in [timeouts config](https://w3c.github.io/webdriver/#dfn-timeouts-configuration):
> If value is neither null nor a number greater than or equal to 0 and less than or equal to the [maximum safe integer](https://w3c.github.io/webdriver/#dfn-maximum-safe-integer) return [error](https://w3c.github.io/webdriver/#dfn-error) with [error code](https://w3c.github.io/webdriver/#dfn-error-code) [invalid argument](https://w3c.github.io/webdriver/#dfn-invalid-argument).

index.html
```diff
@@ -7604,9 +7605,8 @@ <h3><dfn data-lt="adding a cookie">Add Cookie</dfn></h3>
...
-  code</a> <a>invalid argument</a>.
+  or <a>cookie expiry time</a> is not an <a>Integer</a> or less than 0,
```
whimboo: Is there a reason why you are using `Integer` instead now? Above you now kept `integer` so the names won't match.

index.html
```diff
@@ -191,7 +191,8 @@ <h2>Terminology</h2>
 
 <p>
 An <dfn>integer</dfn> is a <a>Number</a> that is unchanged
-under the <a>ToInteger</a> operation.
+under the <a>ToInteger</a> operation and less than or equal to
```
whimboo: Interesting that we as well use `Integer` with a capital `I` when referencing the `integer` type like in pause or wheel actions. Maybe it would be fine to just update it? The checks didn't seem to have issues with it. Would be good to get feedback from others. CC @jgraham.

nit: I would mention the -maximum safe integer first to match the usual range definitions.

index.html
whimboo: Yes, that's a different thing when used with `number`. But I actually wonder if the `timeouts config`'s values should be `integer` instead? Not sure if it makes sense to have settings in microseconds. But that's a different topic.

index.html
whimboo: @yezhizhen could you apply the nit while we are waiting for further feedback?

yezhizhen: I think I already applied in https://github.com/w3c/webdriver/pull/1974/commits/4ecd6b6251f12587b093a71ed6329b12c75ce419?

whimboo: Ah, it was a misunderstanding. The `ToInteger` should still come first, but then have the lower bound check first with the upper bound following.

yezhizhen: See https://github.com/w3c/webdriver/pull/1974/commits/62feb8c7cc28b71f5ad13410f3c4c27986279f81

whimboo (approved):
Looks fine to me.

@OrKoN or @sadym-chromium mind a review as well? Thanks.

